PR #151, opened 2011-01-11
support rpm builds

tilgovi:
With this patch `python setup.py bdist_rpm` works and generates proper dependencies and installation procedures. Tested on CentOS5. I followed [PythonPackaging](https://fedoraproject.org/wiki/Packaging:Python) and [Eggs](http://fedoraproject.org/wiki/Packaging:Python_Eggs) pretty closely and the output matches the rpms I've been generating and deploying in production for a couple months now.

### DIFF
--- a/rpm/install
+++ b/rpm/install
@@ -1,0 +1,13 @@
+%{__python} setup.py install --skip-build --root=$RPM_BUILD_ROOT
+
+%if ! (0%{?fedora} > 12 || 0%{?rhel} > 5)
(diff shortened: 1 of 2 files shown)


jgrund:
+1 I'm having this issue trying to run `python setup.py bdist_rpm` on centos 6.6. Not having this file breaks the command.

benoitc:
@jgrund sounds like a new issue, can you open a ticket describing the version you're using and the symptoms you experiment?

tilgovi:
Could be same issue. See my comment, though. The recommendation is to have your spec file pull the rest of the RPM packaging from github so we are not shipping packaging files for every system in our pypi bundle.

Issue #444, opened 2012-11-30
bind to multiple addresses


gvanrossum:
I have a WSGI app that I would like to be listening on multiple addresses. On my machine, 127.0.0.2 is a special interface; I want to listen on that interface as well as on 127.0.0.1 (the regular localhost). Command line syntax proposed:

--bind 127.0.0.1:8080,127.0.0.2:8080

The bind variable could be a string of that form or a list of strings, e.g.

bind = ['127.0.0.1:8080', '127.0.0.2:8080']

It should also be possible to use different ports on the same IP.

JenishGnanasicamani:
-b :<port1>,:<port2> is throwing error as Error: '8082,:8085' is not a valid port number.

tutorialcreation:
Hey Benoit am trying to bind a port but still giving me default port what could be the issue??..

tilgovi:
@tutorialcreation can you show us what you run on the command line? Do you have the `PORT` environment variable defined? Any other relevant details? You may also want to open a new issue, as your question seems not quite related to the issue here.

tutorialcreation:
Hello sorry for the late reply, here is my code
```
#!/bin/bash

NAME="post_retirement"                                 					# Name of the application
DJANGODIR=/srv/www/html/post_retirement          # Django project directory
SOCKFILE=localhost:8787   # we will communicte using this unix socket
USER=test_account                                       				# the user to run as
GROUP=test_account                                    				# the group to run as
NUM_WORKERS=3                                    				# how many worker processes should Gunicorn spawn
DJANGO_SETTINGS_MODULE=post_retirement.settings            				# which settings file should Django use
DJANGO_WSGI_MODULE=post_retirement.wsgi                    				# WSGI module name

echo "Starting $NAME as `whoami`"

# Activate the virtual environment
cd $DJANGODIR
source /srv/www/html/venv/bin/activate
export DJANGO_SETTINGS_MODULE=$DJANGO_SETTINGS_MODULE
export PYTHONPATH=$DJANGODIR:$PYTHONPATH

# Create the run directory if it doesn't exist
# RUNDIR=$(dirname $SOCKFILE)
# test -d $RUNDIR || mkdir -p $RUNDIR

# Start your Django Unicorn
# Programs meant to be run under supervisor should not daemonize themselves (do not use --daemon)
exec gunicorn ${DJANGO_WSGI_MODULE}:application \
 --name $NAME \
 --workers $NUM_WORKERS \
[...]

PR #448, opened 2012-12-11
allows gunicorn to bind to multiple address

benoitc:
Allows gunicorn to listen on different interface. It can be either ipv6,
unix or ipv4 sockets.

Ex:

```
$ gunicorn -w3 -b 127.0.0.1:8001 -b 127.0.0.1:8000 test:app
```

PR for to #444

### DIFF
--- a/examples/frameworks/pylonstest/test.ini
+++ b/examples/frameworks/pylonstest/test.ini
@@ -12,8 +12,7 @@ error_email_from = paste@localhost
 
 [server:main]
 use = egg:gunicorn#main
-host = 127.0.0.1
-port = 5000
+host = 127.0.0.1:8000, 127.0.0.1:8001
 workers = 2
 
 [app:main]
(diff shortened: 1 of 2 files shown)


gunicorn/arbiter.py
```diff
@@ -367,9 +374,11 @@ def reload(self):
 
         # do we need to change listener ?
         if old_address != self.cfg.address:
-            self.LISTENER.close()
-            self.LISTENER = create_socket(self.cfg, self.log)
-            self.log.info("Listening at: %s", self.LISTENER)
+            # close all listeners
+            [l.close for l in self.LISTENERS]
+            # init new listeners
+            self.LISTENERS = create_sockeSt(self.cfg, self.log)
```
megahall: typo?

tilgovi: Thanks, @megahall! Was fixed in 1c9b8460b14540b2679cf7fc57e77086dddf6b53

PR #471, opened 2013-01-10
support TCP Fast Open option

benoitc:
**Fast Open** is a optimization to the process of stablishing a TCP connection that
allows the elimination of one round time trip (RTT) from certain kinds of TCP
conversations. Fast Open could result in speed improvements of between 4% and
41% in the page load times on popular web sites.

Available on linux >= 3.7.1.

More info:

- https://lwn.net/Articles/508865/
- http://research.google.com/pubs/pub37517.html

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1292,3 +1292,33 @@ class SyslogFacility(Setting):
     desc = """\
     Syslog facility name
     """
+
+
+class AllowTCPFastOpen(Setting):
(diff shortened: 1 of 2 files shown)


benoitc:
@tilgovi didn't see the notice. Also not sure what you mean by "versionadded directive" ...

tilgovi:
@benoitc see "version added" in config.py docstrings for some options.

benoitc:
@tilgovi done. thanks!

PR #518, opened 2013-04-24
Attempt to get a stack trace from timed out workers before killing them

Jonty:
This changes the arbiter to send SIGUSR2 to the worker two seconds
before it sends SIGKILL, giving the new handle_usr2 handler time to
generate a clean stack trace of where the worker is blocked and log it.

Existing behaviour is maintained if the worker does not die, however the
log message has been changed to reflect that a stack trace was not
obtained.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -402,13 +402,20 @@ class Arbiter(object):
             return
         for (pid, worker) in self.WORKERS.items():
             try:
-                if time.time() - worker.tmp.last_update() <= self.timeout:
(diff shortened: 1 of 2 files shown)


antoine-lizee:
@benoitc can you explain why you chose to not pass the frame to the `worker_abort` in #768 ?

It prevents us from logging something like a stack trace that would be really useful to understand what the worker was doing that made him idling. The initial proposal here had something along those lines, and it promised to be really useful.

Is it because SIGABRT is not handled by sync workers in a useful manner?

PR #639, opened 2013-11-07
Feature/ipc

benoitc:
use a unix socket instead of a temportay file to signal activity

This change replace the use of a temportary file by a socket pair to
notify the master that the worker is still alive.

before we were doing the following:
- create a tmp file / worker
- write to it
- check the last update bey using os.stat

now:

- open a socket pair
- write to it
- on socket read we update the worker state

The advantage of this change is to remove any possible file leaking on some platform also it can possibly be faster since we don't have to read the file size each time.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -10,6 +10,7 @@ import os
 import random
 import select
 import signal
+import socket
 import sys
 import time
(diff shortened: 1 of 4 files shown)


benoitc:
closing. another cleaner path is on its way.

Issue #773, opened 2014-06-04
Dump requests and stack traces on SIGILL


tkornai:
This feature is intended to help analyse the state of gunicorn processes in a given point of time. This can come handy when dealing with performance issues.

A new signal handler for SIGILL would be introduced which triggers the status dump. When the signal is received workers dump their current requests with stack traces to the /tmp dir. If the signal is sent to the master it will broadcast it to its workers.

---

Example output:

[24458] Age: 13623, Thread: <Greenlet at 0x10669bb90: <functools.partial object at 0x10671c890>(<socket at 0x10672f0d0 fileno=11 sock=127.0.0.1:80, ('127.0.0.1', 55528))>, Request: {'wsgi.multiprocess': True, 'SERVER_SOFTWARE': 'gunicorn/19.0.0', 'SCRIPT_NAME': '', 'REQUEST_METHOD': 'GET', 'PATH_INFO': '/', 'SERVER_PROTOCOL': 'HTTP/1.0', 'QUERY_STRING': '', 'HTTP_USER_AGENT': 'ApacheBench/2.3', 'HTTP_CONNECTION': 'Keep-Alive', 'SERVER_NAME': '127.0.0.1', 'REMOTE_PORT': '55528', 'wsgi.url_scheme': 'http', 'SERVER_PORT': '8000', 'wsgi.input': <gunicorn.http.body.Body object at 0x10672f690>, 'HTTP_HOST': '127.0.0.1:8000', 'wsgi.multithread': True, 'HTTP_ACCEPT': '*/*', 'wsgi.version': (1, 0), 'RAW_URI': '/', 'wsgi.run_once': False, 'wsgi.errors': <gunicorn.http.wsgi.WSGIErrorsWraper object at 0x10672f7d0>, 'REMOTE_ADDR': '127.0.0.1', 'gunicorn.socket': <socket at 0x10672f0d0 fileno=11 sock=127.0.0.1:8000 peer=127.0.0.1:55528>, 'wsgi.file_wrapper': <class 'gunicorn.http.wsgi.FileWrapper'>}
Stack:   File "/usr/local/lib/python2.7/site-packages/gevent/greenlet.py", line 327, in run
    result = self._run(*self.args, **self.kwargs)
  File "/usr/local/Cellar/python/2.7.6_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/gunicorn-19.0.0-py2.7.egg/gunicorn/workers/async.py", line 47, in handle
[...]

PR #793, opened 2014-06-15
773-sigill-introspection

tkornai:
https://github.com/benoitc/gunicorn/issues/773

This feature is intended to help analyse the state of gunicorn processes in a given point of time. This can come handy when dealing with performance issues.

A new signal handler for SIGILL would be introduced which triggers the status dump. When the signal is received workers dump their current requests with stack traces to the /tmp dir. If the signal is sent to the master it will broadcast it to its workers.


benoitc:
closing since there are no activity since 2 years.

PR #874, opened 2014-09-06
Shutdown socket writes before close

### LINKED ISSUE #872: Connection not closed properly after responding with 401
Running gunicorn behind nginx.  Been having a strange issue where sometimes my server responds with a 502 Bad Gateway instead of a 401 Unauthorized.  (this is bad for me because someone else's legacy software is looking for a 401 before resubmitting request w/ username and password.)

After some tcpdump sleuthing, here's what I found:
Gunicorn is indeed sending a 401 Unauthorized response.  It sends it quickly; before nginx is finished sending the POST body in fact.  But Gunicorn also sends a RST packet to nginx right behind the 401 unauthorized message.  This causes a race between the OS and nginx.  It's a race to see if nginx can process the 401 message before the OS processes the RST packet and closes the socket for good.  If the RST packet gets processed first, when nginx goes to dig the 401 message out of the socket, the call errors out and lacking a response, nginx returns a 502 Bad Gateway.

Gunicorn should close the socket gracefully in this case, sending a fin packet instead of a RST packet.  This will allow nginx to dig the 401 Unauthorized response out of the socket even if it has already been closed.

I know this might be a tough one to test, if I can be of any more help please let me know.

tilgovi:
Calling shutdown(2) with SHUT_WR before close(2) ensures that a FIN
packet is sent before any RST packet. Doing so ensures that clients
and proxies get a clean shutdown signal when the server terminates the
connection while an upload is pending.

Fix #872

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -271,6 +271,7 @@ def set_non_blocking(fd):
 
 def close(sock):
     try:
+        sock.shutdown(socket.SHUT_WR)
         sock.close()
     except socket.error:
         pass


benoitc:
I was thinking about this issue and wonder if couldn't solve it with such design:

- in a linger pool all sockets are put in an event loop using the selector API
- each time we come back in the accept loop we check if a socket is available for read in the event loop, if yes we then definitely close it using the `shutdown` method followed by the `close` method.

I think it could work, but will add an extra time when accepting a connection. Maybe we should maintain a thread? Or just put the linger pool inside the arbiter?

Also how do we handle the gevent and eventlet workers? We have little control on them using their server object ( I wonder if we shouldn't provide our own server impl at some point...) . Maybe we could ignore them for the next release?

Let me know what you think about that design anyway.

benoitc:
superseded by #1025 all discussions now happen in that ticket.

PR #961, opened 2014-12-20
Close existing sockets in master process when stopping.

wooparadog:
Previously, when `gunicorn` is told to restart, it'll signal
workers to stop listening for new connections. But listening
ports are still open in master process. This lets load balancer,
such as haproxy, think that this server is still up and dispatch
requests to this server. But because all workers have stopped
accepting new connections. New requests will block until timeout.
This is especially severe when conducting a restart in a setup with
long graceful-timeout setting, causing many requests to timeout.

This patch fix that by closing all `LISTENERS` in `stop` method of
`Arbitor`.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -389,7 +389,8 @@ class Arbiter(object):
         # do we need to change listener ?
         if old_address != self.cfg.address:
             # close all listeners
-            [l.close() for l in self.LISTENERS]
+            for l in self.LISTENERS:
+                l.close()
             # init new listeners
             self.LISTENERS = create_sockets(self.cfg, self.log)
             self.log.info("Listening at: %s", ",".join(str(self.LISTENERS)))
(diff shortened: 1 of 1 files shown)


wooparadog:
Any followup about this please?

benoitc:
i didn't have the time to test it yet. Should be done tomorrow.

dsully:
+1 - would really like to see this fixed.

Thank you for the patch!

benoitc:
we need tests on that issue on the different platforms we support. If everything is OK let's merge it. So please if you can test this change and report the results here it will help :)

wooparadog:
@benoitc Could you give me a list of which tests need to be run please? I've only tested it in our prod. (ubuntu + centos, python 2.7).

benoitc:
i think this issue have been fixed. @tilgovi ?

tilgovi:
Yes, #1221

PR #962, opened 2014-12-20
stop to accept more requests when maximum accepted is achieved

### LINKED ISSUE #908: Threaded workers accept connections even when no ready thread exist
When using threads (with sync worker class), worker accepts connection even when all threads is busy. In our environment,  sometimes (because a deadlock in cairocffi library) all threads of a worker are locked, but worker continues accepting connections until limit of open-files is reached.

I looked at the code in `gthread.py` and it regards the `worker_connections` setting (despite what the documentation says, that it only affects Eventlet and Gevent workers).

I think in threaded workers, `worker_connections` always should be set to number of threads.

benoitc:
this change makes sure that a worker don't handle more requests than it can achieved.  The new workflow is quite more simple:

1. listeners are put in the poller. On read we try to accept on them.
2. When a connection is accepted it is put in the execution queue
3. When a request is done and the socket can be kept alived, we put it in the poller, on read event we will try to handle the new request. 
4. If it is not put out of the poller before the keepalive timeout the socket will be closed.
5. if all threads are busy we are waiting until one request complet. If it doesn't complete before the timeout we kill the worker.

fix #908

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -65,12 +65,13 @@ class TConn(object):
 
             # initialize the parser
             self.parser = http.RequestParser(self.cfg, self.sock)
-            return True
-        return False
 
     def set_timeout(self):
         # set the timeout
         self.timeout = time.time() + self.cfg.keepalive
+
+    def close(self):
+        util.close(self.sock)
 
     def __lt__(self, other):
         return self.timeout < other.timeout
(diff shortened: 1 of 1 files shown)


benoitc:
Anything to add/edit on that patch? If no-one disagree I will merge it in the day and prepare the next release. Let me know.

gunicorn/workers/gthread.py
```diff
@@ -83,68 +85,94 @@ class ThreadWorker(base.Worker):
...
-    def handle_client(self, conn, client):
-        # unregister the client from the poller
-        self.poller.unregister(client)
+    def reuse_connection(self, conn, client):
+        with self._lock:
+            # unregister the client from the poller
+            self.poller.unregister(client)
+            # remove the connection from keepalive
+            try:
+                self._keep.remove(conn)
```
benoitc: Maybe we could just tag the connection here instead of removing it from the queue and then locking. Then next time the connection is getting out from the que check if it can be used or not. 

Something like: `conn.removed = true`. 

Thoughts?

diwu1989:
O nice, this is finally fixed, can't wait for the next release to show up on pypi

PR #966, opened 2015-01-15
Fixed broken link to gafferctl in docs

alasdairnicol:
The gafferctl page was removed from the docs in benoitc/gaffer@00a4f629b0. This is the best replacement url I could find.

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -159,7 +159,7 @@ used to monitor gunicorn. A simple configuration is::
     cmd = gunicorn -w 3 test:app
     cwd = /path/to/project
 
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Thanks for the PR, but since gafferctl was removed, I think we should update the Gunicorn docs to mention `gaffer` instead: http://gaffer.readthedocs.org/en/latest/gaffer.html

alasdairnicol:
Thanks for the quick reply. 

I'm not familiar with gaffer to be honest, I was planning to deploy gunicorn with supervisor, and was browsing the other options when I spotted the broken link. 

I'll close this pull request and open an issue about updating the docs to mention gaffer.

berkerpeksag:
Great, thanks Alasdair! :)

PR #968, opened 2015-01-23
Update django_wsgi.py for django 1.7

cenkbircanoglu:


collinanderson:
django_wsgi.py is deprecated, right?

Shouldn't people just use the wsgi that comes with django?
https://github.com/django/django/blob/master/django/conf/project_template/project_name/wsgi.py

berkerpeksag:
Correct. Please use the `gunicorn` command instead of `gunicorn_django`. See http://docs.gunicorn.org/en/latest/run.html for more information.

PR #973, opened 2015-02-03
Fix import error for django 1.7

spanasik:
This is the fix for import error 'No module named django.core.management.validation'

Original post: http://www.snip2code.com/Snippet/81958/Possible-fix-to-gunicorn-s-Django-1-7-bu

### DIFF
--- a/gunicorn/app/django_wsgi.py
+++ b/gunicorn/app/django_wsgi.py
@@ -18,7 +18,6 @@ except:
 
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Thanks for patch, but `gunicorn_django` is deprecated and will be removed in Gunicorn 20.0. Please run `gunicorn` instead: http://docs.gunicorn.org/en/latest/run.html#gunicorn

PR #976, opened 2015-02-03
increase hard limit for headers

joycepg:
Kerberos tickets can be greater than 8190 bytes. This can occur when the ticket comes from Active Directory for a user who is a member of many groups. If the ticket is used as the value for the Authorization header, it will be rejected by gunicorn.
Authorization: Negotiate YIIYkAYGKwYBBQUCoIIYhDCCGICgMDAuBgkqhkiC9xIBAgIGCSqGSIb3EgECAg.....<another ~8500 bytes>
causes a 400 Bad Request, with Error parsing headers: 'limit request headers fields size'.
Workaround is to set --limit-request-field_size=0, but this disables that header checking functionality completely. 

I note that my solution also increases the default value. You may want to alter your code to retain 8190 as the default but allow the setting to set a value higher.

Also note that Microsoft have an IIS setting for similar reasons, and it can go to 64k max. see http://support.microsoft.com/kb/2020943

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -19,7 +19,7 @@ from gunicorn.six.moves.urllib.parse import urlsplit
 
 MAX_REQUEST_LINE = 8190
 MAX_HEADERS = 32768
-MAX_HEADERFIELD_SIZE = 8190
+MAX_HEADERFIELD_SIZE = 16384
 
 HEADER_RE = re.compile("[\x00-\x1F\x7F()<>@,;:\[\]={} \t\\\\\"]")
 METH_RE = re.compile(r"[A-Z0-9$-_.]{3,20}")


tilgovi:
Does setting limit-request-field-size higher not change the behavior?

joycepg:
no, because of this line https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/message.py#L45 and the next one.
it checks to see if the value set in configuration is greater than the max, and if so sets it back down.

tilgovi:
I feel like maybe we shouldn't be capping user supplied configurations. Users should increase it at their own risk, no?

benoitc:
@tilgovi Agree. Maybe we should add a warning when checking the config instead?

tilgovi:
That would be a great solution. @joycepg can you do that?

joycepg:
sure, i can remove the cap. what sort of warning? 
"value for limit-request-field-size is greater than the default value"?
btw: did you intend that the command line switch is limit-request-field_size (note dash,dash,underscore)? it's the only command line switch with an underscore.

benoitc:
@joycepg yes sych warning. the settings should probably be `limit-request-field-size` indeed...

tilgovi:
Bump.

benoitc:
done in latest master

PR #980, opened 2015-02-06
Make Logger.loglevel a class attribute.

berkerpeksag:
This way it will be easier to change log level of a custom
Logger class:

    class MyLogger(gunicorn.glogging.Logger):

        loglevel = logging.WARNING


benoitc:
+1

PR #986, opened 2015-02-22
Raise TypeError instead of AssertionError.

berkerpeksag:
assert statements will be removed if you run Python
in optimized mode (e.g. with -O flag).


benoitc:
+1

PR #990, opened 2015-02-27
Response to HEAD requests must not return body

tkornai:
The HTTP/1.1 specification states that: 'The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.'
(http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html)

Gunicorn currently does not comply with that as it sends a "0" in the body as the terminating chunk. This can be easily shown by a curl command against the echo:app example:

```
curl -v -I http://127.0.0.1:8000
* Rebuilt URL to: http://127.0.0.1:8000/
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 8000 (#0)
> HEAD / HTTP/1.1
> User-Agent: curl/7.37.1
> Host: 127.0.0.1:8000
> Accept: */*
>
< HTTP/1.1 200 OK
HTTP/1.1 200 OK
* Server gunicorn/19.2.1 is not blacklisted
< Server: gunicorn/19.2.1
Server: gunicorn/19.2.1
< Date: Fri, 27 Feb 2015 11:22:17 GMT
Date: Fri, 27 Feb 2015 11:22:17 GMT
< Connection: close
Connection: close
< Content-type: text/plain
Content-type: text/plain
< Content-Length: 14
Content-Length: 14
< X-Gunicorn-Version: 19.2.1
X-Gunicorn-Version: 19.2.1
< Test: test тест
Test: test тест

<
* Excess found in a non pipelined read: excess = 14 url = / (zero-length body)
* Closing connection 0
```

Note the line: "* Excess found in a non pipelined read: excess = 14 url = / (zero-length body)"

With my proposed fix the same curl command does not show the above warning.

I've found this issue when the nodejs http parser failed on paring gunicorn's responses on HEAD requests.

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -322,9 +322,9 @@ class Response(object):
 
     def write(self, arg):
         self.send_headers()
-        
+
         if self.req.method == "HEAD":
-            # HEAD request must not have body
+            # response to HEAD requests must not have body
             return
 
         if not isinstance(arg, binary_type):
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -322,6 +322,11 @@ class Response(object):
 
     def write(self, arg):
         self.send_headers()
+        
+        if self.req.method == "HEAD":
+            # HEAD request must not have body
+            return
+
         if not isinstance(arg, binary_type):
             raise TypeError('%r is not a byte' % arg)
         arglen = len(arg)


benoitc:
the gateway doesn't have to touch that normally. What is the application that is returing a value on head ?

tkornai:
Hi Benoit,

it is a Django 1.4 application. I understand your point that the app could handle properly HEAD requests and not return body content at all. That would require changing all the views or the introduction of a new middleware (e.g. ConditionalGetMiddleware). Do you see any reason why not enforce the right behaviour from gunicorn?

benoitc:
@tkornai  one of the reason is that the application can still try to send a body while gunicorn closed the socket or decided to keep alive. That's why it should be handled at the application level. In that case django should handle that imo....

tilgovi:
I would consider accepting a change that unwinds the body iterator but discards it. I'm with @benoitc that it concerns me a little bit that the application might try to send and fail. I'm not sure, though.

darkrain42:
Perhaps I'm reporting a different issue (though I don't think so), but I think this doesn't quite have anything to do with the backend application, and it's actually all behavior that gunicorn itself does.  I have a Django application served via Gunicorn w/ the async worker where I'm seeing the response to a `HEAD` request end up with the terminating chunk indicator (as described by @tkornai).

In my case, Django is _not_ returning content, and is not specifying a `Content-Length` or `Transfer-Encoding` (in my test scenario, Django is issuing a redirect).  From my reading of gunicorn/http/wsgi.py, that causes Gunicorn to decide [the response should be chunked (and thus add a T-E header)](https://github.com/benoitc/gunicorn/blob/ad07bccb4e582f6dcba0433115271824cc85fde9/gunicorn/http/wsgi.py#L279).

Later on, when the worker calls [resp.close()](https://github.com/benoitc/gunicorn/blob/ad07bccb4e582f6dcba0433115271824cc85fde9/gunicorn/workers/async.py#L114), the response object [appends the termination chunk indicator](https://github.com/benoitc/gunicorn/blob/ad07bccb4e582f6dcba0433115271824cc85fde9/gunicorn/http/wsgi.py#L421)

And I end up with an against-RFC response that contains a message-body (it's an empty message body, but a message body nonetheless).

My suggestion of how to fix this would be slightly different; instead of throwing away all (potential) response data from the app, have the response object alter its chunking/closing behavior:

```patch
diff --git a/gunicorn/http/wsgi.py b/gunicorn/http/wsgi.py
index 2ca2426..344b051 100644
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -227,6 +227,8 @@ class Response(object):
             return True
         if self.response_length is not None or self.chunked:
             return False
+        if self.req.method == 'HEAD':
+            return False
         if self.status_code < 200 or self.status_code in (204, 304):
             return False
         return True
@@ -284,6 +286,8 @@ class Response(object):
             return False
         elif self.req.version <= (1, 0):
             return False
+        elif self.req.method == 'HEAD':
+            return False
         elif self.status_code in (204, 304):
             # Do not use chunked responses when the response is guaranteed to
             # not have a response body.
```

I could see this causing _different_ problems if an application misbehaves and does start including a message body in its response, but those seem more theoretical.

tilgovi:
@darkrain42 that looks like a very sensible patch. I think it would be good if we did exactly that _and_ unwound any body iterator and discarded it. Would you be able to open a PR? You can start with just the patch you've posted.

benoitc:
IMO when the application try to send a body on HEAD request we should raises an error and return a 502  instead of silently discarding it. 

For the rest we should indeed makes sure we don't send any chunk at all, and the PR looks good for that purpose +1 to merge it.

tilgovi:
That sounds good. But what does it mean for the application? Return None?
An empty iterator? An empty string?

Some middleware might translate the return value to an iterator no matter
what. We should be tolerant.

tilgovi:
The issue of Gunicorn sending a terminating chunk when the application sends no response body is fixed by #1079.

It's hard to reason about the impact of doing anything further, so for now I will close this issue. Feel free to open a new discussion about additionally improvements that could be made.

Thank you, everyone.

tilgovi:
And since OP was using a Django application it's worth noting that Django has a filter applied to all responses, `django.http.conditional_content_removal`, that strips the body from a HEAD response.

So I believe #1079 should have fixed the issue described here.

PR #992, opened 2015-02-28
Delete unused test command.

berkerpeksag:
We are already using tox and requirements files to run
the test suite.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,10 +4,8 @@
 # See the NOTICE for more information.
 
 import os
-import sys
 
 from setuptools import setup, find_packages
-from setuptools.command.test import test as TestCommand
 
 from gunicorn import __version__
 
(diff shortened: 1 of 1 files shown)


tilgovi:
I understand that using tox (or running pytest directly) would be the way to run the test suite, but wouldn't this break `python setup.py test`? I don't know if that's a problem for downstream packagers.

berkerpeksag:
> but wouldn't this break python setup.py test

Correct, it will break `python setup.py test`.

> I don't know if that's a problem for downstream packagers.

That is a good point, but `python setup.py test` was already broken until recently fixed in https://github.com/benoitc/gunicorn/commit/2753aac65bd193ab29a70b4d992d955a2555daca. To be honest, I don't know anything about downstream packaging :) And of course I don't want to make the life harder for packagers.

tilgovi:
I have no strong opinion. I like how much cleaner setup.py gets. I would hope distro packagers can manage. @benoitc, have any thoughts?

benoitc:
hrm, I recon I don't use it myself but some I are still using it around. I would keep it if it harmless.

tilgovi:
I would keep it. It doesn't change. It's not that noisy. I think it is good to support `setup.py test` since that's well known.

berkerpeksag:
Ok, let's keep it :)

Issue #964, opened 2015-01-08
Reloader should survive SyntaxErrors


frabcus:
When using `--reload` if you save a file with a syntax error, gunicorn exits completely. This defeats the point and use of `--reload` during development.

The code is in `init_process` in `gunicorn/workers/base.py`. The callback for the Reloader kills the worker, forcing them to reload. The process exits when there is a syntax error.

Obvious ways to fix it would be: 1) Make the callback or reloader check the syntax of the Python files that have changed (with "compile") and output the syntax error and don't kill the process if they have a syntax error, 2) When in `--reload` mode catch SyntaxErrors in the process startup code.

[Flask-Failsafe](https://pypi.python.org/pypi/Flask-Failsafe/) handles a similar case for simpleserver, and might have useful code in.

ohadpartuck:
+1 any ETA on this @benoitc

GMLudo:
+1 for this feature.
We use Gunicorn in [API-Hour](www.api-hour.io/#api-hour) on prod and dev environment, it's certainly a nice to have.

tilgovi:
@ohadpartuck @GMLudo open a pull request

tilgovi:
Closed by #994

benoitc:
thanks I forgot that!

Issue #993, opened 2015-03-11
Support for custom error pages


aartur:
Looks like currently an error page content is hardcoded in util.write_error, as a simple html page. For some scenarios this needs to be changed, eg. an API service that should always return JSON.

Yes, custom error pages should be generated by a user app, but if an error happens before passing control to a user app (eg. a malformed HTTP header), the page is always generated by gunicorn.

aartur:
There's one specific, probably common case which triggers error page generation by gunicorn: when gunicorn is behind nginx, and a request contains a space in GET parameters, like:
```
$ curl 'http://site/?a=1 &'

<html>
  <head>
    <title>Bad Request</title>
  </head>
  <body>
    <h1><p>Bad Request</p></h1>
    Invalid HTTP Version 'Invalid HTTP Version: '&amp; HTTP/1.0''
  </body>
</html>
```

So nginx passes it to gunicorn without triggering an error, and gunicorn generates an error without calling wsgi app. Note also that the error message contains "Invalid HTTP Version" twice.

benoitc:
Reading the issue and the comment I am guessing we have 2 different issues. 1 is to use a different error page if needed and I would be OK for having a setting for it. The second is about parsing the status line. Though if I read the spec:

    5.1 Request-Line

    The Request-Line begins with a method token, followed by the Request-URI and the protocol version, and ending with CRLF. The elements are separated by SP characters. No CR or LF is allowed except in the final CRLF sequence.

            Request-Line   = Method SP Request-URI SP HTTP-Version CRLF

Gunicorn is correct in its behaviour. The URI shouldn't contains any space. If I would change a thing there it would be making the message more explicit telling to the user it failed to parse the request line. Thoughts? 

For the initial problem, i think it would be good to have it in next release.

aartur:
I only put the example with a space in a url to show that there are not-that-impossible situations when gunicorn generates an error page, even if a wsgi app has implemented full error handling and nginx's error pages are also set up. And yes, gunicorn's behavior is correct (and nginx's behaviour is incorrect if specs say that an invalid request must not be passed) and a better error message would be helpful.

For my purpose I monkey-patched util.write_error, but a supported solution would be nice.

jspeis:
+1

danidee10:
Any news on this? I'm also facing this issue on Django. My error pages showup when i run `django manage.py runserver` but if i run my app with gunicorn, It ignores the error pages i setup and displays it's own error page.

[...]

PR #994, opened 2015-03-11
reloader should survive SyntaxError

wong2:
This implements the feature requested in issue #964.

When there is `SyntaxError` raised while importing the app, provides a fallback app to display those errors instead of completely exit gunicorn.

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -541,6 +541,6 @@ def make_fail_app(msg):
             ("Content-Type", "text/plain"),
             ("Content-Length", str(len(msg)))
         ])
-        return iter([msg])
+        return [msg]
 
     return app
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -532,3 +532,15 @@ def warn(msg):
 
     print("!!!\n", file=sys.stderr)
     sys.stderr.flush()
+
+
+def make_fail_app(msg):
+
+    def app(environ, start_response):
+        start_response("500 Internal Server Error", [
+            ("Content-Type", "text/plain"),
+            ("Content-Length", str(len(msg)))
+        ])
+        return iter([msg])
+
+    return app
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -8,6 +8,7 @@ import os
 import signal
 import sys
 import time
+import traceback
 from random import randint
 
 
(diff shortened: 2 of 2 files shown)


wong2:
@benoitc plz review

berkerpeksag:
Couldn't this be handled in https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L89 or even in the Reloader class?

gunicorn/util.py
```diff
@@ -532,3 +532,15 @@ def warn(msg):
...
+
+
+def make_fail_app(msg):
+
+    def app(environ, start_response):
+        start_response("500 Internal Server Error", [
+            ("Content-Type", "text/plain"),
+            ("Content-Length", str(len(msg)))
+        ])
+        return iter([msg])
```
berkerpeksag: Why do you need to add `iter()` here? If I remember correctly, `[msg]` is enough in the WSGI spec.

wong2: oh I copied it from the example on http://gunicorn.org/ , and you're right, `[msg]` is enough

wong2:
@berkerpeksag I don't think this can be handled in `Reloader` or `reloader callback`, cause the `SyntaxError` is happened when the new worker is starting, while `Reloader` or `reloader callback` are in charge of terminating the old worker.

gunicorn/util.py
wong2: @berkerpeksag I've removed the `iter()` here

tilgovi:
This looks pretty good to me.

wong2:
@tilgovi what else should I do before this can be merged?

tilgovi:
I think it seems fine, I just want to give others a chance to look at it.

benoitc:
The patch works for me. But I would do the following changes:

- only display the traceback in debug mode so people using the reloader for another purpose won't have it displayed
- make the error template configurable related to #993

Thoughts?

berkerpeksag:
Perhaps it would be good to make the try-except logic a method. This way, people can customize make_fail_app etc.

gunicorn/workers/base.py
```diff
@@ -116,7 +117,19 @@ def changed(fname):
 
         self.init_signals()
 
-        self.wsgi = self.app.wsgi()
+        try:
+            self.wsgi = self.app.wsgi()
+        except SyntaxError as e:
+            if not self.reloader:
```
berkerpeksag: I'd use `self.cfg.reload`

tilgovi: +1

wong2:
@benoitc is there still a `debug mode` in gunicorn? I can't find it in the [doc](http://docs.gunicorn.org/en/latest/settings.html)

benoitc:
@wong2 no only `debug log level` (which provides the same informations) and `spew` setting.

wong2:
I've moved the try-catch codes to a new method `load_wsgi` so it's more customizable

tilgovi:
Looks great.

berkerpeksag:
Thanks @wong2 :)

georgexsh:
:beer:

PR #997, opened 2015-03-21
subclass Exception replace BaseException

s7v7nislands:
All user-defined exceptions should also be derived from Exception
https://docs.python.org/3/library/exceptions.html?highlight=baseexception#base-classes

### DIFF
--- a/gunicorn/errors.py
+++ b/gunicorn/errors.py
@@ -4,7 +4,7 @@
 # See the NOTICE for more information.
 
 
-class HaltServer(BaseException):
+class HaltServer(Exception):
     def __init__(self, reason, exit_status=1):
         self.reason = reason
         self.exit_status = exit_status
@@ -13,7 +13,7 @@ class HaltServer(BaseException):
         return "<HaltServer %r %d>" % (self.reason, self.exit_status)
 
 
-class ConfigError(BaseException):
+class ConfigError(Exception):
     """ Exception raised on config error """
 
 


berkerpeksag:
I think the BaseException usage was intentional: https://github.com/benoitc/gunicorn/commit/8d453fb341eceb13e4cb99e24844d198bda73391 However, I'd at least change ConfigError to be derived from Exception.

s7v7nislands:
@berkerpeksag what's the reason change in 
https://github.com/benoitc/gunicorn/commit/8d453fb341eceb13e4cb99e24844d198bda73391

berkerpeksag:
Well, I thought it was for to not catch HaltServer in gunicorn/arbiter.py explicitly, but I was wrong: https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L200 Unless there is a reason to use BaseException, the patch LGTM.

benoitc:
If I remember well the reason was to make sure  it is not accidentally caught by code. There was also something in python 3 vs python2 but I can't recollect right now.

benoitc:
So to be clear `BaseException` is used there in the same way it's used on low level modules in gunicorn. I would keep it for HaltServer then. Config I don't know but also don't really care about it. Does someone else think that change is needed?

tilgovi:
This is the typical style. We should change it to be `Exception` unless there's a reason it won't work.

berkerpeksag:
> If I remember well the reason was to make sure it is not accidentally caught by code.

Actually, this is a good reason. +1 for keeping HaltServer as it is and changing other instances to inherit from Exception instead of BaseException.

tilgovi:
:+1: @s7v7nislands if you want to update the PR we can merge just the `ConfigError` change.

tilgovi:
Maybe add a comment to the `HaltServer` that says that we don't inherit from `Exception` to avoid having it be caught by application code.

benoitc:
@tilgovi good idea. just did it :)

PR #998, opened 2015-03-22
remove no need import

s7v7nislands:
remove no need import in the run() function


berkerpeksag:
The imports are there because a) run() functions are used as setuptools entry points b) to avoid circular import problems.

s7v7nislands:
Thank you for review!

Issue #1002, opened 2015-04-04
Error Gunicorn 19.3.0 with Django 1.7 (import error get_validation_errors)


allanfigueiredo:
Traceback (most recent call last):
  File "/Users/allanfigueiredo/.virtual_envs/pfci-django/lib/python2.7/site-packages/gunicorn/arbiter.py", line 507, in spawn_worker
    worker.init_process()
  File "/Users/allanfigueiredo/.virtual_envs/pfci-django/lib/python2.7/site-packages/gunicorn/workers/base.py", line 118, in init_process
    self.wsgi = self.app.wsgi()
  File "/Users/allanfigueiredo/.virtual_envs/pfci-django/lib/python2.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/Users/allanfigueiredo/.virtual_envs/pfci-django/lib/python2.7/site-packages/gunicorn/app/djangoapp.py", line 105, in load
    mod = util.import_module("gunicorn.app.django_wsgi")
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/Users/allanfigueiredo/.virtual_envs/pfci-django/lib/python2.7/site-packages/gunicorn/app/django_wsgi.py", line 21, in <module>
    from django.core.management.validation import get_validation_errors
ImportError: No module named validation

berkerpeksag:
Did you use `gunicorn_django` to run your application? If yes, please use the `gunicorn` command instead: http://docs.gunicorn.org/en/19.3/run.html#gunicorn

allanfigueiredo:
Hi Berker,

Yes I'm using gunicorn_django. Ok sorry about the little confusion. I'll
try to use gunicorn instead of gunicorn_django.

Thanks so much!

All the best

berkerpeksag:
Thanks for the quick response :) Closing this.

PR #1003, opened 2015-04-04
Adding support for Django 1.7 (fixing #1002)

allanfigueiredo:
Recently I had problems to run my Django 1.7 application with Gunicorn in the latest version. Verifying the error is it about an import that was deleted in Django 1.7. I made a correction and my application is running and i like to share the solution with you guys. Thanks a lot.

### DIFF
--- a/gunicorn/app/django_wsgi.py
+++ b/gunicorn/app/django_wsgi.py
@@ -18,7 +18,19 @@ except:
 
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Thanks for the PR, but `gunicorn_django` is deprecated. Please use `gunicorn` (see http://docs.gunicorn.org/en/19.3/run.html#gunicorn) instead.

PR #1004, opened 2015-04-08
Retry socket binding up to five times upon EADDRNOTAVAIL.

ryanpetrello:
I'm using gunicorn in an embedded virtual networking application, and it's not uncommon for the bind address to take a few moments to actually be added to the interface.  Given that other socket binding failures in this method lead to a retry loop, it seems reasonable that this one would as well?


tilgovi:
Seems reasonable.

berkerpeksag:
+1

berkerpeksag:
Thanks!

PR #1007, opened 2015-04-14
Add pyvenv and IntelliJ IDEA files in .gitignore

GMLudo:
Hi,

To avoid to commit wrong files on my laptop, this .gitignore file will help me to avoid mistakes.
Thanks for the merge.

Regards.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.pyc
 *#*
 venv*
+pyvenv*
 __pycache__
 MANIFEST
 build
@@ -20,3 +21,5 @@ examples/frameworks/pylonstest/PasteScript*
 examples/frameworks/pylonstest/pylonstest.egg-info/
 examples/frameworks/django/testing/testdb.sql
 .tox
+.idea/
+*.iml


berkerpeksag:
Thanks for the patch, but I think it's better to keep editor related rules in your global .gitignore :)

GMLudo:
@berkerpeksag thanks for the tip of global .gitignore, I didn't know.
Nevertheless, it should be cool for new comers to have that, others projects have that like [AsyncIO](https://github.com/python/asyncio/blob/master/.gitignore#L17) or [aiohttp](https://github.com/KeepSafe/aiohttp/blob/master/.gitignore#L14) (Yes, it's also me who I've added that ;-) )

tilgovi:
Agree with @berkerpeksag. I also learned about global ignores recently after making a similar request on another project. It's definitely more maintainable to have each person manage their own global ignores and have the project only ignore its own artifacts.

benoitc:
ignoring `pyvenv` seems OK imo

tilgovi:
`pyvenv` command from Python 3 requires a directory argument. It's not necessarily called `pyvenv`.

GMLudo:
@tilgovi you're right, nevertheless, it's like venv
it's a manner to identify virtualenvs made by venv and pyvenv, because you don't have the same content inside, at least from my point of view, pyvenv content it's cleaner.

tilgovi:
I think venv shouldn't be there, either :).

GMLudo:
Come on, @tilgovi, it isn't a big change for you and it will simplify life for the simple developers like me ;-)

[corruption] If I give you a [Westvleteren beer](http://en.wikipedia.org/wiki/Westvleteren_Brewery) at the next EuroPython, could you merge this pull request ? [/corruption]

FYI, Westvleteren beer is one of most famous Belgian beer, and almost Belgian people never drink this beer because it's almost impossible to buy it: [you need to call during a specific schedule with several limitations](http://en.wikipedia.org/wiki/Westvleteren_Brewery#Availability).

But, as a Telco guy, I've some tips&tricks to call efficiently their land line, I've an interesting stock ;-)
(Yes, API-Hour isn't always used for professional reasons ;-) )

BTW, @benoitc, if you're interested in, I'll give you the next time we will see.

tilgovi:
I don't much care either way and I wouldn't throw a fit if someone merged this, I'm just trying to share knowledge about things like global ignores so that you don't feel that every project you try to work on is complicating your life because they don't have ignore rules for your editor of choice.

GMLudo:
@tilgovi: it was an attempt to do a spirit joke to be funny. I'm sorry if my proposal could offense you, sincerely, it wasn't the goal.
At least to me, the discussion is over.

tilgovi:
No offense! Thanks.

benoitc:
@tilgovi right ;) 

So if we follow that rule let's clean the `.gitignore` :)

tilgovi:
@GMLudo I would still gladly have a beer with you :)

GMLudo:
@tilgovi: deal ;-)

PR #1010, opened 2015-04-20
Fix statsd Logging in Python 3

krallin:
`statsd` logging appears to be (silently) broken in Python 3, because it passes it passes `unicode` (i.e. not `bytes`) to the `sock.send` method.

This PR:

  + Ensures `bytes` are used
  + Adds tests that actually go through an actual (UNIX) socket to ensure those errors can be caught in uni testing in the future.
  + Logs `statsd` errors as warnings (the exceptions being silently swallowed made them difficult to troubleshoot)

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -224,8 +224,8 @@ class Logger(object):
     def debug(self, msg, *args, **kwargs):
         self.error_log.debug(msg, *args, **kwargs)
 
-    def exception(self, msg, *args):
-        self.error_log.exception(msg, *args)
+    def exception(self, msg, *args, **kwargs):
+        self.error_log.exception(msg, *args, **kwargs)
 
     def log(self, lvl, msg, *args, **kwargs):
         if isinstance(lvl, string_types):
--- a/gunicorn/instrument/statsd.py
+++ b/gunicorn/instrument/statsd.py
@@ -8,7 +8,9 @@
 import socket
 import logging
 from re import sub
+
 from gunicorn.glogging import Logger
+from gunicorn import six
 
 # Instrumentation constants
 STATSD_DEFAULT_PORT = 8125
@@ -82,7 +84,7 @@ class Statsd(Logger):
             if msg is not None and len(msg) > 0:
                 Logger.log(self, lvl, msg, *args, **kwargs)
         except Exception:
-            pass
+            Logger.warning(self, "Failed to log to statsd", exc_info=True)
 
     # access logging
     def access(self, resp, req, environ, request_time):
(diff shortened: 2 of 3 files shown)


gunicorn/glogging.py
```diff
@@ -224,8 +224,8 @@ def info(self, msg, *args, **kwargs):
     def debug(self, msg, *args, **kwargs):
         self.error_log.debug(msg, *args, **kwargs)
 
-    def exception(self, msg, *args):
-        self.error_log.exception(msg, *args)
+    def exception(self, msg, *args, **kwargs):
```
berkerpeksag: Good catch!

gunicorn/instrument/statsd.py
```diff
@@ -98,29 +100,24 @@ def access(self, resp, req, environ, request_time):
...
-            if self.sock:
-                self.sock.send("{0){1}:-{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
-        except Exception:
-            pass
+        self._sock_send("{0){1}:-{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
 
     def histogram(self, name, value):
+        self._sock_send("{0}{1}:{2}|ms".format(self.prefix, name, value))
+
+    # Internal methods
```
berkerpeksag: I'd remove this comment since `_spam` names are considered private/internal.

tests/test_010-statsd.py
```diff
@@ -24,7 +31,23 @@ def __init__(self, failp):
...
+        server.bind(sock_file)
+
+        client = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+        client.connect(sock_file)
+
+        client.send(msg)
+
+        self.msgs.append(server.recv(1024))
+
+        client.close()
```
berkerpeksag: Can you wrap these three cleanup lines with a `try...finally` clause?

krallin:
@berkerpeksag ,

Fixed the two issues you pointed out. Note that if `server.bind(sock_file)` or `client.connect(sock_file)` fail, the tests might end up closing a socket that was never open, but that shouldn't raise an exception anyway.

Cheers,

berkerpeksag:
LGTM (thanks for squashing commits by the way!).

My only question is that if we test the statsd logging against a real socket, shouldn't we use a different class since the class' name is "MockSocket"? In theory, we should be able to test whether the input is str or not by tweaking the MockSocket.send() method (without creating a real socket).

Ping @benoitc @tilgovi.

krallin:
@berkerpeksag 

The idea behind using an actual socket is to catch similar errors in unit tests in the future (if they can even exist). 

This time, it was a text vs bytes issue, but maybe another issue will crop up some day (I have no idea what that issue might be, but after all that's why we write tests!). 

I'm fine with renaming the class, though — let me know!

Cheers,

gunicorn/instrument/statsd.py
```diff
@@ -98,29 +100,22 @@ def access(self, resp, req, environ, request_time):
...
-            pass
+        self._sock_send("{0){1}:-{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
 
     def histogram(self, name, value):
+        self._sock_send("{0}{1}:{2}|ms".format(self.prefix, name, value))
+
+    def _sock_send(self, msg):
         try:
+            if isinstance(msg, six.text_type):
+                msg = msg.encode("ascii")
```
benoitc: why not reusing the `bytes_to_str` function from the `_compat` module instead?

krallin: @benoitc 

I think we'd need `str_to_bytes` instead, wouldn't we? 

Looking at the definition for `bytes_to_str`, I'm a bit unclear on what it's supposed to returns. It looks like it gives you bytes on Py2 and text Py3? I'm probably misreading though.

Cheers,

benoitc: mmm you're right should probably be the reverse in your case. anyway could be useful to have it in a `_compat` module, so we don't have to do such things on python 2.x. thoughts?

krallin: Makes sense. Can you just confirm what those methods should return to clear up the confusion above?

berkerpeksag: If we are going to add a function like this, we need to make it more general(e.g. you can pass different encodings and error handlers). util.to_bytestring is another alternative, but it uses the UTF-8 encoding by default. I'd say merge this now and create a new issue or PR to discuss details.

benoitc: +1 let's merge now and handle the rest separately.

krallin:
Do you guys need me to rebase or squash further or anything? Just let me know : )

berkerpeksag:
@krallin if you have time, please squash the commits. If not, I can do it today :) Thanks!

krallin:
@berkerpeksag Sure — done

berkerpeksag:
Thanks!

PR #1016, opened 2015-04-23
notify systemd about main pid

urbaniak:
Send notification to systemd about the new master pid.

It's required when re-execing gunicorn (using `ExecReload=/bin/kill -SIGUSR2 $MAINPID`). systemd gets confused about the main pid when the previous master is still alive finishing requests and scheduling next reload.


urbaniak:
closing now, cause found some issues, will re-open it when it's ready

PR #1018, opened 2015-04-28
Fix argument to disable sendfile

### LINKED ISSUE #856: gunicorn dumps garbage when sendfile is enabled on virtualbox
Solution: Please add configuration option to disable sendfile support in Gunicorn 

Problem:
This occurs when running on virtual box, any ubuntu > 9, and is a known issue. The workaround for about 3 years seems to be to tell your server Nginx or Apache to disable sendfile support. Gunicorn under the covers uses sendfile and there should be a configuration option to disable it and fall back to a less performant solution.

Here are some related tickets:
https://github.com/mitchellh/vagrant/issues/351#issuecomment-1339640
https://www.virtualbox.org/ticket/9069

Here are some SO posts with references:
http://stackoverflow.com/questions/21422426/i-add-a-line-of-text-to-my-css-file-garbage-comes-through-the-browser
http://serverfault.com/questions/269420/disable-caching-when-serving-static-files-with-nginx-for-development

Random blogs post on it:
https://coderwall.com/p/ztskha
http://abitwiser.wordpress.com/2011/02/24/virtualbox-hates-sendfile/

toby-sullivan-hs:
The --sendfile parameter which is supposed to allow disabling use of sendfile doesn't actually accomplish that. Here we replace with --no-sendfile which works as expected.

Fix #856 for realz.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -828,6 +828,8 @@ class Sendfile(Setting):
         Disables the use of ``sendfile()``.
 
         .. versionadded:: 19.2
+        .. versionchanged:: 19.4
+            Swapped --sendfile with --no-sendfile to actually allow disabling
         """
 
 class Chdir(Setting):
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -820,12 +820,12 @@ class PreloadApp(Setting):
 class Sendfile(Setting):
     name = "sendfile"
     section = "Server Mechanics"
-    cli = ["--sendfile"]
+    cli = ["--no-sendfile"]
     validator = validate_bool
-    action = "store_true"
+    action = "store_false"
     default = True
     desc = """\
-        Enables or disables the use of ``sendfile()``.
+        Disables the use of ``sendfile()``.
 
         .. versionadded:: 19.2
         """


tilgovi:
LGTM

georgexsh:
add a changelog entry?

gunicorn/config.py
```diff
@@ -820,12 +820,12 @@ class PreloadApp(Setting):
...
+    cli = ["--no-sendfile"]
     validator = validate_bool
-    action = "store_true"
+    action = "store_false"
     default = True
     desc = """\
-        Enables or disables the use of ``sendfile()``.
+        Disables the use of ``sendfile()``.
 
         .. versionadded:: 19.2
```
berkerpeksag: Can you add a versionchanged directive and describe the change there. Something like:

```rst
.. versionchanged:: 19.4
   Foo bar
```

Also, perhaps we need to keep `--sendfile` as an alias to `--no-sendfile` for a few release.

toby-sullivan-hs: Will add a versionchanged directive. 

To address second suggestion, --sendfile never actually did _anything_ as far as I can tell. So unless someone can demonstrate otherwise, I don't see a benefit to keeping it around at all.

toby-sullivan-hs: Added change directive.

berkerpeksag:
Looks good, thanks! Can you please squash the commits?

toby-sullivan-hs:
No. I don't have time for this. Either take it as is or close the pull request.

tilgovi:
@toby-sullivan-hs I've done this for you[1].

That was <1m for me to do. If you need help doing this in the future, please let me know and I'll be happy to walk you through it.

Thank you very much for the patch.

[1] https://github.com/benoitc/gunicorn/commit/5c98b3deb23e0e0f21bdaebfb2a1e33acdef6c9b

PR #1019, opened 2015-04-29
design.rst: grammar

mgalgs:
Currently, we introduce this document as a description
of "an" architecture of Gunicorn, which makes it sound
like this document describes one of many Gunicorn
architectures, which doesn't make a lot of sense.

Clarify the grammar by introducing the document as a
description of "the" Gunicorn architecture.

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -5,7 +5,7 @@
 Design
 ======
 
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Thanks! :)

PR #1027, opened 2015-05-12
clean gitignore

benoitc:
people should be encouraged to use a global gitignore file.
https://help.github.com/articles/ignoring-files/

related to #1007

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,17 +1,10 @@
-*.gem
-*.sw*
 *.pyc
-*#*
-venv*
 __pycache__
 MANIFEST
 build
 dist
 setuptools-*
-.svn/*
-.DS_Store
 *.so
-.Python
 *.egg
 *.egg-info
 nohup.out


berkerpeksag:
*.gem, *#*, .svn/, .DS_Store and .Python can be removed too.

benoitc:
@berkerpeksag  removed other patterns (also rebased the changes). Should be be OK now, let me know.

berkerpeksag:
LGTM (not sure what setuptools-* is for, but I think it can be removed too)

benoitc:
can't remember, but think it was needed during testing ... let's keep it
for now :)

GMLudo:
Don't forget the main goal of `.gitignore` is to avoid that somebody makes a pull request with wrong/personal content.
I'm not a Mac user, however, it should be newcomer friendly to keep `.DS_Store` to avoid mistakes from people that don't aware about global gitignore.

benoitc:
I would rather solve that in a short developer documentation. would it be
enough?

tilgovi:
We'll find out if we get PRs from people with .DS_Store files. :)

benoitc:
indeed :)

PR #1031, opened 2015-05-15
Fix self.nr usage in ThreadedWorker so that auto restart works as expected

### LINKED ISSUE #1030: Auto restart is not working with gunicorn.workers.gthread.ThreadWorker
Hi,

Using a slightly modified test application so that I can see what is
happening:

    print '--- started'

        print '* handling request'
        data = 'Hello, World!\n'
        status = '200 OK'
        response_headers = [
            ('Content-type','text/plain'),
        ]
        return iter([data])

and running the gunicorn with

    gunicorn --workers=1 --threads=2 --max-requests=2 test:app

I would expect the worker to be restarted after two requests. This does not happen:

    $ gunicorn --workers=1 --threads=2 --max-requests=2 test:app
    [2015-05-15 23:47:19 +0000] [2532] [INFO] Starting gunicorn 19.3.0
    [2015-05-15 23:47:19 +0000] [2532] [INFO] Listening at: http://127.0.0.1:8000 (2532)
    [2015-05-15 23:47:19 +0000] [2532] [INFO] Using worker: threads
    [2015-05-15 23:47:19 +0000] [2537] [INFO] Booting worker with pid: 2537
    --- started
    * handling request
    * handling request
    * handling request
    * handling request
    * handling request
    * handling request

Expected result: worker restarted after max-requests
Observer result: worker is not restarted

I think the problem is that `self.nr` is used in the `ThreadedWorker` to
count the number of active connections instead the number of reqeuests
the worker has seen. The counter is increased in the `accept()` and
decreased in the `murder_keepalived()`. In the `run()` method there is a condition:

    if self.nr < self.worker_connections:

but at the same time in `handle_request()`:

    if self.alive and self.nr >= self.max_requests:
        self.log.info("Autorestarting worker after current request.")

so the worker would actually restart if there were more than
`self.max_requets` number of concurrent requests?

In `async.py` and `sync.py` the `self.nr` is used for counting the handled
requests:

    self.nr += 1
    if self.alive and self.nr >= self.max_requests:
        self.log.info("Autorestarting worker after current request.")

I think that the `ThreadedWorker` should do the same and use another
variable for counting the concurrent connections.

Version information:

	bash-4.2$ gunicorn --version
	gunicorn (version 19.3.0)
	bash-4.2$ python --version
	Python 2.7.5
	bash-4.2$ pip freeze
	futures==3.0.2
	-e git+https://github.com/benoitc/gunicorn.git@cc7459d953856d61cd810c1359f78796d181d902#egg=gunicorn-master
	pluggy==0.3.0
	py==1.4.27
	tox==2.0.1
	virtualenv==12.1.1

jhekkanen:
Use `nr_conns` for counting the connections and `self.nr` for counting the requests.

Closes #1030

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -92,6 +92,7 @@ class ThreadWorker(base.Worker):
         self._lock = None
         self.futures = deque()
         self._keep = deque()
+        self.nr_conns = 0
 
     @classmethod
     def check_config(cls, cfg, log):
@@ -131,7 +132,7 @@ class ThreadWorker(base.Worker):
             client, addr = listener.accept()
             # initialize the connection object
             conn = TConn(self.cfg, listener, client, addr)
-            self.nr += 1
+            self.nr_conns += 1
             # enqueue the job
             self.enqueue_req(conn)
         except socket.error as e:
@@ -170,7 +171,7 @@ class ThreadWorker(base.Worker):
                     self._keep.appendleft(conn)
                 break
             else:
-                self.nr -= 1
+                self.nr_conns -= 1
                 # remove the socket from the poller
                 with self._lock:
                     try:
@@ -202,7 +203,7 @@ class ThreadWorker(base.Worker):
             self.notify()
 
             # can we accept more connections?
-            if self.nr < self.worker_connections:
+            if self.nr_conns < self.worker_connections:
                 # wait for an event
                 events = self.poller.select(0.02)
                 for key, mask in events:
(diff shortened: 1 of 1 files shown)


benoitc:
thanks!

kenrobbins:
Hi, will this be in a release soon? Stumbled on this issue when going from 19.1.1 to 19.3.0. Thanks.

sseg:
Though I couldn't find notice of it in the changelog, this patch is at least included in version 19.4.5

Issue #1032, opened 2015-05-19
graceful shutdown for tornado is not thread-safe.


ccl0326:
ERROR:tornado.application:Exception in callback None
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/tornado/ioloop.py", line 840, in start
    handler_func(fd_obj, events)
  File "/Library/Python/2.7/site-packages/tornado/stack_context.py", line 275, in null_wrapper
    return fn(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/tornado/netutil.py", line 211, in accept_handler
    connection, address = sock.accept()
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
    return getattr(self.sock, name)
  File "/Library/Python/2.7/site-packages/gunicorn-19.3.0-py2.7.egg/gunicorn/sock.py", line 37, in __getattr__
[...]

PR #1033, opened 2015-05-19
make graceful shutdown thread-safe

ccl0326:


benoitc:
+1 can anyone else review it? So far looks good for a merge

benoitc:
how can I reproduce the issue?

benoitc:
Sorry I didn't see the related issue #1032 . looking at it.

gunicorn/workers/gtornado.py
```diff
@@ -50,10 +49,27 @@ def watchdog(self):
             self.log.info("Parent changed, shutting down: %s", self)
             self.stop()
```
fatelei: `stop` method is not defined. You have remove the method `stop`.

ccl0326:
@fatelei Thanks.I'm so sorry for missed it out.Parent changed and Auto-restart has been tested.

gunicorn/workers/gtornado.py
```diff
@@ -48,12 +46,29 @@ def watchdog(self):
 
         if self.ppid != os.getppid():
             self.log.info("Parent changed, shutting down: %s", self)
-            self.stop()
+            self.alive = False
+
+    def heartbeat(self):
+        if not self.alive:
+            if self.server_alive:
+                if hasattr(self, 'server'):
```
berkerpeksag: Can't we write this like the following snippet? (untested)

```py
if hasattr(self, 'server'):
    try:
        self.server.stop()
    except Exception:
        pass

self.server_alive = False
```

benoitc: indeed, if the server doesn't exist we need to stop the worker anyway.

ccl0326: The exception is ignored,but in fact we don't know how to deal with it.LGTM.

berkerpeksag:
LGTM, thanks!

Two minor comments:

* Could you please squash the commits?
* Please delete the empty lines in gunicorn/workers/gtornado.py:

  ```diff
-# -*- coding: utf-8 -
+
+
   ```

ccl0326:
done.

PR #1041, opened 2015-05-29
empty ACCEPT header support

mkorenkov:
Some of our legacy clients stopped working ffter we upgraded from _18.0_ to _19.3_.
As we find out, the reason was they were sending an empty `Accept` header. 
Some authors recommend replacing treating empty `Accept:` as `Accept: */*`([link](http://stackoverflow.com/questions/12130910/how-to-interpret-empty-http-accept-header)).
This fix is treating `Accept:` as `Accept: */*`.

**Before**:

    curl -i 'https://host.example.com/' -H 'Accept:'

>HTTP/1.1 500 Internal Server Error UNKNOWN STATUS CODE
>Server: nginx/1.8.0
>Date: Thu, 28 May 2015 14:56:20 GMT
>Content-Type: text/html; charset=utf-8
>Transfer-Encoding: chunked
>Connection: keep-alive

log:
>2015-05-28 07:56:20 +0000] [19480] [ERROR] Error handling request
>Traceback (most recent call last):
>  File ".../env/local/lib/python2.7/site-packages/gunicorn/workers/sync.py", line 177, in handle_request
>    resp.write(item)
>  File ".../env/local/lib/python2.7/site-packages/gunicorn/http/wsgi.py", line 326, in write
>    raise TypeError('%r is not a byte' % arg)
>TypeError: u'The server has either erred or is incapable of performing\r\nthe requested operation.\r\n<br><br>\nNone\nNone' is not a byte

**After**:

    curl -i 'http://127.0.0.1:8000' -H 'Accept:' 

>HTTP/1.1 302 FOUND
>Server: gunicorn/19.3.0
>Date: Fri, 29 May 2015 04:22:28 GMT
>Connection: close
>Transfer-Encoding: chunked
>Vary: Cookie
>Content-Type: text/html; charset=utf-8
>Location: http://127.0.0.1:8000/user/login/?next=/

    curl -i 'http://127.0.0.1:8000' -H 'Accept: */*' 

>HTTP/1.1 302 FOUND
>Server: gunicorn/19.3.0
>Date: Fri, 29 May 2015 04:22:32 GMT
>Connection: close
>Transfer-Encoding: chunked
>Vary: Cookie
>Content-Type: text/html; charset=utf-8
>Location: http://127.0.0.1:8000/user/login/?next=/

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -88,7 +88,8 @@ def default_environ(req, sock, cfg):
         "REQUEST_METHOD": req.method,
         "QUERY_STRING": req.query,
         "RAW_URI": req.uri,
-        "SERVER_PROTOCOL": "HTTP/%s" % ".".join([str(v) for v in req.version])
+        "SERVER_PROTOCOL": "HTTP/%s" % ".".join([str(v) for v in req.version]),
+        "HTTP_ACCEPT": "*/*"
     })
     return env
 
@@ -144,6 +145,9 @@ def create(req, sock, client, server, cfg):
             continue
         elif hdr_name == "CONTENT-LENGTH":
             environ['CONTENT_LENGTH'] = hdr_value
+            continue
+        elif hdr_name == "ACCEPT":
+            environ['ACCEPT'] = hdr_value
             continue
 
         key = 'HTTP_' + hdr_name.replace('-', '_')


benoitc:
Thanks for the patch. However I'm inclined to close it as **WONT_FIX**. 

Indeed it's not the responsibility of the server to set the default headers and if the `Accept` is given to the server it should already be present as `HTTP_ACCEPT`in the WSGI environment. If the purpose to set by default to `*.*` the case where the header is not given should be handled by the application itself (since no header exactly mean `*.*`).. If you really want to set it to `*.*` I would suggest to use a WSGI middleware then.

What others think about it?

tilgovi:
I can't tell from the traceback whether Gunicorn _can_ respond to requests with an empty Accept header (and it's just the framework or application that is struggling) or if it's impossible and Gunicorn errors. We should at least pass an empty header through to the application without choking, I think.

mkorenkov:
@benoitc @tilgovi gunicorn responds to the request with HTTP 500 with exception. 
I can move the fix to the app logic (I hope), but in my opinion gunicorn code should not raise an exception in that case. I would vote for not setting environment (ignoring) header if it has no header value.

benoitc:
Hrm without any patch on the latest master with python 3.4x or python 2.7x  gunicorn is able to accept an empty header:

screen shot 2015-05-29 at 23 18 20

The 500 error is most likely due to this error:

    File ".../env/local/lib/python2.7/site-packages/gunicorn/http/wsgi.py", line 326, in write
    raise TypeError('%r is not a byte' % arg)
    TypeError: u'The server has either erred or is incapable of performing\r\nthe requested operation.\r\n

Which means that you're probably returning an unicode to the response. Could you check that?

Looking at the [spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1) we can read the following:

> If no Accept header field is present, then it is assumed that the client accepts all media types.

For me it seems that Gunicorn is doing the appropriate job there. As a gateway it shouldn't change the request headers also in my opinion.

tilgovi:
Bump. @mkorenkov can you confirm that gunicorn handles the request fine but the error you experience is due to returning a non-byte body?

benoitc:
@mkorenkov bump

mkorenkov:
Sorry, folks, I was pretty swamped with work. Looks like we had the issue somewhere down the stack. In any case, the problem is gone:

    $pip freeze
    Django==1.8.4
    gunicorn==19.3.0
    wsgiref==0.1.2

    curl -i 'http://127.0.0.1:8000/' -H 'Accept:'
    HTTP/1.0 200 OK

PR #1042, opened 2015-06-04
Updated gunicorn/pidfile.py

brijeshb42:
Added a try block to convert the contents of the `pidfile.py` if it can be converted to integer or not. If not, then `wpid` is assigned 0 and the `validate` function returns as usual. If the pidfile has a valid int, the function continues to kill the process as before.

### DIFF
--- a/gunicorn/pidfile.py
+++ b/gunicorn/pidfile.py
@@ -67,11 +67,8 @@ class Pidfile(object):
         try:
             with open(self.fname, "r") as f:
                 try:
-                    wpid = int(f.read())
+                    wpid = int(f.read() or 0)
                 except ValueError:
-                    wpid = 0
-
-                if wpid <= 0:
                     return
 
                 try:
--- a/gunicorn/pidfile.py
+++ b/gunicorn/pidfile.py
@@ -66,7 +66,10 @@ class Pidfile(object):
             return
         try:
             with open(self.fname, "r") as f:
-                wpid = int(f.read() or 0)
+                try:
+                    wpid = int(f.read())
+                except ValueError:
+                    wpid = 0
 
                 if wpid <= 0:
                     return
--- a/gunicorn/pidfile.py
+++ b/gunicorn/pidfile.py
@@ -67,7 +67,7 @@ class Pidfile(object):
         try:
             with open(self.fname, "r") as f:
                 try:
-                    wpid = int(f.read() or 0)
+                    wpid = int(f.read())
                 except ValueError:
                     return
 


benoitc:
+1

berkerpeksag:
LGTM. Just curious: How did you noticed this? Did you see a traceback or just read the code?

brijeshb42:
I was trying to run a Flask app using gunicorn but the pidfile that I was providing was already created but its had a single empty line in it. So it was not starting and giving `invalid literal for int() with base 10 ''`. Then through the stacktrace, I was able to go to that particular line. So I added an extra check.

emamirazavi:
Why don't i see this problem?!

brijeshb42:
I was using gunicorn:0.19.3 .
I guess this problem won't appear if the pidfile you are supplying does not exist or contains a valid integer or exists but is completely empty.
Mine had a single empty line.

benoitc:
hrm `f.read() or 0` should fix the empty case normally. The valid integer case is still valid. Which version of Python are you using?

brijeshb42:
I am using `sys.version_info(major=2, minor=7, micro=6, releaselevel='final', serial=0)`

gunicorn/pidfile.py
```diff
@@ -66,7 +66,10 @@ def validate(self):
             return
         try:
             with open(self.fname, "r") as f:
-                wpid = int(f.read() or 0)
+                try:
+                    wpid = int(f.read())
+                except ValueError:
+                    wpid = 0
```
tilgovi: Might be clearer to simply return rather than relying on the subsequent if statement.

benoitc: also the case where the pid is an empty file should be handled. I would do the following:
```
try:
    wpid = int(f.read() or 0)
except ValueError:
    raise HaltError("pidfile is invalid")
```

or any other error message. Thoughts?

tilgovi: If the zero case fails anyway (we return right after) and `int('')` is a `ValueError` just return in the `except` clause and remove the subsequent `if`.

benoitc: right... I forgot we were returning. so indeed let just return.

benoitc: @brijeshb42 can you do the change on the PR? Will merge it right after :)

brijeshb42:
@benoitc Updated the code.

tilgovi:
I think the `or 0` just has to be removed. The intention is that if `read` returns the empty string that `int` raises `ValueError` and that is how we can skip the subsequent `if`.

brijeshb42:
@tilgovi added your suggestion.

tilgovi:
Thanks!

PR #1068, opened 2015-07-07
Prefix config file with its type

### LINKED ISSUE #836: import error in 19.1.0
I've the following error:

    Error: relative imports require the 'package' argument

No such error on previous installed version (19.0.0).

slava-sh:
Fixes #836 as described in @benoitc's [comment](https://github.com/benoitc/gunicorn/issues/836#issuecomment-50957348).

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -16,13 +16,19 @@ Config File
 config
 ~~~~~~
 
-* ``-c FILE, --config FILE``
+* ``-c CONFIG, --config CONFIG``
 * ``None``
 
-The path to a Gunicorn config file, or python module.
+The Gunicorn config file.
+
+A string of the form ``PATH``, ``file:PATH``, or ``python:MODULE_NAME``.
 
 Only has an effect when specified on the command line or as part of an
 application specific configuration.
+
+.. versionchanged:: 19.4
+   Loading the config from a Python module requires the ``python:``
+   prefix.
 
 Server Socket
 -------------
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -465,14 +465,20 @@ class ConfigFile(Setting):
     name = "config"
     section = "Config File"
     cli = ["-c", "--config"]
-    meta = "FILE"
+    meta = "CONFIG"
     validator = validate_string
     default = None
     desc = """\
-        The path to a Gunicorn config file, or python module.
+        The Gunicorn config file.
+
+        A string of the form ``PATH``, ``file:PATH``, or ``python:MODULE_NAME``.
 
         Only has an effect when specified on the command line or as part of an
         application specific configuration.
+
+        .. versionchanged:: 19.4
+           Loading the config from a Python module requires the ``python:``
+           prefix.
         """
 
 class Bind(Setting):
--- a/docs/source/run.rst
+++ b/docs/source/run.rst
@@ -47,8 +47,8 @@ You can now run the app with the following command::
 Commonly Used Arguments
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-* ``-c CONFIG, --config=CONFIG`` - Specify the path to a config file or
-  Python module.
+* ``-c CONFIG, --config=CONFIG`` - Specify a config file in the form
+  ``$(PATH)``, ``file:$(PATH)``, or ``python:$(MODULE_NAME)``.
 * ``-b BIND, --bind=BIND`` - Specify a server socket to bind. Server sockets
   can be any of ``$(HOST)``, ``$(HOST):$(PORT)``, or ``unix:$(PATH)``.
   An IP is a valid ``$(HOST)``.
--- a/gunicorn/app/base.py
+++ b/gunicorn/app/base.py
@@ -108,10 +108,15 @@ class Application(BaseApplication):
         Exception or stop the process if the configuration file contains a syntax error.
         """
 
-        try:
-            cfg = self.get_config_from_module_name(module_name=location)
-        except ImportError:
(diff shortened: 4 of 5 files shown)


tests/test_003-config.py
```diff
@@ -186,8 +186,15 @@ def test_load_config():
...
+def test_load_config_explicit_file():
+    with AltArgs(["prog_name", "-c", "file:%s" % cfg_file()]):
+        app = NoConfigApp()
+    t.eq(app.cfg.bind, ["unix:/tmp/bar/baz"])
+    t.eq(app.cfg.workers, 3)
+    t.eq(app.cfg.proc_name, "fooey")
+
 def test_load_config_module():
-    with AltArgs(["prog_name", "-c", cfg_module()]):
+    with AltArgs(["prog_name", "-c", "python:%s" % cfg_module()]):
```
berkerpeksag: I think it would be nice to keep the test untouched to make sure that we don't break backwards compatibility. We can add a separate test for the `python:` prefix.

gunicorn/config.py
```diff
@@ -465,11 +465,13 @@ class ConfigFile(Setting):
...
     cli = ["-c", "--config"]
-    meta = "FILE"
+    meta = "CONFIG"
     validator = validate_string
     default = None
     desc = """\
-        The path to a Gunicorn config file, or python module.
+        The Gunicorn config file.
+
+        A string of the form 'PATH', 'file:PATH', or 'python:MODULE_NAME'.
```
berkerpeksag: Can you replace the single quotes with backticks(e.g. 'PATH' -> \`\`PATH\`\`)?

berkerpeksag: We can also add a "versionadded" directive (or we can add a note to docs/source/news.rst)

docs/source/settings.rst
```diff
@@ -310,7 +312,7 @@ chdir
 ~~~~~
 
 * ``--chdir``
-* ``/Users/benoitc/work/gunicorn/py27/gunicorn/docs``
+* ``/Users/slava/Desktop/gunicorn/docs``
```
berkerpeksag: This looks like an unrelated change(and caused by `make -C docs html`) to me. Can you revert it? :)

docs/source/settings.rst
```diff
@@ -413,7 +415,7 @@ temporary directory.
 secure_scheme_headers
 ~~~~~~~~~~~~~~~~~~~~~
 
-* ``{'X-FORWARDED-PROTOCOL': 'ssl', 'X-FORWARDED-PROTO': 'https', 'X-FORWARDED-SSL': 'on'}``
+* ``{'X-FORWARDED-PROTOCOL': 'ssl', 'X-FORWARDED-SSL': 'on', 'X-FORWARDED-PROTO': 'https'}``
```
berkerpeksag: Same as above.

berkerpeksag:
Thanks for the PR, @slava-sh :)

Three comments:

* Maybe we can also add `py:` as an alternative to `python:`.
* Can you please squash the commits into one when you're done?
* An additional "thank you" for adding tests :)

tests/test_003-config.py
slava-sh: From my understanding of your [comment](https://github.com/benoitc/gunicorn/issues/836#issuecomment-50957348), we do want to break backward compatibility here in order to separate the two cases. What do you think?

gunicorn/config.py
slava-sh: Okay. Note that other descriptions, such as [bind](https://github.com/benoitc/gunicorn/blob/72a8e2940d75b3ef7a92edcb3e4247f43426706f/gunicorn/config.py#L494) and [syslog_addr](https://github.com/benoitc/gunicorn/blob/72a8e2940d75b3ef7a92edcb3e4247f43426706f/gunicorn/config.py#L1157), use single quotes as well.

slava-sh:
I don't see a point in having two different prefixes for doing one thing. We can choose a prefix different from `python:` though.

tilgovi:
:+1:

No opinion on `python` vs `py`. Both are fine. We should pick one.

benoitc:
@berkerpeksag why `py` instead of `python`? Shorter?

slava-sh:
How about `module`?

tilgovi:
An alternative idea: use a colon to distinguish between paths and modules, with the colon separating the package from the relative path.

Colon is uncommon in filenames because it's not very portable, so it's likely unambiguous.

Pyramid does this and calls them asset specifications.

"foo/bar" is a relative path
"/path/to/foo/bar" is an absolute path
"foo:bar" is the "bar" path within the "foo" package.

benoitc:
@tilgovi interesting idea. IF the pattern is already used why not. I was just reusing the pattern existing in trac and other stuff like this.

berkerpeksag:
> @berkerpeksag why py instead of python? Shorter?

Yes.

I also like @tilgovi's suggestion since it's already being used in the wild.

tilgovi:
I don't have a preference. Just mentioning it. I haven't seen it anywhere other than Pyramid. If Trac and others have `file:` that's fine, too. Paste uses `config:` and `egg:`. Anything is acceptable.

tilgovi:
I think I like `file:` and `python:` (or `module:`) since there is no guessing about the meaning. It is very explicit.

benoitc:
i would go for python: the code already exists :)

tilgovi:
:+1:

tilgovi:
Needs a rebase, though.

slava-sh:
I think this can be merged now.

benoitc:
LGTM

berkerpeksag:
Thanks!

Issue #791, opened 2014-06-14
add regression tests for the core.


benoitc:
we need to add regression tests for the workers and arbiter. I usually test everything, but since the team is growing we need them. It will prevent a lot of small issues introduced by some patches.

berkerpeksag:
I think we can create a test framework to test workers, configs, cmdline etc. properly and then people can use that framework to increase the test coverage in PyCon/EuroPython/DjangoCon sprints.

I'll do some cleanup in the test suite.

berkerpeksag:
Related issues:

* https://github.com/benoitc/gunicorn/issues/950
* https://github.com/benoitc/gunicorn/issues/947
* https://github.com/benoitc/gunicorn/issues/1063

benoitc:
Totally. I will to help. Things are slowly coming back under control :)

javabrett:
I was going to raise a new mailing-list topic "do we need a framework for system/acceptance testing", but found this issue open, so asking here instead, although this issue mentions regression-tests specifically.

Wondering if there's been any advancement in thoughts for some framework or approach for this.  Specifically I'm looking for the kind of coverage produced by an actual running _gunicorn_ server process, bound to a real socket, with a real client sending test-requests.

At the moment, if I for-example modify one of the workers and raise an exception on `handle`, this doesn't seem to be caught by an existing unit test, and if there's a sysint test I don't know how to run it.

I'm looking to help on issues like #1810, but don't know how to best add coverage for things like SSL which might be best covered by system tests.

Any thoughts appreciated.

PR #1071, opened 2015-07-08
Include request URL in error message.

### LINKED ISSUE #1061: Include request URL in error message
It would be really helpful if the logged error message were "Error handling request http://host/path/etc" instead of just "Error handling request".

wosc:
Fixes #1061.

I didn't write a test since there don't seem to be any similar tests yet,
and I'm not familiar enough with the test infrastructure.


berkerpeksag:
LGTM. I couldn't find an easy way to test this. See also #791.

PR #1072, opened 2015-07-08
Use the --install option instead of sudo.

berkerpeksag:


tilgovi:
How reliably is the `--user` install bin going to be in the user's path?

benoitc:
I have no ideas on what is doing this command, so I will trust you about it :)

berkerpeksag:
> How reliably is the --user install bin going to be in the user's path?

No idea, but it's better than sudo anyway :)

berkerpeksag:
Actually, we can just remove virtualenv lines.

```sh
$ pip install gunicorn
$ cat myapp.py
       data = b"Hello, World!\n"
       start_response("200 OK", [
       ])
       return iter([data])
$ gunicorn -w 4 myapp:app
```

looks simpler to me.

benoitc:
+1

tilgovi:
Sounds good to me. Virtualenv is the user's choice. We can add a note about virtualenv generally, pointing to a tutorial, if we want to.

benoitc:
@berkerpeksag so let's go with out sudo then :)

benoitc:
bump.

tilgovi:
:+1: to removing virtualenv from the instructions.

berkerpeksag:
I will update this today, thanks for the reviews!

PR #1074, opened 2015-07-09
Improve markup of settings.rst

slava-sh:


slava-sh:
I'm not sure if the ssl renaming doesn't break anything.

gunicorn/config.py
```diff
@@ -572,11 +570,11 @@ class WorkerClass(Setting):
         * ``gevent``   - Requires gevent >= 0.13
         * ``tornado``  - Requires tornado >= 0.2
 
-        Optionally, you can provide your own worker by giving gunicorn a
-        python path to a subclass of gunicorn.workers.base.Worker. This
-        alternative syntax will load the gevent class:
-        ``gunicorn.workers.ggevent.GeventWorker``. Alternatively the syntax
-        can also load the gevent class with ``egg:gunicorn#gevent``
+        Optionally, you can provide your own worker by giving Gunicorn a
+        Python path to a subclass of :class:`gunicorn.workers.base.Worker`.
```
berkerpeksag: There's no point to use the class role since we don't have document the *Worker classes.

gunicorn/config.py
```diff
@@ -644,7 +642,7 @@ class MaxRequestsJitter(Setting):
     type = int
     default = 0
     desc = """\
-        The maximum jitter to add to the max-requests setting.
+        The maximum jitter to add to the `max_requests` setting.
```
berkerpeksag: We are using \*foo\* instead of \`foo\` for this in the Python documentation.

gunicorn/config.py
```diff
@@ -1031,7 +1030,7 @@ class AccessLog(Setting):
     desc = """\
         The Access log file to write to.
 
-        "-" means log to stderr.
+        ``-`` means log to `stderr`.
```
berkerpeksag: \`stderr\` doesn't look very useful here. I'd say keep it as is.

gunicorn/config.py
```diff
@@ -1049,12 +1048,12 @@ class AccessLogFormat(Setting):
         Identifier  Description
         ==========  ===========
         h           remote address
-        l           '-'
-        u           currently '-', may be user name in future releases
+        l           ``-``
```
berkerpeksag: '-' means a string here, so \`\`'-'\`\` would be a better choice. See Python's codecs module documentation for a similar use case(e.g. \`\`'strict'\`\`).

gunicorn/config.py
```diff
@@ -1097,11 +1096,11 @@ class Loglevel(Setting):
 
         Valid level names are:
 
-        * debug
-        * info
-        * warning
-        * error
-        * critical
+        * ``debug``
```
berkerpeksag: I think the current version of this list looks clearer.

gunicorn/config.py
```diff
@@ -1323,9 +1322,9 @@ class Paste(Setting):
     validator = validate_string
     default = None
     desc = """\
-        Load a paste.deploy config file. The argument may contain a "#" symbol
-        followed by the name of an app section from the config file, e.g.
-        "production.ini#admin".
+        Load a ``paste.deploy`` config file. The argument may contain a ``#``
```
berkerpeksag: paste.deploy is actually a project, so the correct name should be PasteDeploy.

gunicorn/config.py
```diff
@@ -1546,13 +1545,13 @@ def nworkers_changed(server, new_value, old_value):
...
     desc = """\
-        Called just after num_workers has been changed.
+        Called just after `num_workers` has been changed.
 
         The callable needs to accept an instance variable of the Arbiter and
         two integers of number of workers after and before change.
 
-        If the number of workers is set for the first time, old_value would be
-        None.
+        If the number of workers is set for the first time, ``old_value`` would
```
berkerpeksag: \*old_value\*

gunicorn/config.py
slava-sh: It's OK. There won't a link if the class is not documented.

This is better semantically though.

berkerpeksag: Well, semantically you can also use :const:\`None\` instead of \`\`None\`\`, but there is no point to overuse Sphinx roles.

gunicorn/config.py
berkerpeksag: Another alternative: Activate the sphinx.ext.intersphinx extension and use the data role as :data:\`sys.stderr\`.

gunicorn/config.py
slava-sh: Okay, I'll revert this. It's [been used](https://github.com/benoitc/gunicorn/commit/33d35a1a665b4388a4d5cb77ea9903d0a2973d49#diff-9f531e15751f24da705e11f66baca7e3R10) by @matrixise though.

slava-sh:
Done.

slava-sh:
I think this is ready to be merged.

berkerpeksag:
Thanks, I will review and merge it today.

berkerpeksag:
Rebased and merged in https://github.com/benoitc/gunicorn/commit/0d86a9094d6b4d969c3075e39831076df6d2d0fc.

Thanks @slava-sh :)

PR #1075, opened 2015-07-10
New CLI argument --paste-server

dmdm:
Added app and server section "smeagol" to `examples/frameworks/pylonstest/development.ini`. Without really installing pylonstest and smeagolapp, we can determine by the log and error messages which server and app is loaded:
- app "main" is "pylonstest", "smeagol" is "smeagolapp"
- server "main" is on port 5000, "smeagol" on port 5111
  (Don't know how to write test cases for this config.)

Try combinations of:

```
--paste development.ini
--paste development.ini#main
--paste development.ini#main --paste-server smeagol
--paste development.ini#smeagol --paste-server main
--paste development.ini#smeagol
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ examples/frameworks/pylonstest/PasteScript*
 examples/frameworks/pylonstest/pylonstest.egg-info/
 examples/frameworks/django/testing/testdb.sql
 .tox
+.idea
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -419,7 +419,7 @@ temporary directory.
 secure_scheme_headers
 ~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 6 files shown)


benoitc:
what's the usage for it?

tilgovi:
Thank you for your contribution. I keep meaning to do something about this pull request, but I haven't been able to decide what to do. At this point, I'm going to close it. Between the gunicorn command line and the config file, gunicorn has enough ways to configure it. Using a server block in the paste file doesn't seem like a very useful addition to me, and is possibly confusing.

I've just made a PR to remove support for `egg:gunicorn#main` in the server block anyway. It is confusing to have a server block that may or may not mention gunicorn, but have gunicorn use settings from it.

Not all gunicorn features work as intended when invoked by the paste server. For instance, paste loads the app, which is sort of a force `--preload`. This can lead to confusion.

We might revisit this in the future, but we probably need to clarify the intent and see what can be done tom make the `egg:gunicorn#main` format work with paster well. If you are invoking Gunicorn rather than `pserve` or `paste` I think it makes sense to use Gunicorn's command line and config rather than the server block.

PR #1076, opened 2015-07-11
Django/Paster app cleanup

tilgovi:
Let's get this in now. :-D


tilgovi:
Oops, this isn't going to work because I took out the use: egg:gunicorn support.

benoitc:
you mean for paster? I would also trash it. It is also preloading apps by default. Which should be really discouraged.

tilgovi:
Hmmm. That means just skipping the whole server section and I think paste .ini users will still want to configure the server block. But maybe not. Maybe I can just remove it.

tilgovi:
I think I'm fine to enforce configuration of the server through CLI and gunicorn.conf.py

tilgovi:
It makes me feel bad that I just asked @dmdm to write #1075, though.

benoitc:
hrm... we can keep for now the `--paste` option. There maybe a way to not preload all the stuff there.

berkerpeksag:
Should this go in 19.4 or should we keep it for 20.0?

tilgovi:
Probably 20, yeah?

benoitc:
20 sounds ok. do we need a 19.4?

dmdm:
Please do not remove the whole paster code. The commands 'gunicorn-paster' etc you marked deprecated a while ago, so its ok for them, I think. But please keep '--paster' and its brother.
@benoitc said, "It is also preloading apps by default". Alas, I'm not gunicorn-learned enough to know what that means. Still, wouldn't it be possible to rewrite the paster initialisation code in such a way that it loads the apps at the right point?
Please

tilgovi:
`--paste` is not going anywhere. It's essential.

We might take out the support for `[server]` blocks entirely and `use: egg:gunicorn`.

tilgovi:
Instead, users will have to use the CLI and the gunicorn config file.

tilgovi:
Also, I'm sure I can be convinced to keep support for a lot of things if you keep sending pictures of cats that are that cute.

dmdm:
Afaik  `use = egg:gunicorn` is located in said `[server]` section, where else would you place it? Also, one cannot breed other settings from that egg, but has to use a different config file (.py or a startup script)... feels like peppering my settings all over the place ;) -- but other people might feel differently.

Sorry, no cats today. Had a long day trying to build a small websocket server with asyncio *sigh*

benoitc:
I propose the following:

1. keep the `--paste` setting for paster ecosystem. (I wish sometimes everyone would go to full wsgi but anyway)
2. keep the the `server`part for now, but deprecate it and add a notice about it making it clear that reload doesn't work quite well with it.

Thoughts?

benoitc:
@dmdm about preload: 
http://docs.gunicorn.org/en/19.3/settings.html#preload-app . 

The paster command afaik is not lazy and load a bunch of things before providing the WSGI application this is why it is recommended in production to not use it but instead use the `--paste` setting.

dmdm:
@benoitc Thanks for the pointer. Does that mean, if `max_requests` is > 0, and I change some code files, the workers will restart and eventually pick up the new version? (But still it would be non-deterministic how many requests each worker has left, leading some to serve new, and other to serve old code.)

About the "paster" stuff and "going full wsgi"... From the start some years ago I used Pyramid whose init returns a full wsgi application. So, maybe naively, I assumed the only thing `--paster` does is read the config from the ini file, and then handle it like any other wsgi app. Is there more pastery stuff going on?

tilgovi:
I think reload works fine with the server block as long as you use gunicorn
--paster and not pserve or others.

I'll check it out.

berkerpeksag:
> 20 sounds ok. do we need a 19.4?

I don't know. I asked because you were talking about releasing a bug fix release last week or so :)

> We might take out the support for [server] blocks entirely and use: egg:gunicorn.

It would be nice to deprecate them first. For example, GNU MediaGoblin uses them since Paster Server has a premature Python 3 support: http://git.savannah.gnu.org/cgit/mediagoblin.git/tree/mediagoblin/app.py#n370

benoitc:
what's the status of that PR ? @tilgovi do we still need it?

tilgovi:
I've lost a lot of the context here, but I think it's still the right thing for R20.

I think `--paste` should be the only supported way to load a paste .ini.

What about Django, though? Is it still right to remove all the code I removed in this PR?

berkerpeksag:
Yes, the Django part of the patch still needs to be deleted.

tilgovi:
Added to the 20.0 milestone. I'll bring this up to date.

berkerpeksag:
I'm planning to commit the Django part of this PR this week. Any objections? :)

benoitc:
go ahead

benoitc:
btw if you sprint at pycon let me know we should let other know about it on
twittrr and co :)

benoitc:
@tilgovi maybe we should just apply the patch now. Let me know.

benoitc:
let's remove compat this update doesn't have to wait for another major bump imo.

berkerpeksag:
Ok, will cherry-pick the Django part. I'm not familiar with the other half of the PR :)

berkerpeksag:
Applied Django part in https://github.com/benoitc/gunicorn/commit/3da2f88dfe7a79219138298d5a2c328014a9da64.

tilgovi:
Closing this. I'll open a new PR with just the remaining paster parts.

PR #1080, opened 2015-07-13
Clarify pythonpath setting description

malept:
Show that setting multiple paths requires using a comma-separated list.


benoitc:
thanks!

PR #1083, opened 2015-07-13
Use travis cache

tilgovi:


berkerpeksag:
LGTM, but caching .tox might cause funny failures.

We are using a different approach for non-wheel packages: https://github.com/python/pythondotorg/blob/master/.travis.yml#L9-13 (probably the recent versions of pip solved this problem. I have to check first)

tilgovi:
Tox is smart. It should notice when requirements change, I think.

PR #1085, opened 2015-07-14
Fix implicitly shared mutable instance properties.

rwe:
Several properties in `Arbiter` and other classes were actually mutable class properties masquerading as instance properties.

Note that because `Arbiter` is bound to a single PID, you would never expect to have multiple instances of it anyway. But that is not enforced so either way there is some weirdness here in that case, and it looked like the code was doing this unaware.

If it were desired (for example, to support unit testing),  there are a few other things preventing multiple Arbiter instances operating correctly in the same process:
1. The PID file.
2. The signal handlers.
3. The use of `sys.exit`.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -36,14 +36,6 @@ class Arbiter(object):
     # A flag indicating if an application failed to be loaded
     APP_LOAD_ERROR = 4
 
-    START_CTX = {}
-
-    LISTENERS = []
-    WORKERS = {}
-    PIPE = []
-
-    # I love dynamic languages
-    SIG_QUEUE = []
     SIGNALS = [getattr(signal, "SIG%s" % x)
                for x in "HUP QUIT INT TERM TTIN TTOU USR1 USR2 WINCH".split()]
     SIG_NAMES = dict(
@@ -62,13 +54,18 @@ class Arbiter(object):
         self.reexec_pid = 0
         self.master_name = "Master"
 
+        self.listeners = []
+        self.workers = {}
+        self.sig_queue = []
+        self.pipe = None
+
         cwd = util.getcwd()
 
         args = sys.argv[:]
         args.insert(0, sys.executable)
 
         # init start context
-        self.START_CTX = {
+        self.start_ctx = {
             "args": args,
             "cwd": cwd,
             0: sys.executable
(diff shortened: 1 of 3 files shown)


berkerpeksag:
LGTM, thanks. It also would be nice to make testing easier.

benoitc:
I will read the patch, but one thing is that Arbiter is not bound to a single pid when using USR2 signal. You will have 2 instances of it at that time when we fork it:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L366

Anyway will review the patch ASAP.

erydo:
@benoitc — Interesting, I didn't realize that: the question then is whether those two `Arbiter` instances residing in the same process are intended to share the same `WORKERS` and `SIG_QUEUE`.

Looking more carefully at how/where each of these were initialized, I see that:

_:white_check_mark: Original behavior preserved but clarified in this PR:_
1. `START_CTX` was always re-initialized as an instance member in `__init__`.
2. `PIPE` would first try to close the FDs in the empty class member list (a no-op), and then be re-initialized an instance member in `init_signals`.
3.  `LISTENERS` would always be re-initialized as an instance member in `start`.

_:warning: Potentially altered and unclear whether sharing was intended:_
1. `WORKERS` was never re-initialized as an instance member, and so was implicitly shared between instances.
2. `SIG_QUEUE` was never re-initialized as an instance member, and so was implicitly shared between instances.

tilgovi:
The `Arbiter` instances on `USR2` are not the same process. The arbiter forks.

tilgovi:
This looks really great. Thanks.

tilgovi:
I don't anticipate any problems with this. There have never been two arbiter instances in the same process, only an arbiter in each of the parent and child forked processes. There has never been any sharing. These can all safely be instance variables AFAICT.

benoitc:
@tilgovi but data are shared just like preload. At least this is how i
understabd execvpe

benoitc:
the sig queue is interresting to keep. for the rest i agree that we don't
need to share them.

tilgovi:
There is a `fork()` before `exec`. Data is not shared between forked processes. Data is _definitely_ not shared across `exec` because the whole point is to wipe everything and re-execute a new binary. The only thing shared, I think, is maybe some file descriptors and environment variables.

benoitc:
@tilgovi you're right!  I do think that technically workers, signals and pipe are class attributes more than instance class variables for me but well that's also true that we can only have one arbiter / process for now (there maybe a time it won't be but..)

So go ahead and sorry for the noise, I'm tired these days :)

benoitc:
side note, I would like to make a release before making that changes.

tilgovi:
Time to revisit this.

tilgovi:
I'd like to rebase this for master now as I think it's less confusing to have these be instance variables with lower case names than to appear to be class-level constants. Any objection?

erydo:
@tilgovi No objection from me — it's quite an old branch so hopefully that rebase goes smoothly. I've enabled edits from maintainers on this PR. Thanks!

PR #1092, opened 2015-08-03
Add SSL support to gaiohttp worker

tumb1er:
This pr just passes required ssl options from [Gunicorn config](http://gunicorn-docs.readthedocs.org/en/latest/settings.html#ssl) to `SSLContext` instance used in `create_server`
This is just minimal working example, but it is tested with self-signed certificate:
```sh
gunicorn -k gaiohttp --bind=127.0.0.1:443  --keyfile=ssl.key --certfile=ssl.crt app
```
Without patch gunicorn runs HTTP server over 443 port with ssl arguments present.

### DIFF
--- a/gunicorn/workers/_gaiohttp.py
+++ b/gunicorn/workers/_gaiohttp.py
@@ -23,7 +23,7 @@ class AiohttpWorker(base.Worker):
     def __init__(self, *args, **kw):  # pragma: no cover
         super().__init__(*args, **kw)
         cfg = self.cfg
-        if cfg.ssl_version:
+        if cfg.is_ssl:
             self.ssl_context = ssl.SSLContext(cfg.ssl_version)
             self.ssl_context.load_cert_chain(cfg.certfile, cfg.keyfile)
         else:
--- a/gunicorn/workers/_gaiohttp.py
+++ b/gunicorn/workers/_gaiohttp.py
@@ -24,11 +24,9 @@ class AiohttpWorker(base.Worker):
         super().__init__(*args, **kw)
         cfg = self.cfg
         if cfg.is_ssl:
-            self.ssl_context = ssl.SSLContext(cfg.ssl_version)
-            self.ssl_context.load_cert_chain(cfg.certfile, cfg.keyfile)
+            self.ssl_context = self._create_ssl_context(cfg)
         else:
             self.ssl_context = None
-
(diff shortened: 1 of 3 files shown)


gunicorn/workers/_gaiohttp.py
```diff
@@ -16,6 +22,12 @@ class AiohttpWorker(base.Worker):
 
     def __init__(self, *args, **kw):  # pragma: no cover
         super().__init__(*args, **kw)
+        cfg = self.cfg
+        if cfg.ssl_version:
```
benoitc: it should be `self.cfg.is_ssl` there, `cfg.ssl_version` is always set. (default is [ssl.PROTOCOL_TLSv1](https://github.com/benoitc/gunicorn/blob/master/gunicorn/config.py#L1642) .

Also why not wrapping the socket directly?

tumb1er: You are right about is_ssl. It was only an idea how to enable ssl for worker.
About wrapping directly: if you mean that `_create_server` method is redundant, consider a situation with subclassing `AiohttpWorker` and doing some magic with factory or socket before/after server creation. 
Without `_create_server` it's necessary to copy-paste and change the whole `_run` method (40 LOC) - not a good idea.

tilgovi:
LGTM. I know nothing about the gaoihttp server but this looks very straightforward.

tumb1er:
Not all ssl options are handled in this PR, but I don't know where to look for an example. Sync workers do this in `handle` method, not at server start.

benoitc:
@tumb1er i was thinking to [ssl.wrap_socket](https://docs.python.org/2/library/ssl.html#ssl.wrap_socket). Why using `ssl.SSLContex` here?

benoitc:
@tumb1er forget my question. I understood :)

benoitc:
@tumb1er can you rebase your commits in one? Will merge it right after.

tumb1er:
@benoitc what about other [ssl options](http://gunicorn-docs.readthedocs.org/en/latest/settings.html#ssl)? Should I add support for it?

benoitc:
@tumb1er oh right, yes it would be awesome :)

tumb1er:
This rebase magic is too complex for me :) I'll just create a new PR from other branch

PR #1094, opened 2015-08-06
Simplify sendfile logic

### LINKED ISSUE #1038: OSError [Error 29] llegal seek
Hi

   I have a python application on hand ,based on falcon(https://github.com/falconry/falcon),  and actually , I used the application as a proxy to access another one.

  in the application,  it just using requests.request to access another url and feed back current response
  when I run the application with 

gunicorn "abc.wsgi:make_api()" --bind="127.0.0.1:8000" --timeout=600 -w 1

  and run the 'curl -i http://127.0.0.1:8000/', it will failed with 

  2015-05-24 03:45:10 [129721] [ERROR] Error handling request
  File "/usr/local/lib/python2.7/dist-packages/gunicorn/workers/sync.py", line 134, in handle_request
    resp.write_file(respiter)
  File "/usr/local/lib/python2.7/dist-packages/gunicorn/http/wsgi.py", line 341, in write_file
    fd_offset = os.lseek(fileno, 0, os.SEEK_CUR)
OSError: [Errno 29] Illegal seek

  any if I write a verify simple app and run with gunicorn , I can't reproduce it 

  so what do you think can cause such error?

tilgovi:
A safe and reliable check for whether a file descriptor supports mmap
is to directly check if it is seekable. However, some seekable file
descriptors may also report a zero size when calling fstat. If there
is no content length specified for the response and it cannot be
determined from the file descriptor then it is not possible to know
what chunk size to send to the client. In this case, is it necessary
to fall back to unwinding the body by iteration.

The above conditions together reveal a straightforward and reliable
way to check for sendfile support. This patch modifies the Response
class to assert these conditions using a try/catch block as part of
a new, simplified sendfile method. This method returns False if it
is not possible to serve the response using sendfile. Otherwise, it
serves the response and returns True. By returning False when SSL is
in use, the code is made even simpler by removing the special support
for SSL, which is served well enough by the iteration protocol.

Fix #1038

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -12,7 +12,6 @@ import sys
 from gunicorn._compat import unquote_to_wsgi_str
 from gunicorn.six import string_types, binary_type, reraise
 from gunicorn import SERVER_SOFTWARE
-import gunicorn.six as six
 import gunicorn.util as util
 
 try:
@@ -23,6 +22,10 @@ except ImportError:
         from ._sendfile import sendfile
     except ImportError:
         sendfile = None
+
+# Send files in at most 1GB blocks as some operating systems can have problems
+# with sending files in blocks over 2GB.
+BLKSIZE = 0x3FFFFFFF
 
 NORMALIZE_SPACE = re.compile(r'(?:\r\n)?[ \t]+')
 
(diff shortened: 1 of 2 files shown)


gunicorn/http/wsgi.py
```diff
@@ -344,77 +347,44 @@ def write(self, arg):
...
-            while sent != nbytes:
-                sent += sendfile(sockno, fileno, offset + sent, nbytes - sent)
+        try:
+            fileno = respiter.filelike.fileno()
+            offset = os.lseek(fileno, 0, os.SEEK_CUR)
+            nbytes = max(os.fstat(fileno).st_size - offset, 0)
+            if self.response_length:
+                nbytes = self.response_length
+            assert(nbytes > 0)
+        except:
```
benoitc: shouldn't it catch a specific exception?

benoitc:
:+1: . I will make some test until monday but so far so good :)

gunicorn/http/wsgi.py
tilgovi: Not sure. I limited the number of lines here very carefully. If they fail, I'm not sure it matters why they fail, the result is the same: we shouldn't use sendfile.

If you want, we can specify. It would look like `(AssertionError, OSError, io.UnsupportedOperationError)` I think.

benoitc: @tilgovi forgot to answer. But yeah, a general catch should be enough. I'm +1 for it

tilgovi: You mean leave it as I have it or list the classes?

jstasiak: I'd at least catch ``Exception`` here; catching everything may result in catching and silencing an exception that shouldn't be caught and silenced (I don't know what context does this code execute in but for example ``KeyboardInterrupt`` and ``SystemExit`` would be something to watch out for).

tilgovi: Thanks. I added the specific exception classes.

gunicorn/http/wsgi.py
```diff
@@ -344,77 +347,44 @@ def write(self, arg):
...
-            sent += sendfile(sockno, fileno, offset + sent, nbytes - sent)
-            while sent != nbytes:
-                sent += sendfile(sockno, fileno, offset + sent, nbytes - sent)
+        try:
+            fileno = respiter.filelike.fileno()
+            offset = os.lseek(fileno, 0, os.SEEK_CUR)
+            nbytes = max(os.fstat(fileno).st_size - offset, 0)
+            if self.response_length:
+                nbytes = self.response_length
+            assert(nbytes > 0)
```
berkerpeksag: Perhaps we need to raise a ValueError here. If Python is invoked with -O or -OO, Python will remove all assert statements. So if nbytes is greater than 0 there will be no AssertionError.

gunicorn/http/wsgi.py
```diff
@@ -344,77 +347,44 @@ def write(self, arg):
...
-    def sendfile_all(self, fileno, sockno, offset, nbytes):
-        # Send file in at most 1GB blocks as some operating
-        # systems can have problems with sending files in blocks
-        # over 2GB.
+        return (self.cfg.sendfile and sendfile is not None)
 
-        BLKSIZE = 0x3FFFFFFF
+    def sendfile(self, respiter):
+        if self.cfg.is_ssl or not self.can_sendfile():
+            return False
```
berkerpeksag: Should we print an info message if sendfile is not available?

gunicorn/http/wsgi.py
```diff
@@ -344,77 +347,44 @@ def write(self, arg):
         util.write(self.sock, arg, self.chunked)
 
     def can_sendfile(self):
-        return (self.cfg.sendfile and (sendfile is not None))
-
-    def sendfile_all(self, fileno, sockno, offset, nbytes):
-        # Send file in at most 1GB blocks as some operating
-        # systems can have problems with sending files in blocks
-        # over 2GB.
+        return (self.cfg.sendfile and sendfile is not None)
```
berkerpeksag: Just a nitpick: () can be removed

gunicorn/http/wsgi.py
tilgovi: Maybe once, on startup, if we can reliably check whether it exists, to warn them that the option is ignored. Here is probably not the place. Maybe we can do that in a separate PR?

gunicorn/http/wsgi.py
tilgovi: Okay. I cleaned this up a bit. Thanks, @berkerpeksag!

gunicorn/http/wsgi.py
berkerpeksag: > Maybe we can do that in a separate PR?

+1

berkerpeksag:
LGTM

PR #1096, opened 2015-08-11
Forward SCRIPT_URL and REDIRECT_URL into wsgi env

ashald:
Main change: prevent prefixing these headers with 'HTTP_'.

These headers are very useful when you host your web-app under sub-path on the server and don't know 'SCRIPT_NAME' beforehand.
Example usage in Django core: https://github.com/django/django/blob/master/django/core/handlers/wsgi.py#L200

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -26,6 +26,8 @@ except ImportError:
 
 NORMALIZE_SPACE = re.compile(r'(?:\r\n)?[ \t]+')
 
+CGI_HEADERS = ("SCRIPT_URL", "REDIRECT_URL")
+
 log = logging.getLogger(__name__)
 
 
@@ -121,10 +123,12 @@ def create(req, sock, client, server, cfg):
 
     # set secure_headers
     secure_headers = cfg.secure_scheme_headers
+    cgi_headers = CGI_HEADERS
     if client and not isinstance(client, string_types):
         if ('*' not in cfg.forwarded_allow_ips
                 and client[0] not in cfg.forwarded_allow_ips):
             secure_headers = {}
+            cgi_headers = ()
 
     # add the headers to the environ
     for hdr_name, hdr_value in req.headers:
@@ -139,6 +143,9 @@ def create(req, sock, client, server, cfg):
             host = hdr_value
         elif hdr_name == "SCRIPT_NAME":
             script_name = hdr_value
+        elif cgi_headers and hdr_name in cgi_headers:
+            environ[hdr_name] = hdr_value
+            continue
         elif hdr_name == "CONTENT-TYPE":
             environ['CONTENT_TYPE'] = hdr_value
             continue


Ashald:
Any thoughts on this one?...

benoitc:
sounds good for me. @tilgovi ?

gunicorn/http/wsgi.py
```diff
@@ -139,6 +139,9 @@ def create(req, sock, client, server, cfg):
             host = hdr_value
         elif hdr_name == "SCRIPT_NAME":
             script_name = hdr_value
+        elif hdr_name in ("SCRIPT_URL", "REDIRECT_URL"):
```
benoitc: looking at the django path , maybe we should handle these values differently though. Thoughts?

benoitc:
commented the code. imo if we handle these settings then we should probably do like django.

Ashald:
@benoitc you mean deriving SCRIPT_NAME from them? Or preferring SCRIPT_URL and falling back to REDIRECT_URL if SCRIPT_URL is not defined?

benoitc:
@Ashald yes

Ashald:
@benoitc err, yes to which of statements? :) Deriving SCRIPT_NAME from them?

Ashald:
Oh, wait. I guess I got your point. You mean that we should do actually the same stuff as Django: process SCRIPT_URL, REDIRECT_URL and SCRIPT_NAME in the same way and set correct SCRIPT_NAME and PATH_INFO eventually? Well, that make's sense.

On the other hand, do we really need to do all that stuff here, in the web-server, if it's handled by frameworks?

benoitc:
@Ashald reimplementing https://github.com/django/django/blob/master/django/core/handlers/wsgi.py#L200-L226 . I wonder what @berkerpeksag and @tilgovi think about it?

Ashald:
@benoitc OK, let's wait for @tilgovi and @berkerpeksag replies.

On one hand, seeing how Django deals with it I'd like to keep it simple in Gunicorn and just pass these values into WSGI environ.

On the other hand, just took a quick look into Flask and didn't see anything regarding SCRIPT_URL there - it only handles SCRIPT_NAME. So it seems that if we want to make it uniform then we should really re-implement that logic here.

benoitc:
@Ashald exactly :) Maybe we should should support them for all. It's not fully WSGI compliant though. We also need to check which servers are using them, to make sure it's needed.

Ashald:
@benoitc so far I didn't see nor Apache nor nginx setting this values automatically but on the other hand it's very handy to be able to set this value manually in the web-server and thus make sure you WSGI app will derive correct SCRIPT_NAME and PATH_INFO from them.

Ashald:
Doing a bit more research on the topic (though I guess you guys already know this).

It seems that SCRIPT_URL is part of of CGI/1.1 spec [not the best source but still] (http://www.aprelium.com/data/doc/2/abyssws-win-doc-html/cgivars.html) - "The original request URI sent by the client".

According to this [discussion] (http://www.perlmonks.org/bare/?node_id=70301) it's automatically set by mod_rewrite.

According to [PEP0333] (https://www.python.org/dev/peps/pep-0333/#environ-variables):
> A server or gateway should attempt to provide as many other CGI variables as are applicable.

From this point of view, I guess, Gunicorn shouldn't process SCRIPT_URL value as Django does but rather pass it into WSGI environment. Though if we will be following this approach, then it may make sense to pass not only SCRIPT_URL value into WSGI environment but rather all other CGI vars too.

Ashald:
@tilgovi and @berkerpeksag, may I ask you about your thoughts on this?

tilgovi:
I'm surprised to find that we accept any of these as headers. Should that be limited the same way X-Forwarded-For trust is controlled by config?

Ashald:
@tilgovi so you don't mind forwarding them as long as they come from trusted proxy?
What do you think about how should we process them? Just pass them into environment? Or derive SCRIPT_NAME and PATH_INFO from them?

tilgovi:
We should do as little as possible, I think, and just pass them into the environment.

berkerpeksag:
+1 to what @tilgovi said.

Ashald:
@benoitc, @tilgovi, @berkerpeksag I updated the PR to pass SCRUPT_URL and REDIRECT_URL into WSGI env only if they come from trusted proxy (the same check as for secure headers).

Are you happy with this version? :)

tilgovi:
This looks great to me.

Ashald:
@tilgovi I'm glad you liked it. :)

berkerpeksag:
Can you squash the commits into one please?

Ashald:
@berkerpeksag done.

Off-topic: do I need to mention people so they will get notified if they're already participants? Sorry/thanks.

tilgovi:
@Ashald a comment is usually enough, without the mention. However, a commit/push is not enough to trigger a notification, I think.

gunicorn/http/wsgi.py
```diff
@@ -121,10 +121,12 @@ def create(req, sock, client, server, cfg):
 
     # set secure_headers
     secure_headers = cfg.secure_scheme_headers
+    cgi_headers = ("SCRIPT_URL", "REDIRECT_URL")
```
benoitc: Maybe we can set them as a global variable?

tilgovi: Like just put the default tuple into a global `CONSTANT`?

benoitc: indeed :)

Ashald: Does the constant at the top of the same module good enough for this?

tilgovi: I would think so.

benoitc: totally :)

Ashald:
Put headers into module-level constant. Waiting for further comments. :)

@tilgovi thanks for the tip (sorry for one more explicit mention :) )

Ashald:
Can we merge this one? :)

benoitc:
@Ashald patch is OK for me. I just asked to the django team the reason of these variables, since it's seems unclear for me why HTTP headers are not used until the spec is updated.

benoitc:
Getting further on that topic, it seems that SCRIPT_NAME is only added when using CGI in apache. RewriteCond has the following variables:

    SCRIPT_FILENAME
    PATH_INFO

http://httpd.apache.org/docs/2.2/fr/mod/mod_rewrite.html#rewritecond

It seems in that case we could reconstruct the script_name using the PATH_INFO variable. Like mod_wsgi does:
https://github.com/GrahamDumpleton/mod_wsgi/blob/develop/src/server/mod_wsgi.c#L466-L490

It looks better for me than adding new variables to the WSGI environnement. Thoughts?

tilgovi:
Will that solve the issue?

benoitc:
Reading the apache doc yes since we will be able to create a script_name env from the PATH_INFO. I don't think we need to pass other headers after that. Isn't it?

tilgovi:
@Ashald do you actually need these variables honored and passed or do you just need SCRIPT_NAME to be properly calculated for you?

Ashald:
Actually I don't care about that variables if things will work for me without them. :)

Let me give an example of what I'm dealing with. Please note that although it's an example derived from real-world infrastructure it's an abstract one since I cannot disclose real infrastructure.

I have an app (let's say its name is `foo`) that provides some urls like `/index` and `/info` - i.e. it expects to be 'mounted at the root of url path'.

In my environment (which I have no control over), let's say we have following hosts:
- `example.com`
- `foo.example.com`

All of them serve as a [Nginx-based] reverse proxies that's so:
`example.com/foo/index -> foo.example.com/index -> gunicorn/index`

As you can see in example above, 1st reverse proxy strips `/foo` prefix off while passes the request to `foo.example.com`.

What I need to achieve is to be able to access the app as:
- `example.com/foo/`[index, info]
- `foo.example.com/`[index, info]

and have this app to generate all urls (which are derived from SCRIPT_NAME and PATH_INFO in Django and other frameworks) properly.

With this environment the only thing I can do is to set up or pass arbitrary headers in reverse proxies and change `gunicorn` configuration.

Obviously there are no problems with making `foo.example.com/`[index, info] work - the app is mounted to the root there.

Though it's a bit tricky when we need to make it work through 2 proxies.
As I see, there are 2 ways how to deal with this:

1. In 1st proxy (`example.com`) set `SCRIPT_NAME=/foo` and make sure that it will be passed by 2nd proxy (`foo.example.com`) into the `gunicorn`. With this we need ability to say `gunicorn` to trust `SCRIPT_NAME` HTTP header that came from 2nd proxy. Also in this case `Gunicorn` should just use provided `SCRIPT_NAME` and not try to remove it from `PATH`/`PATH_INFO`. This is addressed in PR #1097.
2. Provided that Django (right now I'm experimenting with Django app) can derive `SCRIPT_NAME` from `SCRIPT_URL` WSGI env by removing `PATH_INFO` from the end of it, we can set up 1st proxy (`example.com`) to store original request path into `SCRIPT_URL` header.

Generally speaking, both of these solutions are achievable with `uwsgi` using `uwsgi_param` in `nginx` configuration (so you can take value from header and pass it directly into WSGI env). Though I'm more used to `gunicorn` and have a lot of apps based on it so I'd rather prefer to stick to further and that's why I'm trying to contribute some enhancements to deal with my problem. Although I don't think that my case is very common I'm sure that I'm not the only one who need to deal with apps behind several proxies so I think it will be useful to other people as well.

OTOH it might be that there is some other solution to my problem that I didn't think of...

benoitc:
Do we still need to something there?

tilgovi:
If mod_rewrite uses REWRITE_URL and SCRIPT_URL and it's helpful for users to have support in Gunicorn I don't see any problem with this patch.

tilgovi:
Maybe SCRIPT_NAME should be in cgi_headers as well.

benoitc:
@tilgovi well at least `SCRIPT_NAME` is part of the pep333 (for what it worth). 

I'm fine with the script, but we should keep a list somewhere of the additions to the WSGI spec. Thoughts?

tilgovi:
:+1: We need to document this if we're going to merge it.

Ashald:
As I mentioned above, according to [PEP0333] (https://www.python.org/dev/peps/pep-0333/#environ-variables):
> A server or gateway should attempt to provide as many other CGI variables as are applicable.

From this point of view we can go as far as forwarding not only there 2 variable but the whole bunch of other CFI variables. What do you think about it?

benoitc:
@Ashald but neither are part of the [RFC 3875](http://tools.ietf.org/html/rfc3875#section-4.1) :) In that case like in WSGI the spec [recommend](http://tools.ietf.org/html/rfc3875#section-4.1.18) to prefix them by `HTTP_` .

Anyway I don't see any reason to not accept the patch but we should really document it :)

Ashald:
Do you have in mind where it would be appropriate to put the docs for this (I mean what section would you prefer)?

tilgovi:
Maybe just in the deployment docs. These already discuss nginx in detail.
Maybe it can be in the example config there and mentioned in a little
subsection about cgi variables below?

What do you think?
http://docs.gunicorn.org/en/latest/deploy.html#nginx-configuration

benoitc:
bump

tilgovi:
I think we should close this as it's been inactive and we never seem to have reached a consensus on exactly what CGI variables to support. At this point, I would prefer we don't try to support that.

tilgovi:
Closing due to inactivity.

PR #1097, opened 2015-08-11
Don't fail if SCRIPT_NAME isn't part of PATH

Ashald:
If SCRIPT_NAME isn't part of PATH assume that it already was removed from path PATH and use PATH value as PATH_INFO.

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -193,7 +193,7 @@ def create(req, sock, client, server, cfg):
     # set the path and script name
     path_info = req.path
     if script_name:
-        path_info = path_info.split(script_name, 1)[1]
+        path_info = path_info.split(script_name, 1)[-1]
     environ['PATH_INFO'] = unquote_to_wsgi_str(path_info)
     environ['SCRIPT_NAME'] = script_name
 


benoitc:
hrm why it wouln't be part of the path if it has been set?

Ashald:
@benoitc it could be already removed from the PATH by front-end web server. In fact, that's the situation I'm dealing with.

Also, not failing is better that failing, IMHO. :)

Ashald:
@benoitc do you mind merging this one? :)

Ashald:
@tilgovi, @berkerpeksag can I ask your opinion on this one too? :)

benoitc:
Sorry for teh delay to answer. I am really busy these days... IMO if SCRIPT_NAME is hanfled before it shouldn't be in the env. But anyway, I would rather handle such case more explicitly in the code ie. using try/except or catching this error later and log it eventually ignoring it. Thoughts?

Ashald:
@benoitc no worries, I just though that hearing collaborators' opinions would be useful.

As for the SCRIPT_NAME, let me explain the situation I'm facing - maybe it will change your mind. ;)

Let's say I have an app (actually, a Django app) that has all URLs starting from `/my-app`. And now I need to host it under the `/foo/my-app path`. For whatever reason there is a front-end web-serevr that strips `/foo` from the PATH and passes only the `/my-app` part of the PATH into the app. In this case in order to generate proper URLs and redirects the app should know that its SCRIPT_NAME actually is `/foo. All this is valid as long as the app knows that in order to get the full path it should use SCRIPT_NAME + PATH_INFO.

As for how I ended up in such situation, we have a number of apps and each of them also has a part of web-server configuration that upon deployment is just being picked up by the front-edn web-server. Our goal is to make apps 'universal' and 'movable' that's so you can deploy them under any path (either root `/` or some sub-path like `/foo/`) without changing app configuration or web-server configuration includes.
If we will be able to set SCRIPT_NAME in the main front-end webs-erver configuration it will mean that we can include apps' web-server configs without any changes.

All that said, PR #1096 addresses the same issue even in more flexible way. So if that one will be merged I care less about this PR. Though I still think that we shouldn't require SCRIPT_NAME to be part of the PATH or at least provide an option in order to allow such behavior.  Basically speaking I would say that allowing SCRIPT_NAME to be missing from PATH is like assuming that 'there might be someone in front of us who already dealt with SCRIPT_NAME and PATH_INFO and we should just take them as is'.

tilgovi:
Wait, why do we try to split the SCRIPT_NAME off anyway? Why would PATH_INFO and SCRIPT_NAME be set with PATH_INFO having SCRIPT_NAME at the start? Shouldn't they be disjoint?

Ashald:
@tilgovi I guess that's because (some? or all?) Python-based web-frameworks expect that PATH_INFO in WSGI environ wouldn't contain SCRIPT_NAME and Gunicorn does the 'sanity check' ensuring that there won't be SCRIPT_NAME prefix within PATH_INFO. OTOH currently Gunicorn doesn't trust anybody to do that for him: if SCRIPT_NAME isn't a part of PATH_INFO (i.e. someone already removed it), Gunicorn will fail. That's what this PR is intended to fix.

tilgovi:
Thanks. I think we should consider removing that stripping behavior altogether. It would actually do the wrong thing if you legitimately had an app located at /foo/foo/*

benoitc:
@tilgovi this a direct implementation of the PEP. Traveling for now but I will find the link to it.

tilgovi:
Okay, please do. I can't find any text that says that. In PEP-0333.

benoitc:
well here the description is self explanatory:

    SCRIPT_NAME
        The initial portion of the request URL's "path" that corresponds to the application object, so that the application knows its virtual "location". This may be an empty string, if the application corresponds to the "root" of the server. 
    PATH_INFO
        The remainder of the request URL's "path", designating the virtual "location" of the request's target within the application. This may be an empty string, if the request URL targets the application root and does not have a trailing slash. 


and the example there:
https://www.python.org/dev/peps/pep-0333/#url-reconstruction

In my understanding, PATH_INFO is the part without the SCRIPT_NAME since it's the remainder. Am I wrong? It has not been an issue since many years anyway.

tilgovi:
It's not likely to be an issue. It would only be an issue if you were hosting an application at "/wiki/wiki" and had a reverse proxy in front of gunicorn send "/wiki" as the script name and path info. Gunicorn would drop one of the parts.

No?

Ashald:
How about the following approach: 
1) is there is PATH_INFO header set by trusted proxy then we just providing it into the WSGI env as is
1.1) If SCRIPT_NAME header is set in env or in header then providing it as is
1.) if there is no SCRIPT_INFO - using '/'
2) if there is not PATH_INFO in headers or proxy is not trusted we can try to derive PATH_INFO from PATH (and if SCRIPT_NAME is present then try to remove it from the PATH)

This way the implementation will be backward compatible and flexible enough for different deployment scenarios.

tilgovi:
I think that sounds right!

benoitc:
@tilgovi hrm I am not sure I understand the example you provided. If the path is `/wiki/wiki`and `SCRIPT_NAME` is `/wiki`, shouldn't the `PATH_INFO` be `/wiki`then ? This how I understand the reconstruction url example. 

@tilgovi @Ashald is it legal to send a `PATH_INFO` though? In that case I agree we should keep that one. I will look at it. (I am in vacation this month, so there will be some latency).

tilgovi:
Since `path_info` here is `req.path` and never a `PATH_INFO` header it is definitely correct to strip the `SCRIPT_NAME`.... unless the proxy already has :-/.

benoitc:
closing issue. nothing to solve there, the current path_info resolution works as described in the spec.If someone want to go further in the discussion please open a new ticker.

PR #1102, opened 2015-08-22
ISO-8859-1 encoded http headers

ephes:
Hi,

gunicorn uses utf8 encoding for http response headers. I don't know
much about http standards, but this is probably not correct:

http://stackoverflow.com/questions/4400678/http-header-should-use-what-character-encoding

best regards,
Jochen

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -508,6 +508,15 @@ def to_bytestring(value):
     return value.encode("utf-8")
 
 
+def to_latin1(value):
+    """Converts a string argument to a byte string"""
+    if isinstance(value, bytes):
+        return value
+    if not isinstance(value, text_type):
+        raise TypeError('%r is not a string' % value)
+    return value.encode("latin-1")
+
+
 def is_fileobject(obj):
     if not hasattr(obj, "tell") or not hasattr(obj, "fileno"):
         return False
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -317,7 +317,7 @@ class Response(object):
         tosend.extend(["%s: %s\r\n" % (k, v) for k, v in self.headers])
 
         header_str = "%s\r\n" % "".join(tosend)
-        util.write(self.sock, util.to_bytestring(header_str))
+        util.write(self.sock, util.to_latin1(header_str))
         self.headers_sent = True
 
     def write(self, arg):
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -1,6 +1,14 @@
+# -*- encoding: utf-8 -*-
+
 import t
+from gunicorn import util
 from gunicorn.http.body import Body
+from gunicorn.http.wsgi import Response
 from gunicorn.six import BytesIO
+try:
+    import unittest.mock as mock
+except ImportError:
+    import mock
 
 
 def assert_readline(payload, size, expected):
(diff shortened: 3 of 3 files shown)


benoitc:
The tests don't pass. `unitests.mock` don't pass. 

Also reading the new [RFC 7230](https://tools.ietf.org/html/rfc7230#section-3.2.4) :+1: 

>    Historically, HTTP has allowed field content with text in the
   ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
   through use of [RFC2047] encoding.  In practice, most HTTP header
   field values use only a subset of the US-ASCII charset [USASCII].
   Newly defined header fields SHOULD limit their field values to
   US-ASCII octets.  A recipient SHOULD treat other octets in field
   content (obs-text) as opaque data.

Not sure how to handle opaque data there. Also including such change we should also test if it works with some value given as unicode (which happen sometime in some countries...). Maybe we should have more tests there.

jamadden:
>  Also including such change we should also test if it works with some value given as unicode (which happen sometime in some countries...). 

Of course, according to the [WSGI spec](https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types), that's not supposed to happen in Python 2. Headers are specified to be given as the "native string type", so they should already be bytes and applications that send `unicode` values are in non-compliance with the spec (I've seen middleware break due to a buggy application that had a `unicode` header value). Likewise under Python 3 (where the native string type is unicode) including non-latin-1-encodable data is *also* out of compliance with the spec (the HTTP spec this time). So either case will enter implementation-defined behaviour and not be interoperable.

tilgovi:
:+1: to this change

gunicorn/util.py
```diff
@@ -508,6 +508,15 @@ def to_bytestring(value):
     return value.encode("utf-8")
 
 
+def to_latin1(value):
+    """Converts a string argument to a byte string"""
+    if isinstance(value, bytes):
+        return value
+    if not isinstance(value, text_type):
+        raise TypeError('%r is not a string' % value)
+    return value.encode("latin1")
```
berkerpeksag: latin1 -> latin-1. latin1 is an alias of latin-1.

tests/test_http_header.py
```diff
@@ -0,0 +1,28 @@
...
+import gunicorn.util as util
+
+from gunicorn.http.wsgi import Response
+try:
+    import unittest.mock as mock
+except ImportError:
+    import mock
+
+def test_http_header_encoding():
+    """ tests wether http response headers are ISO-8859-1 encoded """
```
berkerpeksag: typo: wether

berkerpeksag: We can move this to tests/test_http_body.py and rename it to test_http.py.

berkerpeksag:
Good catch, thanks! Could you please squash the commits?

tests/test_http_header.py
benoitc: +1

ephes:
Ok, squashed the commits :).

docs/source/run.rst
```diff
@@ -57,7 +57,7 @@ Commonly Used Arguments
   Check the :ref:`faq` for ideas on tuning this parameter.
 * ``-k WORKERCLASS, --worker-class=WORKERCLASS`` - The type of worker process
   to run. You'll definitely want to read the production page for the
-  implications of this parameter. You can set this to ``$(NAME)``
+  implications of this parameter. You can set this to ``egg:gunicorn#$(NAME)``
```
berkerpeksag: This change shouldn't be here :) See the original commit: https://github.com/benoitc/gunicorn/commit/8de5eb9b15bc429d6acece2935c60ec9f323d4f9

In general, the patch LGTM except this, but I can take care of it if you don't have time.

Thanks!

ephes:
Yup, this line was a leftover from an unintentional merge :/. Thanks for pointing it out - it's now removed.

berkerpeksag:
Thanks!

PR #1105, opened 2015-08-25
SSL support for gaiohttp worker

tumb1er:
Initialize and pass `SSLContext` to `asyncio.loop.create_server` to enable ssl support.

Once again - see #1092


benoitc:
:+1:

benoitc:
thanks!

berkerpeksag:
If we have a _create_socket method to make subclassing and customization easier, shouldn't we make it a public method(e.g. without a _ prefix and document its intent in the docstring)?

PR #1106, opened 2015-08-26
remove checking timeouts while graceful stopping tornado worker

keakon:
Tornado add many timeouts internally which can be ignored while graceful stopping.

### DIFF
--- a/gunicorn/workers/gtornado.py
+++ b/gunicorn/workers/gtornado.py
@@ -58,7 +58,7 @@ class TornadoWorker(Worker):
                         pass
                 self.server_alive = False
             else:
-                if not self.ioloop._callbacks and len(self.ioloop._timeouts) <= 1:
+                if not self.ioloop._callbacks:
                     self.ioloop.stop()
 
     def run(self):


tilgovi:
I think this is probably fine. @benoitc?

benoitc:
:+1:  @keakon can you eventually squash the commits? Sorry for the delay to review it, I'm incredibly busy these days...

keakon:
Hello, @benoitc. The first 2 commits were about 2 years ago, and you should have merged it.

Actually, there would be anther issue when I using Supervisor to manage Gunicorn process.

The default graceful_timeout of Gunicorn is 30 seconds. So when I restarting Gunicorn master process, it will wait its workers to exit up to 30 seconds.  
But the default stopwaitsecs of Supervisor is 10 seconds. So once the Gunicorn workers didn't exit within 10 seconds, the master would be killed by Supervisor, then the workers would become zombie processes.

I have changed the graceful_timeout to 8 seconds in my project, but I think there could be a mechanism to kill the workers once the master got killed.

benoitc:
I should probably have but I am really busy these days launching a new business. 

About the patches can you split the prs? the possible infinite loop should be done in another PR. 

About the timeout can't you tell to supervisor to send a SIGTERM instead of a SIGQUIT? There is indeed already a mechanism to exit immediately Gunicorn. If it doesn't work with the tornado worker, let me know.

keakon:
OK, I have updated the pull request.

Changing the signal is fine, but not graceful. Maybe adding a note in the document would be helpful.

tilgovi:
It seems like this is ready to go. I'm going to merge it.

benoitc:
@tilgovi thanks!

@keakon it's document there: 
http://docs.gunicorn.org/en/19.3/signals.html

If you want something more explicit let me know :)

Issue #1087, opened 2015-07-20
dictConfig in logconfig


roboslone:
From logging docs:

> Note: The fileConfig() API is older than the dictConfig() API and does not provide functionality to cover certain aspects of logging. For example, you cannot configure Filter objects, which provide for filtering of messages beyond simple integer levels, using fileConfig(). If you need to have instances of Filter in your logging configuration, you will need to use dictConfig(). Note that future enhancements to configuration functionality will be added to dictConfig(), so it’s worth considering transitioning to this newer API when it’s convenient to do so.

Could you suppport [`dictConfig`](https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig) in `logconfig` option?

tilgovi:
Can you confirm what you mean. Having a dictionary somehow in the command line with `--log-config`, is not what you want, correct? But a dictionary when using the `logconfig` setting in your gunicorn conf python file you want to set a dictionary?

RoboSloNE:
My application reads config from file and part of this config is for logging. I want to pass that part to dictConfig. I'm not using command line arguments for logging config at all.

berkerpeksag:
Looks like a reasonable request to me. I think we can also deprecate `logconfig` (or encourage to use the new setting) and add a new setting for this.

benoitc:
+1 . Any PR is welcome :)

MatMoore:
Hello - I'm also interested in this feature. Is the PR above similar to what you had in mind?

benoitc:
I fail to see how the command line can be used. Can someone provide a proper example? And test?

MatMoore:
@benoitc is a CLI option needed for every setting? It's not something I've addressed in the pull request above - that just works with the python configuration file.

If it was a CLI option, how would you use it? The only way I could see it working is if the value is dumped in a json or yaml file and the filename is used as the setting - but that just makes things more complicated if you are already using the python config.

akaihola:
Maybe the way Circus handles logging configuration could be used as a model? See [circus/util.py:configure_logger()](/dotmobo/circus/blob/967ec0296c101ec6deb4e9275bfb95f603d722d0/circus/util.py#L662).

benoitc:
@akaihola how is it different from the way we are doing it now? Also we should not forget that the logging backend is pluggable in Gunicorn.

ssalonen:
I found a workaround using `--config` parameter: one can use `dictConfig` there to setup the logging.

[...]

PR #1110, opened 2015-09-03
Issue 1087: support dictConfig logging configuration

MatMoore:
Hi,

This PR is a proof of concept for issue #1087.

So far I have only done a brief test, and I am not too familiar with the existing config implementation, but I will happily look into this in more detail and clean it up if I can get some feedback on the approach taken.

As pointed out in the linked issue, dictConfig enables certain functionality that is not available through the older fileConfig interface.

I've kept the new option separate from --config for backwards compatibility, but dictConfig takes precedence. I don't know whether --config should be entirely deprecated though.

@RoboSloNE @berkerpeksag what are your thoughts?

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -7,7 +7,7 @@ import base64
 import time
 import logging
 logging.Logger.manager.emittedNoHandlerWarning = 1
-from logging.config import fileConfig
+from logging.config import fileConfig, dictConfig
 import os
 import socket
 import sys
@@ -199,7 +199,11 @@ class Logger(object):
                 self.access_log, cfg, self.syslog_fmt, "access"
             )
 
-        if cfg.logconfig:
+        if cfg.logconfig_dict:
+            config = CONFIG_DEFAULTS.copy()
+            config.update(cfg.logconfig_dict)
+            dictConfig(config)
+        elif cfg.logconfig:
             if os.path.exists(cfg.logconfig):
                 defaults = CONFIG_DEFAULTS.copy()
                 defaults['__file__'] = cfg.logconfig
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -7,7 +7,12 @@ import base64
 import time
 import logging
 logging.Logger.manager.emittedNoHandlerWarning = 1
-from logging.config import fileConfig, dictConfig
+from logging.config import fileConfig
+try:
+    from logging.config import dictConfig
+except ImportError:
+    # python 2.6
+    dictConfig = None
 import os
 import socket
 import sys
@@ -199,7 +204,7 @@ class Logger(object):
                 self.access_log, cfg, self.syslog_fmt, "access"
             )
 
-        if cfg.logconfig_dict:
+        if dictConfig and cfg.logconfig_dict:
             config = CONFIG_DEFAULTS.copy()
             config.update(cfg.logconfig_dict)
             dictConfig(config)
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1138,6 +1138,19 @@ class LogConfig(Setting):
     """
 
 
+class LogConfigDict(Setting):
+    name = "logconfig_dict"
+    section = "Logging"
+    validator = validate_dict
+    default = {}
+    desc = """\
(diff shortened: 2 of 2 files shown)


berkerpeksag:
Thanks for the PR, @MatMoore :) This approach looks good to me. I'll review it more closely soon.

I don't think we should touch --config, but I'm in favor of deprecating --logconfig. Having a separate and outdated setting(also a separate file) for logging configuration is not a good a thing for users :)

I also would like to disable passing --dictconfig since it makes no sense and it confuses users. Is there a simple way to do this, @benoitc?

gunicorn/glogging.py
```diff
@@ -199,7 +199,9 @@ def setup(self, cfg):
                 self.access_log, cfg, self.syslog_fmt, "access"
             )
 
-        if cfg.logconfig:
+        if cfg.logconfigdict:
+            dictConfig(cfg.logconfigdict)
```
berkerpeksag: Haven't played with the patch yet, but perhaps we should update CONFIG_DEFAULTS first and then pass it to dictConfig?

benoitc:
Thanks for the patch! 

Do we really need another config setting? It seems that you can only pass the `dictConfig` via the config file. So we can probably only check it instead and accept it. 

Also in case you didn't notice it the CI is broken with the current patch:

        def validate_dict(val):
            if not isinstance(val, dict):
    >           raise TypeError("Value is not a dictionary: %s " % val)
    E           TypeError: Value is not a dictionary: None

MatMoore:
Doh, sorry about that - this was very rushed, but I'll try to spend some more time on it this weekend (and I'll make the tests pass this time!)

@benoitc Could you clarify what you mean by only check it and accept it? With the new settings object I was only planning on using it for basic validation and documentation, not cli options, so I can do it differently if there's a simpler way?

I suppose we could also reuse logconfig for the dictConfig version, while still supporting the deprecated file values, but it doesn't seem like the setting metaclass was designed for this usage.

gunicorn/glogging.py
MatMoore: Yep, using dictConfig for the default case seems sensible, at least if fileConfig will be eventually removed. I'll take a closer look at this.

MatMoore: I misunderstood your comment earlier, but I've implemented this now (i.e. the defaults will be applied regardless of whether dictConfig or fileConfig is used). As far as I can tell the format is the same for both.

MatMoore:
Something I didn't realise earlier is that dictConfig wasn't added until 2.7, so I've modified my changes to ignore the new setting on older versions. Do you want me to still put a deprecation notice for --logconfig?

gunicorn/glogging.py
```diff
@@ -199,7 +204,11 @@ def setup(self, cfg):
                 self.access_log, cfg, self.syslog_fmt, "access"
             )
 
-        if cfg.logconfig:
+        if dictConfig and cfg.logconfig_dict:
```
berkerpeksag: I think we can print a user friendly message if `dictConfig is None` instead of silently ignore the problem. I will review this more carefully soon. Sorry for my late response!

MatMoore: No problem :) I think making dictConfig an error on 2.6 and below would probably be best, but I'll wait until you're done with the review before changing anything.

benoitc:
bump ? What is needed to close that PR?

berkerpeksag:
I need some time to review this. I can't make any promise for 19.7, but I'll review it eventually :)

benoitc:
@berkerpeksag thanks :)

MatMoore:
Hi, I'm going to close the PR as I don't personally need the feature any more and I'm not particularly engaged with making sure it gets into gunicorn. In the meantime it just clutters my open PRs list.

@benoitc if the change is still useful feel free to raise it again as a new PR. Thanks for following up on this.

benoitc:
@berkerpeksag reopened after #1572 discussion.  What do you think we should do there to apply it?

aconrad:
Would anyone mind to update this PR with the latest master to fix the coverage report?

tilgovi:
Rebased. See #1602.

PR #1112, opened 2015-09-08
Update Debian/Ubuntu installation instructions.

tnir:
Updated Debian/Ubuntu installation docs as installation documents for Debian and Ubuntu got obsolete.

For Ubuntu, the guide for lucid may now be deleted as [it reached EOL in Apr 2015](http://fridge.ubuntu.com/2015/03/18/ubuntu-10-04-lucid-lynx-reaches-end-of-life-on-april-30-2015/).

### DIFF
--- a/docs/site/index.html
+++ b/docs/site/index.html
@@ -66,7 +66,7 @@
       <div class="tab-box active" data-tab="quickstart">
         <h1>Installation</h1>
         <p>
-          Here's a quick rundown on how to get started with gunicorn. For more detail read the documentation.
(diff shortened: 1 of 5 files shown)


berkerpeksag:
Thanks for the great patch! :)

I think we can safely delete the "before 12.04 LTS" part in the Ubuntu section.

tnir:
@berkerpeksag Thanks for the review. An additional commit to remove them was pushed.

berkerpeksag:
Thanks!

PR #1118, opened 2015-09-23
add environment variable to workers for instance identification

adamweiner:
Added support for workers to be able to uniquely identify themselves with the newly added "INSTANCE" environment variable. This can be useful to log which worker handled a specific request, or for use with services such as Consul where workers should register themselves with a unique ID.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -40,6 +40,7 @@ class Arbiter(object):
 
     LISTENERS = []
     WORKERS = {}
+    WORKER_IDS = {}
     PIPE = []
 
     # I love dynamic languages
@@ -466,6 +467,7 @@ class Arbiter(object):
                     if not worker:
                         continue
                     worker.tmp.close()
+                    del self.WORKER_IDS[wpid]
         except OSError as e:
             if e.errno != errno.ECHILD:
                 raise
@@ -494,6 +496,11 @@ class Arbiter(object):
 
     def spawn_worker(self):
         self.worker_age += 1
+        worker_id = -1
+        for i in range(self.num_workers):
+            if i not in self.WORKER_IDS.values():
+                worker_id = i
+                break
         worker = self.worker_class(self.worker_age, self.pid, self.LISTENERS,
                                    self.app, self.timeout / 2.0,
                                    self.cfg, self.log)
@@ -501,6 +508,7 @@ class Arbiter(object):
         pid = os.fork()
         if pid != 0:
             self.WORKERS[pid] = worker
+            self.WORKER_IDS[pid] = worker_id
             return pid
 
         # Process Child
@@ -509,6 +517,8 @@ class Arbiter(object):
             util._setproctitle("worker [%s]" % self.proc_name)
             self.log.info("Booting worker with pid: %s", worker_pid)
             self.cfg.post_fork(self, worker)
+            if self.num_workers > 1 and worker_id != -1:
+                os.environ['GUNICORN_WORKER_INSTANCE'] = str(worker_id)
             worker.init_process()
             sys.exit(0)
         except SystemExit:
@@ -569,6 +579,7 @@ class Arbiter(object):
                     worker = self.WORKERS.pop(pid)
                     worker.tmp.close()
                     self.cfg.worker_exit(self, worker)
+                    del self.WORKER_IDS[pid]
                     return
                 except (KeyError, OSError):
                     return


tilgovi:
I think the idea is okay, but I would set the environment variable after the fork instead of changing the signature of the base worker constructor. That would be less invasive.

adamweiner:
@tilgovi - I updated my implementation to avoid any changes to the base worker constructor: https://github.com/adamweiner/gunicorn/commit/20ebdc1edee6c77d9e98306b33370e32e5a60da4

gunicorn/arbiter.py
```diff
@@ -509,6 +516,7 @@ def spawn_worker(self):
             util._setproctitle("worker [%s]" % self.proc_name)
             self.log.info("Booting worker with pid: %s", worker_pid)
             self.cfg.post_fork(self, worker)
+            os.environ['INSTANCE'] = str(worker_id)
```
berkerpeksag: If INSTANCE is not part of a standard or a widely used best practice, I'd rename it to GUNICORN_WORKER_INSTANCE or GUNICORN_INSTANCE.

adamweiner: Good point - GUNICORN_WORKER_INSTANCE it is.(https://github.com/benoitc/gunicorn/commit/8703fffca4e700445659ca90b24a88faa38bebf0)

adamweiner:
@tilgovi @berkerpeksag anything else you'd like to see changed to give this the best shot at getting merged soon?

adamweiner:
@tilgovi @berkerpeksag anything else you'd like to see changed to give this a shot at getting merged soon?

gunicorn/arbiter.py
```diff
@@ -494,6 +496,11 @@ def manage_workers(self):
 
     def spawn_worker(self):
         self.worker_age += 1
+        worker_id = self.num_workers
+        for i in range(self.num_workers):
```
berkerpeksag: Shouldn't this be `range(1, self.num_workers+1)` instead?

gunicorn/arbiter.py
```diff
@@ -494,6 +496,11 @@ def manage_workers(self):
 
     def spawn_worker(self):
         self.worker_age += 1
+        worker_id = self.num_workers
```
berkerpeksag: We probably shouldn't set the env var If `len(self.num_workers)` is 0.

berkerpeksag:
Also, please squash the commits when you're done :)

gunicorn/arbiter.py
adamweiner: I was thinking that initializing worker_id to self.num_workers could handle some weird edge cases where a worker is started but there isn't an unclaimed id in range(0, self.num_workers). Not sure if such a case would be possible though, I haven't been able to make that happen in any of my testing.

adamweiner:
@berkerpeksag squashed with updates in response to your suggestions. I kept the range of instances starting at 0, but now no instances will ever be assigned an id outside of the proper range. It also only sets the env var when there is more than one worker.

gunicorn/arbiter.py
```diff
@@ -501,6 +508,7 @@ def spawn_worker(self):
         pid = os.fork()
         if pid != 0:
             self.WORKERS[pid] = worker
+            self.WORKER_IDS[pid] = worker_id
```
benoitc: i don't understand why you need another dict? Why not simply adding an "id" to the worker instead?

gunicorn/arbiter.py
```diff
@@ -494,6 +496,11 @@ def manage_workers(self):
 
     def spawn_worker(self):
         self.worker_age += 1
+        worker_id = -1
+        for i in range(self.num_workers):
```
benoitc: That's incredibly inefficient.  I would rather generate a unique ID there or maybe just hash the age with the current time when the worker is created so you don't have to iterate over all the running worker to find something unique.

gunicorn/arbiter.py
```diff
@@ -509,6 +517,8 @@ def spawn_worker(self):
             util._setproctitle("worker [%s]" % self.proc_name)
             self.log.info("Booting worker with pid: %s", worker_pid)
             self.cfg.post_fork(self, worker)
+            if self.num_workers > 1 and worker_id != -1:
```
benoitc: here I would test if the worker id has been set. and then set the env. Although in which case the worker_id couldn't be set there?

benoitc:
Thanks for the patch. I just reviewed it. I think it can be improved see the comments inside. Otherwise I like the id to identify the workers :)

gunicorn/arbiter.py
adamweiner: Thanks for all the feedback, @benoitc.

Despite the added overhead, I think one of the most useful parts of instance identification is having the instance number as a basic counter in a predictable range, as opposed to a hash or other unique identifier. If implementing this with a hash it wouldn't really add much to gunicorn, as each worker's PID can already act as such a unique identifier.

The use cases for this are plentiful, but could include workers logging under the same set of names every time an application is run with gunciorn (app0, app1, etc), or having workers listen on the same set of incremental ports (1000, 1001, etc). I think the predictability that a basic counter provides is pretty important here.

benoitc: why not then simply reusing worker.age which is unique and incremented once
a worker is created?

jamadden: We also had a use case for predictable ids for workers that would stay the same after restarts. This makes using a persistent zodb cache much easier.

adamweiner: The only issue with using worker.age is if a worker is reaped and a new one is spawned in its place, worker.age continues to increment and the instance id will not remain in the anticipated range.

gunicorn/arbiter.py
adamweiner: This would require adding an instance property to the worker in the base worker constructor, which I did in my initial implementation, but was recommended against above (https://github.com/benoitc/gunicorn/pull/1118#issuecomment-142725348).

tilgovi:
@adamweiner do you have enough feedback here?

adamweiner:
@tilgovi Not entirely sure what you would like me to change - my comments above (https://github.com/benoitc/gunicorn/pull/1118#discussion_r40820175, https://github.com/benoitc/gunicorn/pull/1118#discussion_r40824749) were left without response.

tilgovi:
@benoitc?

gunicorn/arbiter.py
benoitc: I am not sure I follow the reasoning there. While I understand the need of not using a random hash we can definitely have predictable hashing function. See something like hash(age + appname + ...) . Ideally the hash function could be set via the configuration with good defaults like `hash(age +  module + app)` or using the pid by default . I think it would work for your cases there and would remove any overhead appart the time took to hash the result. Thoughts?

benoitc:
@tilgovi just commented, thanks for the head-up!

gunicorn/arbiter.py
adamweiner: If just using the PID to identify, adding this feature would be unnecessary, as the instances launched by gunicorn can already use their PID as a unique identifier. Also, if the hash uses worker.age, they would change when a worker is reaped and spawned, as worker.age continues to increment as long as gunicorn is running. This would generate a number of hashes that is greater than the number of workers, which the current implementation avoids. I still believe using a basic counter covers the most use cases and keeps things relatively simple.

One alternative that may satisfy us both would be adding a new flag (ie. --identify) that would be required to enable instance identification. This would not add any overhead to gunicorn use cases that have no reason to employ instance identification.

tilgovi: Having so many workers on one machine that this would actually be a performance issue is not a use case I'm familiar with, but I accept that it's possible.

Would just using a priority queue for the worker IDs be better?

Another thing that just occurred to me. If you TTOU signal and cause the number of workers to drop is it necessary to kill the worker with the highest ID, or is it okay to then have some worker(s) around with IDs higher than the number of workers?

benoitc:
Thanks for all the feedbacks and code but IMO this feature needs more discussion. Closing then the PR in favor of #1278.

Issue #1056, opened 2015-06-25
monkey.patch_all and gunicorn with more than 1 worker


emamirazavi:
I'm using Flask+Gunicorn+Gevent
I've called monkey.patch_all at the first of app passed to gunicorn command line because of using some co-routine tasks in my app. 
My worker class is gunicorn.workers.ggevent.GeventWorker.
When i configure gunicorn to use more than one worker (multiprocessing.cpu_count() \* 2 + 1 that in my laptop is equal to 9) everything is mixed up. My logger (python logging) logs 10 times! and some much more others conflicts in my code occurred.
MY GUNICORN CONFIGURAION

```
# to import server_hook
import os, sys
this_path = os.path.dirname(os.path.realpath(__file__))
sys.path.insert(0, os.path.realpath(this_path))
import server_hook
import multiprocessing
user = "myapp"
max_requests = 60
max_requests_jitter = 5
reload = True
group = "daemon"
host = "0.0.0.0"
bind = ['127.0.0.1:8000', 'unix:/run/gunicorn/socket']
workers = multiprocessing.cpu_count() * 2 + 1
threads = 10
worker_class = "gunicorn.workers.ggevent.GeventWorker"
proc_name = "myapp"
default_proc_name = "myapp"
environment = LANG = "en_US.UTF-8"
preload_app = True
loglevel = "DEBUG"
on_starting = server_hook.on_starting
```

MY SERVER_HOOK FILE (server_hook.on_starting)

```
import threading
from myapp.logger import logger

def on_starting(server):
    thread = threading.Thread(target=test)
    thread.daemon = True
    thread.start()

def test():
    while 1:
        logger.debug('i am test!')
        sleep(10)
```

MY LOG FILE

```
2015-06-25 20:32:33,579 INFO Starting webserver...
2015-06-25 20:32:33,595 DEBUG i am test!

2015-06-25 20:32:43,595 DEBUG i am test!
2015-06-25 20:32:43,596 DEBUG i am test!
2015-06-25 20:32:43,596 DEBUG i am test!
2015-06-25 20:32:43,596 DEBUG i am test!
2015-06-25 20:32:43,595 DEBUG i am test!
2015-06-25 20:32:43,596 DEBUG i am test!
2015-06-25 20:32:43,597 DEBUG i am test!
2015-06-25 20:32:43,597 DEBUG i am test!
2015-06-25 20:32:43,597 DEBUG i am test!
[...]

PR #1121, opened 2015-10-03
Update run.rst

emamirazavi:
According to the issue #1056
Please give me a feedback if the descriptions are partial.

### DIFF
--- a/docs/source/run.rst
+++ b/docs/source/run.rst
@@ -64,6 +64,10 @@ Commonly Used Arguments
   adjust the name of Gunicorn process as they appear in the process system
   table (which affects tools like ``ps`` and ``top``).
 
+.. Gevent note::
+   You must not call monkey.patch_all because it leads to some problems when using ``gevent`` as your ``WORKERCLASS``.
+   To do co-routine tasks in your application write one another file and run it alonside the Gunicorn.
+
 See :ref:`configuration` and :ref:`settings` for detailed usage.
 
 .. _setproctitle: http://pypi.python.org/pypi/setproctitle/


docs/source/run.rst
```diff
@@ -64,6 +64,10 @@ Commonly Used Arguments
   adjust the name of Gunicorn process as they appear in the process system
   table (which affects tools like ``ps`` and ``top``).
 
+.. Gevent note::
```
berkerpeksag: I know @benoitc suggested run.rst, but I'd prefer to see this in http://docs.gunicorn.org/en/latest/settings.html#worker-class

berkerpeksag:
```rst
.. note::
   Foo bar [...]
```

benoitc: that's probably better indeed

emamirazavi: No. This problem is not just with Gevent worker. Generally using
monkey.patch_all leads to problems. If my view is correct Benoit's
suggestion to add in run.rst is better. Give me feedback again.

berkerpeksag: Well, then the current wording needs to be improved since `monkey.patch_all()` is only true for Gevent(for example, Eventlet's API is `eventlet.monkey_patch()`).  Another option is creating a new sub-section at http://docs.gunicorn.org/en/latest/run.html#integration and add all information there regarding to async workers.

emamirazavi: It might be problem with others APIs like eventlet.monkey_patch() but I've
never tested it until now. Have you?

tilgovi: It should work the same way.

I also think maybe we want to mention this in the async worker documentation instead, and maybe a note in the documentation for the `preload_app` setting.

benoitc:
what actions does this ticket need?

emamirazavi:
I've added the documentation mentioned in run.rst, see pull requests but if
you want to move it to other rst file give me a feedback to do. Good luck
Benoit.

benoitc:
do we need any more changes there?

benoitc:
cc @berkerpeksag @tilgovi

berkerpeksag:
Here is an alternative patch:

```diff
diff --git a/gunicorn/config.py b/gunicorn/config.py
index a1aa456..fed75b3 100644
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -570,6 +570,13 @@ class WorkerClass(Setting):
         * ``gevent``   - Requires gevent >= 0.13
         * ``tornado``  - Requires tornado >= 0.2
 
+        .. note:
+           If you want to use ``'gevent'`` and ``'eventlet'`` workers, avoid
+           monkeypatching (e.g. calling the ``monkey.patch_all()`` and
+           ``eventlet.monkey_patch()`` functions respectively) in your
+           application. Gunicorn will do it for you after the ``on_starting``
+           hook ran.
+
         Optionally, you can provide your own worker by giving Gunicorn a
         Python path to a subclass of ``gunicorn.workers.base.Worker``.
         This alternative syntax will load the gevent class:
@@ -819,6 +826,11 @@ class PreloadApp(Setting):
         speed up server boot times. Although, if you defer application loading
         to each worker process, you can reload your application code easily by
         restarting workers.
+
+        .. note::
+           Using this setting with the ``'gevent'`` and ``'eventlet'`` workers
+           is discouraged as it will prevent properly mockeypatching the
+           Python standard library.
         """
 
 class Sendfile(Setting):
```

I open to any suggestions to improve the wording. I can publish as a branch or pull request if that would help to review it.

benoitc:
@berkerpeksag your changes look good go ahead, happy to do the final review on a pr if you need it :)

berkerpeksag:
Superseded by https://github.com/benoitc/gunicorn/pull/1431. Thanks for the PR, @emamirazavi!

PR #1127, opened 2015-10-15
Respect defaults when running a paster application

nickstenning:
The values of both the 'workers' and 'umask' settings have global defaults. In the case of 'workers', this global default can be overridden by the WEB_CONCURRENCY environment variable, which is currently ignored when running an application with `gunicorn --paste`.

Removing this duplication of defaults allows `gunicorn --paste` to respect settings from:

- the environment (WEB_CONCURRENCY)
- the paste configuration file
- CLI flags

### DIFF
--- a/gunicorn/app/pasterapp.py
+++ b/gunicorn/app/pasterapp.py
@@ -44,8 +44,6 @@ def paste_config(gconfig, config_url, relative_to, global_conf=None):
     elif host:
         cfg['bind'] = host.split(',')
 
-    cfg['workers'] = int(lc.get('workers', 1))
-    cfg['umask'] = int(lc.get('umask', 0))
     cfg['default_proc_name'] = gc.get('__file__')
 
     # init logging configuration


benoitc:
Well I guess that people that are still using tools like paster need to use what is coming from the ini file, here the global conf. Why not instead putting the default from the global config?

nickstenning:
> Why not instead putting the default from the global config?

That's already done later in this function. This change just ensures that in the absence of any other setting we fall back to the environment variable.

nickstenning:
q.v. https://github.com/benoitc/gunicorn/pull/1127/files#diff-ed33522d62e9150c02ccd2ce1f1646e8R54

benoitc:
I don't think it is. With is change variables from the ini file in the global section won't be taken in consideration if they are set. Or did I miss anything?

benoitc:
I meant the local config

nickstenning:
Lines 54-62 (right-hand side of the diff) do exactly this. With this patch the priority is:

CLI flag > .ini file > env var > global default from `gunicorn.config`

benoitc:
right :) :+1:

PR #1126, opened 2015-10-15
Remove time.sleep for gevent

xtao:
When using gevent worker, time.sleep will block under gevent.
Please review, thanks.

Sample output:

```
[2015-10-15 11:02:45 +0800] [2526] [INFO] Starting gunicorn 19.3.0
[2015-10-15 11:02:45 +0800] [2526] [INFO] Listening at: http://0.0.0.0:8080 (2526)
[2015-10-15 11:02:45 +0800] [2526] [INFO] Using worker: dae.gworkers.ggevent.GeventWorker
[2015-10-15 11:02:45 +0800] [2529] [INFO] Booting worker with pid: 2529
[2015-10-15 11:02:45 +0800] [2530] [INFO] Booting worker with pid: 2530
^CTraceback (most recent call last):
  File "/Users/xtao/Work/douban/dae/venv/src/gunicorn/gunicorn/workers/base.py", line 158, in handle_quit
Traceback (most recent call last):
    time.sleep(0.1)
  File "/Users/xtao/Work/douban/dae/venv/lib/python2.7/site-packages/gevent/hub.py", line 75, in sleep
  File "/Users/xtao/Work/douban/dae/venv/src/gunicorn/gunicorn/workers/base.py", line 158, in handle_quit
    time.sleep(0.1)
  File "/Users/xtao/Work/douban/dae/venv/lib/python2.7/site-packages/gevent/hub.py", line 75, in sleep
    hub.wait(loop.timer(seconds, ref=ref))
  File "/Users/xtao/Work/douban/dae/venv/lib/python2.7/site-packages/gevent/hub.py", line 341, in wait
    result = waiter.get()
  File "/Users/xtao/Work/douban/dae/venv/lib/python2.7/site-packages/gevent/hub.py", line 568, in get
    return self.hub.switch()
  File "/Users/xtao/Work/douban/dae/venv/lib/python2.7/site-packages/gevent/hub.py", line 330, in switch
    switch_out()
  File "/Users/xtao/Work/douban/dae/venv/lib/python2.7/site-packages/gevent/hub.py", line 334, in switch_out
    raise AssertionError('Impossible to call blocking function in the event loop callback')
AssertionError: Impossible to call blocking function in the event loop callback
    hub.wait(loop.timer(seconds, ref=ref))
  File "/Users/xtao/Work/douban/dae/venv/lib/python2.7/site-packages/gevent/hub.py", line 341, in wait
    result = waiter.get()
[...]

PR #1128, opened 2015-10-15
Make the gevent workers handle the quit signal by deferring to a new greenlet

jamadden:
As discussed in #1126. This fixes the problem interactively. However, I have no idea how to add a unit test; there doesn't seem to be anything similar in place already. Suggestions welcome!

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -159,3 +159,4 @@ Marc Abramowitz <msabramo@gmail.com>
 Hebert J <hebert@mail.ru>
 Kevin Littlejohn <kevin@littlejohn.id.au>
 Wolfgang Schnerring <wosc@wosc.de>
+Jason Madden <jason@nextthought.com>
--- a/gunicorn/workers/ggevent.py
+++ b/gunicorn/workers/ggevent.py
@@ -166,6 +166,11 @@ class GeventWorker(AsyncWorker):
         except SystemExit:
             pass
 
+    def handle_quit(self, sig, frame):
+        # Move this out of the signal handler so we can use
+        # blocking calls. See #1126
+        gevent.spawn(super(GeventWorker, self).handle_quit, sig, frame)
+
     if gevent.version_info[0] == 0:
 
         def init_process(self):


tilgovi:
Hmmm. I'm actually wondering if we shouldn't just remove the `time.sleep()`. I know I tried to explain why I _thought_ we had it, but I may not actually agree with it :). What do you think, @jamadden, or anyone else?

jamadden:
I'm +1 for simply removing `time.sleep`. It's not like it's been working, and we haven't noticed any particular issues simply because the worker exits without sleeping.

xtao:
It works, thanks.

benoitc:
@tilgovi it's here to let some times to the [VM to release locks](https://github.com/benoitc/gunicorn/commit/a459986cfd2d5d04dd07126911eeedbfcd6657de). 

Patch looks good. However I am wondering if there isn't a way to patch the signal handler when launching the worker.  Indeed gevent.spawn is quite the same as  a gevent.sleep or time.sleep when patched. I think it would worth to investigate this possibility imo.

tilgovi:
Gevent < 1.1 does not monkey patch signal, but even with 1.1 that patches `signal.signal` and `time.sleep` the same error occurs.

jamadden:
@tilgovi Signals are delivered to the main greenlet, which happens to be the one running the hub---so the net result is the same on 1.0. So a complete stack trace captured there actually looks like this (captured with pdb):
```
[2015-10-16 07:34:43 -0500] [20467] [INFO] Booting worker with pid: 20467
^C[2015-10-16 07:34:45 -0500] [20464] [INFO] Handling signal: int
> //gunicorn/gunicorn/workers/ggevent.py(175)handle_quit()
-> gevent.spawn(super(GeventWorker, self).handle_quit, sig, frame)
(Pdb) bt
  //lib/python2.7/site-packages/gevent/hub.py(371)run()
-> loop.run()
  //gunicorn/gunicorn/workers/ggevent.py(174)handle_quit()
-> pdb.set_trace()
```

Under 1.1, monkey patching signal only has any effect for SIGCHLD; everything else is left untouched and behaves just as it did in 1.0. 

@benoitc `gevent.spawn` is *not* the same thing as `gevent.sleep` or `time.sleep`. Both of the latter are blocking functions, while `spawn` is not. `spawn` is [explicitly documented](http://www.gevent.org/intro.html#event-loop) as safe to use in situations like this:

>  Note that the callbacks supplied to the libev API are run in the gevent.hub.Hub greenlet and thus cannot use the synchronous gevent API. It is possible to use the asynchronous API there, like gevent.spawn() and gevent.event.Event.set().

gevent does have a function [gevent.signal](http://www.gevent.org/gevent.html#signals) that has the same effect as this patch (runs the handler in a new greenlet), but that API must be invoked directly---it is not monkey-patched in for compatibility.

jamadden:
@benoitc I'm curious, can you share some more details about what locks you're talking about? And why they wouldn't be automatically cleaned up by interpreter shut-down? Is this something gunicorn specific?

jamadden:
Any more thoughts on this?

tilgovi:
From what I can tell, gevent 1.1 _does_ monkey patch `signal.signal` so I'm a bit confused why this problem still exists there.

jamadden:
Yes, there is a monkey patch on signal.signal in 1.1, but it *only* handles the case where the first argument is `SIGCHLD`. Everything else is passed through to the standard library, including `SIGQUIT` and `SIGINT`. So this case is unaffected by the patch. gevent's code literally looks like this:

```python
import signal as _signal
_signal_signal = _signal.signal # capture before patch
def signal(signalnum, handler):
    """
    Exactly the same as :func:`signal.signal` except where
    :const:`signal.SIGCHLD` is concerned.
    """
    if signalnum != _signal.SIGCHLD:
        return _signal_signal(signalnum, handler)
   ...
```

tilgovi:
Oh, right. You said that. Sorry.

tilgovi:
I think this PR looks good enough to me.

tilgovi:
I'm not hearing any objection. Pushing the button.

ant1g:
I am having the exact same issue, but for the **handle_us1**... Shall create a ticket? The solution is going to be the same as for this issue. I can create PR if needed.

tilgovi:
@ant1g go for it! Posting a traceback would be helpful, too, in case there's something else going on that someone can spot.

ant1g:
OK, I have created: https://github.com/benoitc/gunicorn/issues/1645

PR #1131, opened 2015-10-18
Enable SO_REUSEPORT socket option.

smilliken:
Needs:
* Testing on Python <2.7.9 and 3.*
* Configuration option.
* Testing on non-linux systems?

### DIFF
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -38,6 +38,11 @@ class BaseSocket(object):
 
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        try:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        except socket.error as ex:
+            if ex.errno == 92:
+                self.log.warning('SO_REUSEPORT socket option not supported.')
         if not bound:
             self.bind(sock)
         sock.setblocking(0)


gunicorn/sock.py
```diff
@@ -38,6 +38,11 @@ def __getattr__(self, name):
 
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        try:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        except socket.error as ex:
+            if ex.errno == 92:
```
berkerpeksag: I'd use `errno.ENOPROTOOPT` instead of `92`.

berkerpeksag: Also, perhaps we should raise the exception if `exc.errno` is not `92`.

gunicorn/sock.py
```diff
@@ -38,6 +38,11 @@ def __getattr__(self, name):
 
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        try:
+            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        except socket.error as ex:
```
berkerpeksag: Can you please rename `ex` to `e` to be consistent with the rest of the file?

benoitc:
@smilliken interested by making these changes?

gunicorn/sock.py
benoitc: @berkerpeksag in which case it isn't?

berkerpeksag: Even if it returns other than ENOPROTOOPT (e.g. EBADF, EINVAL), we could catch them in line 40 so ignore my last comment :)

tilgovi:
There's been some silence here from @smilliken. If we're happy with this, we should just make the small changes and merge a copy of it, with attribution.

berkerpeksag:
+1

pypeng:
+1

benoitc:
which changes are needed? If someone can update the PR so we can merge it it would be helpful :)

wooparadog:
I have some doubt about adding reuseport support for gunicorn. 

`SO_REUSEPORT` is primarily used for using the same listening port for two unrelated processes. But in gunicorn's case, all workers are forked from master, AFTER master opening the listening port, so all workers are already using the same port.

The only use case I can think of is (even more) graceful restart(starting another gunicorn process tree, using SO_REUSEPORT, then kill the old one), but gunicorn already has support for reloading worker processes...

But of course, adding `SO_REUSEPORT` won't hurt..

berkerpeksag:
Superseded by https://github.com/benoitc/gunicorn/pull/1344.

PR #1069, opened 2015-07-07
Getting username from basic auth header for access logs

brk0v:
Logging Basic auth username to access log:

```
curl -u brk0v: -i -k https://api.example.com/v1/api-call
```

Logs:

```
127.0.0.1 - brk0v [07/Jul/2015:23:37:34 +0300] "GET /v1/api-call HTTP/1.1" 200 307 "-" "curl/7.21.3 libcurl/7.21.3"
```

PR #1133, opened 2015-10-21
Fix config description of accesslog user-name params

bungoume:
Username parameter has been implemented in https://github.com/benoitc/gunicorn/pull/1069 .


berkerpeksag:
Thanks!

PR #1141, opened 2015-11-04
Fix worker class link in FAQ

irachex:
Fix the _worker_class_ link at http://docs.gunicorn.org/en/19.3/faq.html#worker-processes


berkerpeksag:
Thanks!

PR #1144, opened 2015-11-10
Docs: Fix the link to the proc_name setting on the FAQ page

edmorley:
Fixes the link to the `proc_name` setting on the FAQ [here](http://docs.gunicorn.org/en/19.3/faq.html#how-can-i-name-processes), since it no longer exists on the configure page.


berkerpeksag:
Good catch, thanks Ed!

edmorley:
No problem - thank you :-)

Issue #1073, opened 2015-07-09
Paste entry point fixed to 'main'


dmdm:
Hello,
currently, it is impossible to start gunicorn --paste with a different app name than 'main', e.g. --paste development.ini#foo does not work.
It seems, https://github.com/benoitc/gunicorn/blob/master/gunicorn/app/pasterapp.py#L35 is the reason. If I comment this line out, #foo works for me.

tilgovi:
According to this commit, it should work: e5442332fe16ffa92cd4516c68ef45d93cc57349

The line you commented out should only affect the server that gets loaded, not the app. I'll take another look, though, when I can.

tilgovi:
It seems to work fine here. Can you show your .ini file and how you call gunicorn?

dmdm:
Thank you for looking into it. My INI is like this:
```
[app:main]
use = egg:MyProject

pyramid.includes =
    #pyramid_debugtoolbar
	pyramid_tm

[server:main]
use = egg:waitress#main
host = 0.0.0.0
port = 6543

#----------------------------------------------------------------------------

[app:socksrv]
use = egg:MyProject#socksrv

pyramid.includes =
	pyramid_tm

[server:socksrv]
use = egg:gunicorn#main
host = 127.0.0.1
port = 6544
workers = 1
worker_class = guv.GuvWorker
```
If the mentioned line is active (and cuts off he '#name'-part),
```
gunicorn --paste development.ini#socksrv
```
starts "main". With that line commented out, "socksrv" is started correctly.

tilgovi:
I use Pyramid also. I'm surprised that your .ini works. For me, with multi-line `pyramid.includes` I have to make sure the `#` is the first character of the line (before any space) or it isn't treated as a comment.

But, let's assume that's not the issue at all. I can't reproduce this at all on Gunicorn 19.3.0. It works just fine.

What version of Gunicorn are you using?

dmdm:
The same. Just tried it again and got this log, confirming the error:
With commented out line, gunicorn works as expected:
```
$ gunicorn  --paste development.ini#socksrv
[2015-07-10 07:41:34 +0200] [13824] [INFO] Starting gunicorn 19.3.0
[2015-07-10 07:41:34 +0200] [13824] [INFO] Listening at: http://127.0.0.1:6544 (13824)
[2015-07-10 07:41:34 +0200] [13824] [INFO] Using worker: guv.GuvWorker
```
and with it active:
```
[...]

PR #1146, opened 2015-11-16
Fix ignoring specific paster server section

canni:
If config was given with specific app, eg: `development.ini#myotherapp`
gunicorn will load `[sever:main]` section anyway, this changes it to
look for `[server:myotherapp]` instead.


tilgovi:
This has already been discussed at length in #1073.

This syntax is not enough to provide the app and the server. We need another option for the server name because this syntax is already used for selecting the app block.

Issue #1148, opened 2015-11-19
Authorization header parsing can't handle unconventional values


roosmaa:
The [_get_user](https://github.com/benoitc/gunicorn/blob/c8f820b95a559aec81167c854d7b0d869316ecb3/gunicorn/glogging.py#L381) method assumes that Basic authorization method is always used in the Authorization header. This is not true, especially for APIs which can have their own token format (not necessarily b64 encoded).

For instance `Authorization: Bearer whatever` header results in (Python 3.5; current master):

```
Traceback (most recent call last):
  File "/usr/src/deps/gunicorn/gunicorn/workers/async.py", line 116, in handle_request
    self.log.access(resp, req, environ, request_time)
  File "/usr/src/deps/gunicorn/gunicorn/glogging.py", line 287, in access
    request_time))
  File "/usr/src/deps/gunicorn/gunicorn/glogging.py", line 243, in atoms
    'u': self._get_user(environ) or '-',
  File "/usr/src/deps/gunicorn/gunicorn/glogging.py", line 392, in _get_user
    auth = auth.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc2 in position 0: invalid continuation byte
```

Depending on the input header value, the b64 operation itself can fail as well:

```
Traceback (most recent call last):
  File "/usr/src/deps/gunicorn/gunicorn/workers/async.py", line 116, in handle_request
    self.log.access(resp, req, environ, request_time)
  File "/usr/src/deps/gunicorn/gunicorn/glogging.py", line 287, in access
    request_time))
  File "/usr/src/deps/gunicorn/gunicorn/glogging.py", line 243, in atoms
    'u': self._get_user(environ) or '-',
  File "/usr/src/deps/gunicorn/gunicorn/glogging.py", line 390, in _get_user
    auth = base64.b64decode(auth[1].strip().encode('utf-8'))
  File "/usr/local/lib/python3.5/base64.py", line 90, in b64decode
    return binascii.a2b_base64(s)
binascii.Error: Incorrect padding
```

I suggest checking that the auth type is `Basic` before trying to do base64 operations on the 2nd part of the header. Handling those possible exceptions would also be nice, since someone might not know that `Basic` headers need to be base64 encoded.

tilgovi:
:+1: we should fix this

- Should never cause a crash
- Should be quiet except maybe to warn once if the format seems like bad use of basic auth

benoitc:
[...]

PR #1149, opened 2015-11-20
check if the auth type is Basic before getting the user name

benoitc:
a possible fix to #1148 . let me know


benoitc:
@tilgovi do you think this fix is enough? Maybe we should also completely ignore possible errors there?

PR #1165, opened 2015-12-16
Let reloader survive NameError in wsgi app

kvikshaug:
This small change adds `NameError` to the [exceptions handled](https://github.com/benoitc/gunicorn/blob/58108ef7e6836fc545f0614846ffed61f34c48fd/gunicorn/workers/base.py#L131) during wsgi loading [based on this earlier commit](https://github.com/benoitc/gunicorn/commit/d03891a4706dc3106f3d51074e0e6b14cbb8ca0c) ([pull request](https://github.com/benoitc/gunicorn/pull/994)).

The intention is to avoid the master gunicorn process exiting during development with `reload=True` if one were to introduce an error in the wsgi application which causes a `NameError` (e.g. adding a random string "asdf" to the end of a source file).

Note that since a `exceptions.NameError` has no `filename` attribute, the `self.reloader.add_extra_file(exc_val.filename)` line is skipped in this case. I'm not 100% sure of the implications of this and that should be reviewed; however it did not introduce any unit test failures.

### DIFF
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -128,14 +128,15 @@ class Worker(object):
     def load_wsgi(self):
         try:
             self.wsgi = self.app.wsgi()
-        except SyntaxError as e:
+        except (SyntaxError, NameError) as e:
             if not self.cfg.reload:
                 raise
 
             self.log.exception(e)
 
             exc_type, exc_val, exc_tb = sys.exc_info()
-            self.reloader.add_extra_file(exc_val.filename)
+            if hasattr(exc_val, 'filename'):
+                self.reloader.add_extra_file(exc_val.filename)
 
             tb_string = traceback.format_exc(exc_tb)
             self.wsgi = util.make_fail_app(tb_string)


tilgovi:
looks good to me

tilgovi:
Actually, I have another idea. What do you think?

    index 92e6dae..2c50012 100644
    --- a/gunicorn/arbiter.py
    +++ b/gunicorn/arbiter.py
    @@ -525,6 +525,8 @@ class Arbiter(object):
             except:
                 self.log.exception("Exception in worker process:\n%s",
                                    traceback.format_exc())
    +            if self.cfg.reload:
    +                return
                 if not worker.booted:
                     sys.exit(self.WORKER_BOOT_ERROR)
                 sys.exit(-1)

The worker will be restarted about once a second and the arbiter just won't quit in this situation.

tilgovi:
This way, any error starting the application will just be logged repeatedly if the reloader is active, and application will try to start again.

kvikshaug:
Looks like a much better idea to me. Presumably, that implementation suggests that [this commit](https://github.com/benoitc/gunicorn/commit/d03891a4706dc3106f3d51074e0e6b14cbb8ca0c) could be reverted as well.

benoitc:
what should se do there? Also tests don't pass.

kvikshaug:
Closing this PR in favor of [this idea](https://github.com/benoitc/gunicorn/pull/1165#issuecomment-166068993).

Winand:
This issue hasn't been fixed?

benoitc:
@Winand please don't [comment on closed issues](https://github.com/benoitc/gunicorn/blob/master/CONTRIBUTING.md#conventions) . If you have still an issue please open a new ticket and put any logs you have and possibly a way to reproduce it.

PR #1168, opened 2015-12-28
fix sendfile option validation

benoitc:
sendfile option don't set any default for now which makes the validation fail
(and then tests) since no boolean is given. For now just return when the value
is not set.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -281,6 +281,9 @@ Setting = SettingMeta('Setting', (Setting,), {})
 
 
 def validate_bool(val):
+    if val is None:
+        return
+
     if isinstance(val, bool):
         return val
     if not isinstance(val, six.string_types):


benoitc:
#1169 makes travis-ci return an OK. gaiohttp test is still broken on this branch

benoitc:
@tilgovi oups pushed a new version, i missed to commit a part in the config.py on which I moved part of your previous diff. Let me know if it's ok for you.

benoitc:
though maybe we should just keep your change instead of introducing a new property... Anyway let me know :)

benoitc:
nvm I'm tired :) pushed the third version just fixing the test.

tilgovi:
Okay :-). We can do more another time.

benoitc:
ok :) merging it. will make the release asap. thanks for the review!

PR #1175, opened 2015-12-31
reuse util.is_fileobject

### LINKED ISSUE #1174: Correctly handle unavailable .fileno() support for io.BytesIO instances
The changes introduced in https://github.com/benoitc/gunicorn/commit/98c9e3b3756717771fd6521dbb2dbec206ec5fc5#diff-6607b435d6ee868f251dbdd1a6d0edecL357 are wrong and lead to crashes for `io.BytesIO` instances which do have a `.fileno()` method but raise another exception instead:

```
>>> import io
>>> io.BytesIO().fileno()
  File "<ipython-input-3-7062c09d0c52>", line 1, in <module>
    io.BytesIO().fileno()
UnsupportedOperation: fileno
```

This exception was already caught properly in the earlier code, a few lines below.

benoitc:
`util.is_fileobject` usage was removed due to the use of the `tell` method check.
This change remove this check wich allows us to completely test if
fileno() is usable. Also it handle most of the exceptions around created by
breaking changes across Python versions. Hopefully we are good now.

fix #1174

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -353,7 +353,7 @@ class Response(object):
         if self.cfg.is_ssl or not self.can_sendfile():
             return False
 
-        if not util.is_fileobject(respiter.filelike):
+        if not util.has_fileno(respiter.filelike):
             return False
 
         try:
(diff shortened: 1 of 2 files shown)


wbolster:
for the record, #1176 is an alternative fix

PR #1176, opened 2015-12-31
Catch io.UnsupportedOperation exception for io.BytesIO.fileno()

### LINKED ISSUE #1147: gunicorn 19 with WebSocket on Heroku
I'm investigating a strange problem that occurs when using gunicorn 19.0.0 or newer with an application that uses my Flask-SocketIO server, which upgrades client connections to WebSocket. 

The issue, as observed from the client side, is that the connection is extremely slow. Upon inspection of the heroku logs, the slowness appears to be caused by constant timeouts and reconnects. The timeouts occur when using both the eventlet and gevent workers, each using a completely different WebSocket implementation, which I think eliminates eventlet, gevent and the server-side WebSocket code as a possible culprits. You can see detailed heroku logs on [this bug report](https://github.com/miguelgrinberg/Flask-SocketIO/issues/171).

I discovered that switching to gunicorn 18 makes the problem disappear completely. With 18 the WebSocket connection is rock solid.

I have also found that using any of the 19.x releases locally (without a reverse proxy in front of it) works fine, so it appears the problem is related to having the Heroku router/reverse proxy in front of gunicorn.

I went through the list of changes in release 19.0.0 and can't really see anything that could be causing this. Has anyone seen this problem before? Any idea what can cause it?

Edit: forgot to mention that in all local and heroku tests to troubleshoot this issue I used Python 2.7.10.

wbolster:
See #1174.

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -355,7 +355,7 @@ class Response(object):
 
         try:
             fileno = respiter.filelike.fileno()
-        except AttributeError:
+        except (AttributeError, io.UnsupportedOperation):
             return False
 
         try:


benoitc:
I have a preference for #1175 sinces it is also testing if the fileno method is usable and is available in a testable function, but let's wait for @tilgovi for a final answer :)

benoitc:
thanks for the path and the ticket! Similar patch but slightly more complete via #1175 has been merged.

PR #1182, opened 2016-01-09
Fixed log message for listener reloading

### LINKED ISSUE #1181: Error in log message when changing listeners on reload
When arbiter reloads workers on SIGHUP, and if old address is different from new address, it creates new sockets and logs about that. (https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L402)

    self.log.info("Listening at: %s", ",".join(str(self.LISTENERS)))

it takes a `list`, converts it to `str` and then joins resulting string with `,`. Resulting log message is as follows:

    [INFO] Listening at: [,<,g,u,n,i,c,o,r,n,.,s,o,c,k,.,T,C,P,S,o,c,k,e,t, ,o,b,j,e,c,t, ,a,t, ,0,x,7,f,8,e,3,d,2,1,f,7,8,0,>,]

Seems that this line should look like this:

    self.log.info("Listening at: %s", ",".join(map(str, self.LISTENERS)))

Or maybe it should be made like on line 132 (in `Arbiter.start`).

MarSoft:
Fixes #1181

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -399,7 +399,8 @@ class Arbiter(object):
             [l.close() for l in self.LISTENERS]
             # init new listeners
             self.LISTENERS = create_sockets(self.cfg, self.log)
-            self.log.info("Listening at: %s", ",".join(str(self.LISTENERS)))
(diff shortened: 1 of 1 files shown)


benoitc:
Thanks!

Issue #1187, opened 2016-01-22
Allow statsd and custom logger_class


bloodearnest:
Currently, if you configure statsd_host, you cannot customise the logger_class, gunicorn will just use hardcoded gunicorn.intrument.Statsd class.

https://github.com/benoitc/gunicorn/blob/master/gunicorn/config.py#L150

This is not great, as the user's configured logger class is silently swallowed up and not used. 

Could we at least add a warning if the the user has configured logger_class and statsd_host, so at least we know what's happening? :)

A simple fix would be to always prefer the user configured logger_class? And maybe improve docs to mention the Statsd class in logger_class config?

benoitc:
What do you mean by customize? Do you want to have both , ie. logging on file or std + statsd?

We should definitely make it clear when which Logging class is in use. What about printing that information on startup? 

For the rest I don't know, I guess the answer will depends on the question above ;) I do think we should be able to mix different logging strategy, the current implementation should be then improved. Thoughts?

bloodearnest:
I mean I want to configure logging_class to use my own implementation[1], but I also want to use the statsd functionality and config.

I inherited my custom class from gunicorn.instrument.Statsd, but I discovered my class was not being being used when I also configured statsd_host. Works fine if I don't configure statsd_host.

e.g. (for clarity, apologies if obvious)

# log output is as I expect
gunicorn app --logger-class='myclass'  

# log output is default gunicorn, myclass is not being used
gunicorn app --logger-class='myclass'  --statsd_host=localhost:12345



A possible fix (unfinished): 

https://github.com/benoitc/gunicorn/pull/1188

Regards mixing different logging strategy, could we possibly merge the Statsd logger functionality into the base Logger class, this removing this switching logic all together? Some care would need to be taken to avoid a performance hit if statsd is not enabled, but I think that's probably doable.

I'd be happy to do the work if you think that's worth pursuing.

Thanks (for a prompt response, and for gunicorn in general :)



[1] FWIW, my motivations for providing a custom logger_class implementation are:

a) increase timestamp resolution to include msec on access logs (not possible AFAICS with out overriding Logger.now())

b) use a custom formatter on the error log, which is currently hardcoded to logging.Formatter(Logger.error_fmt, Logger.datefmt). This to add some structured logging tags, like a request id, if present.

Both of these could maybe be addressed by extending the logger class, if you think they're worth doing. Then I wouldn't need a custom Logger implementation at all :)

[...]

PR #1188, opened 2016-01-22
Allow configuring logger_class with statsd_host

bloodearnest:
Currently if you configure statsd_host, a configured logger_class will never be used.

I think this makes a user configured logger class always take priority.

(This is PoC change, I will come back and add tests/docs if it's worth pursuing)

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -616,7 +616,10 @@ statsd_host
 ``host:port`` of the statsd server to log to.
 
 Note: enabling this switches the default *logger_class* to
-``gunicorn.instrument.statsd.Statsd``
+``gunicorn.instrument.statsd.Statsd``. If you wish to use statsd with
+a custom *logger_class*, you should make sure your class is API
+compatible with ``gunicorn.instrument.statsd.Statsd``, or inherit from
+it.
 
 .. versionadded:: 19.1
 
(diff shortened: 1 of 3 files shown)


bloodearnest:
This is a proposed fix for #1187

tilgovi:
Is it the case that a user should probably not specify both arguments together? Should we issue a warning when both are provided?

bloodearnest:
On 2 February 2016 at 23:37, Randall Leeds <notifications@github.com> wrote:

> Is it the case that a user should probably not specify both arguments
> together? Should we issue a warning when both are provided
>

My use case is exactly that, though.

I did look at adding a warning about makeing sure that your custom logger
class inherited from the statsd logger, when both both are specified. But
since logging has not yet been initialised (by definition), I wasn't sure
how to do that.  Is there a way to defer a warning until logging is
configured?

Thanks

-- 
Simon

>


-- 
Simon

tilgovi:
I think this is fine. What do you think about maybe where you have the "N.B." in the docs for the statsd option to add a note that any custom logger class specified should be API-compatible with the statsd logger. That seems like a reasonable place to document this and easier than trying to issue a warning.

This is great, thank you.

bloodearnest:
On 11 February 2016 at 21:26, Randall Leeds <notifications@github.com>
wrote:

> I think this is fine. What do you think about maybe where you have the
> "N.B." in the docs for the statsd option to add a note that any custom
> logger class specified should be API-compatible with the statsd logger.
> That seems like a reasonable place to document this and easier than trying
> to issue a warning.
>
Good idea, done.


> This is great, thank you.
>

Thank *you* for taking the time to look at it :)


-- 
Simon

bloodearnest:
Rebased on master, be great if you could land

tilgovi:
Beautiful. Thanks for your effort and patience, @bloodearnest!

PR #1193, opened 2016-01-27
fix access logging in gaiohttp worker

urbaniak:
Access logs were wrongly formatted when using gaiohttp worker

It also fixes the statsd instrumentation with this type of worker.

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -237,7 +237,9 @@ class Logger(object):
     def atoms(self, resp, req, environ, request_time):
         """ Gets atoms for log formating.
         """
-        status = resp.status.split(None, 1)[0]
+        status = resp.status
+        if isinstance(status, str):
+            status = status.split(None, 1)[0]
         atoms = {
             'h': environ.get('REMOTE_ADDR', '-'),
             'l': '-',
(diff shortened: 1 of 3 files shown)


gunicorn/glogging.py
```diff
@@ -237,7 +237,10 @@ def log(self, lvl, msg, *args, **kwargs):
     def atoms(self, resp, req, environ, request_time):
         """ Gets atoms for log formating.
         """
-        status = resp.status.split(None, 1)[0]
+        if type(resp.status) is int:
```
benoitc: I would do :

status = resp.status
if isinstance(status, str): 
     status = status.split(None, 1)[0]

to be consistent with the style in gunicorn. Thoughts?

benoitc:
Good catch! Thanks.

Modulo the style bikesheeding, the patch looks fine for me. If you can do the change that would help :)

urbaniak:
I've fixed the formatting.

Found also one more issue, which should be fixed in https://github.com/urbaniak/gunicorn/commit/4f810ee7bd126257c62729bdad080f5eb1bb13fb

Let me know if that's appropriate for other workers.

urbaniak:
Thought about that, done it that way because req_headers was written like that.

Should be good to go now ;)

gunicorn/instrument/statsd.py
```diff
@@ -93,9 +93,14 @@ def access(self, resp, req, environ, request_time):
         """
         Logger.access(self, resp, req, environ, request_time)
         duration_in_ms = request_time.seconds * 1000 + float(request_time.microseconds) / 10 ** 3
+        if type(resp.status) is int:
```
benoitc: style should also be changed there :)

benoitc:
@urbaniak thanks! Once last change and we are good :+1:

tilgovi:
@urbaniak would you mind squashing the commits, please? You can just force push and keep the same PR.

urbaniak:
No prob, squashed.

Hope to get that released soon.

tilgovi:
Thanks!

PR #1197, opened 2016-02-01
use EnvironmentError instead of socket.error (closes #939)

### LINKED ISSUE #939: BrokenPipeError being occasionally thrown with eventlet worker
In some cases on a very busy application, we get:

```
[2014-11-17 04:18:49 -0500] [13520] [ERROR] Error handling request
  File "/usr/local/lib/python3.4/site-packages/gunicorn/workers/async.py", line 108, in handle_request
    resp.write(item)
  File "/usr/local/lib/python3.4/site-packages/gunicorn/http/wsgi.py", line 344, in write
    util.write(self.sock, arg, self.chunked)
  File "/usr/local/lib/python3.4/site-packages/gunicorn/util.py", line 301, in write
    sock.sendall(data)
  File "/usr/local/lib/python3.4/site-packages/eventlet/greenio.py", line 375, in sendall
    tail = self.send(data, flags)
  File "/usr/local/lib/python3.4/site-packages/eventlet/greenio.py", line 358, in send
    total_sent += fd.send(data[total_sent:], flags)
  File "/usr/local/lib/python3.4/site-packages/eventlet/greenio.py", line 351, in send
    return fd.send(data, flags)
BrokenPipeError: [Errno 32] Broken pipe
```

The reason why is because the user closed out of the tab while it is loading (or otherwise halted navigation).  As such, it is probably not worthwhile to log this as a fatal error.

kaniini:
This pull request merges working code to fix the `BrokenPipeError` exceptions being thrown on Python 3 + eventlet.  It has been tested on a large application serving millions of requests hourly.

### DIFF
--- a/gunicorn/workers/async.py
+++ b/gunicorn/workers/async.py
@@ -58,7 +58,7 @@ class AsyncWorker(base.Worker):
                 exc_info = sys.exc_info()
                 # pass to next try-except level
                 six.reraise(exc_info[0], exc_info[1], exc_info[2])
-            except socket.error:
+            except EnvironmentError:
                 exc_info = sys.exc_info()
                 # pass to next try-except level
                 six.reraise(exc_info[0], exc_info[1], exc_info[2])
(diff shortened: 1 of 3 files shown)


tilgovi:
Thanks, @kaniini.

@berkerpeksag can you elaborate on your comment? https://github.com/benoitc/gunicorn/issues/939#issuecomment-72308337

Are you concerned that the exception classes are now more general, and might unintentionally catch errors? I think everything is already re-raised correctly if it's not EPIPE or ECONNRESET.

berkerpeksag:
I think EnvironmentError is the most Python 2 and Python 3 compatible way to fix this. We can replace it with OSError when we drop Python 2 support.

tilgovi:
:+1:

In that case, this is good to merge from my perspective.

berkerpeksag:
Thanks!

PR #1201, opened 2016-02-09
Clean up traceback printing in arbitar

bloodearnest:
Currently, on some errors, tracebacks are manually included in the log message. This patch instead uses the stdlibs logging traceback printing.  This allow better log customisation and control.

Additionally, because the call to logging.exception included the traceback in its log message, the traceback was printed twice the logs, as logging.exception always prints the traceback.

I've left the exceptions logged at debug/info as they are, but it seems to me they'd be better off as error level, since we sys.exit() after both anyway,

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -207,8 +207,8 @@ class Arbiter(object):
         except SystemExit:
             raise
         except Exception:
-            self.log.info("Unhandled exception in main loop:\n%s",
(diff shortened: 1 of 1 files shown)


benoitc:
:+1: can you rebase the commits in 1 ? will merge it right after.

bloodearnest:
Done, thanks

benoitc:
thanks!

PR #1209, opened 2016-02-16
Fix URL and default worker type in example_config

radzinzki:
This fixes the URL in the example_config.py

### DIFF
--- a/examples/example_config.py
+++ b/examples/example_config.py
@@ -32,8 +32,9 @@ backlog = 2048
 #       for your particular application's work load.
 #
 #   worker_class - The type of workers to use. The default
-#       async class should handle most 'normal' types of work
-#       loads. You'll want to read http://gunicorn/deployment.hml
+#       sync class should handle most 'normal' types of work
+#       loads. You'll want to read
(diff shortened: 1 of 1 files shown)


examples/example_config.py
```diff
@@ -33,7 +33,8 @@
 #
 #   worker_class - The type of workers to use. The default
 #       async class should handle most 'normal' types of work
-#       loads. You'll want to read http://gunicorn/deployment.hml
+#       loads. You'll want to read
+#       http://docs.gunicorn.org/en/latest/design.html
```
berkerpeksag: http://docs.gunicorn.org/en/latest/design.html#choosing-a-worker-type would be a better URL.

examples/example_config.py
```diff
@@ -33,7 +33,8 @@
 #
 #   worker_class - The type of workers to use. The default
 #       async class should handle most 'normal' types of work
```
berkerpeksag: Could also update update "async" to "sync"? Thanks!

radzinzki:
@berkerpeksag done

berkerpeksag:
Thanks!

PR #1213, opened 2016-03-08
update django doc url to 1.8 lts

stu-zumper:
When updating our company's gunicorn I noticed the doc link for django was out of date.  Looks like django removed all docs before 1.7.  Another option, which would avoid updating these docs every few years would be to point those urls to `dev`, eg: https://docs.djangoproject.com/en/dev/howto/deployment/wsgi/gunicorn/

Thanks for the project!


benoitc:
thanks!

PR #1217, opened 2016-03-13
[eventlet] handle QUIT in a new coroutine

tilgovi:
Just like in the gevent worker, a blocking call should only be executed
from within a coroutine. The AssertionError around the main loop of the
eventlet worker can be removed, since it was there to catch
the exception raised by the sleep.


benoitc:
:+1:

PR #1218, opened 2016-03-13
[gevent] remove obsolete exception clauses in run

tilgovi:
Since the changes from #1126, errors are not raised into the main
loop during a non-graceful shutdown. Therefore, these exception
clauses shouldn't be needed anymore.


tilgovi:
Similar to #1217

PR #1219, opened 2016-03-13
fix infinite recursion when destroying sockets

tilgovi:
By having a `getattr` implementation that proxies to the `sock`
attribute, there is a risk of infinite recursion when the socket
attribute is absent. After closing the socket and destroying it,
the recursion can be prevented by setting the attribute to `None`.


benoitc:
+1

Issue #922, opened 2014-10-25
Close listening socket immediately after stop signal received


tahajahangir:
When stopping gunicorn, server should wait for current requests to complete, but this may take several seconds (even 30 seconds, with the default timeout). Clients connecting in this interval, will wait forever, and will receive connection-reset error after server successfully stopped.

When using two or more gunicorn servers behind a reverse proxy (like nginx), this behavior leads to unreasonable raise in total upstream time (normally, real clients won't wait for 30 seconds, and will close tab, or hit refresh).

This also reduces impact of #901 (locked connections). Also related to #908

pikeas:
Was this ever fixed?

benoitc:
I din't tested it yet. thanks for the head-up it's on the todo for today
now!

mleventi:
+1 on getting this in

benoitc:
Re-reading this issue, I think we are missing some details, was the server closed using SIGQUIT or SIGTERM? It's totally expected to not close immediately  the socket on SIGTERM while SIGQUIT close it immediately. What did I miss?

tilgovi:
We're not closing the socket immediately which I think means that a
downstream proxy will keep trying to pass connections to gunicorn. I think
the request here is that during shutdown we close the listening socket
quickly.

tilgovi:
- [x] Close in arbiter shutdown (#1221)
- [ ] Close in workers at graceful termination
  - [x] sync (not applicable)
  - [x] gevent
  - [ ] eventlet (#1222)
  - [ ] thread
  - [ ] asyncio

tilgovi:
I believe that this can be marked complete as soon as the two PRs I have open for eventlet and the threaded worker are merged.

benoitc:
@tilgovi :+1:

tilgovi:
Closing as all my open PRs have been merged. Opened #1236 to track future work arising from this discussion and related PRs.

PR #1221, opened 2016-03-13
[arbiter] close sockets on shutdown

tilgovi:
Close all the listeners when the arbiter shuts down. By doing so,
workers can close the socket at the beginning of a graceful shut
down thereby informing the operating system that the socket can
be cleaned up. With this change, graceful exits with such workers
will refuse new connections while draining, allowing load balancers
to respond more quickly and avoiding leaving connections dangling
in the listen backlog, unaccepted.

Ref #922

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -334,6 +334,8 @@ class Arbiter(object):
         :attr graceful: boolean, If True (the default) workers will be
         killed gracefully  (ie. trying to wait for the current connection)
         """
+        for l in self.LISTENERS:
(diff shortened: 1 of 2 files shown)


tilgovi:
Added tests.

benoitc:
+1 . (review from the plane still on the ground)

tilgovi:
:earth_americas: :airplane: :shipit: :cocktail: :earth_africa:

benoitc:
indeed! thx for the patches! going off now for the next 9h

PR #1223, opened 2016-03-14
Demote broken pipe error to info

mdemmitt-lyft:


mdemmitt-lyft:
This was mistakenly submitted to the upstream rather than to a fork

PR #1226, opened 2016-03-16
Add PSF License Agreement to selectors.py

zx-zheng:
I want to use gunicorn in production environment. However, I found a license problem which prevent me from using gunicorn in production of my company.
It looks like that selectors.py is the copy of https://github.com/python/cpython/blob/master/Lib/selectors.py.
I think we should add PSF License Agreement to this file according to the license.
https://docs.python.org/2/license.html
https://docs.python.org/3/license.html

Cloud you add license agreement?

Thanks.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -122,3 +122,56 @@ util/unlink.py
 --------------
 
 backport frop python3 Lib/test/support.py
+
+
+gunicorn/selectors.py
+---------------------
+Copyright (c) 2001-2016 Python Software Foundation; All Rights Reserved
+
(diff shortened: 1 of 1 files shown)


benoitc:
Thanks for your change. But this licenseshould not been added there but in the NOTICE file and the notice header added here instead. Can you make that change?

zx-zheng:
Thanks for your comment.
I moved the agreement to the NOTICE.
Cloud you check it again?

berkerpeksag:
Thanks! I thought I already did this :)

tilgovi:
Attention to detail is so appreciated! Thank you!

PR #1228, opened 2016-03-18
Fix delayed socket close caused by circular ref from assigning traceback to local var

paulnivin:
As part of pulling in 2e231d20d51677949067247d8133734943ca0a72 and a3f6df5b7e37eba4d45c841b0c0c5b3fc983258b, EPIPEs result in the `EnvironmentError` exception handler in `handle()` being run. In many environments, EPIPEs can be extremely common. The exception handler stores the traceback into a local variable, creating a circular reference, and prevents the system socket (holding a fd open along with other socket resources) from being closed until the next Python full garbage collection cycle. The behavior of `sys.exec_info()` is documented at https://docs.python.org/2/library/sys.html#sys.exc_info with a large red warning statement. This new exception path for EPIPEs can create significant additional resource usage for gunicorn installations.

Removing the assignment into `exc_info` causes the socket fd to be freed quickly (handled by ref counting), as was the case before the two prior patches were merged.

This commit also fixes the `ssl.SSLError` path.

### DIFF
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -134,11 +134,18 @@ class Worker(object):
 
             self.log.exception(e)
 
-            exc_type, exc_val, exc_tb = sys.exc_info()
-            self.reloader.add_extra_file(exc_val.filename)
+            # fix from PR #1228
+            # storing the traceback into exc_tb will create a circular reference.
+            # per https://docs.python.org/2/library/sys.html#sys.exc_info warning,
+            # delete the traceback after use.
+            try:
+                exc_type, exc_val, exc_tb = sys.exc_info()
+                self.reloader.add_extra_file(exc_val.filename)
 
-            tb_string = traceback.format_exc(exc_tb)
-            self.wsgi = util.make_fail_app(tb_string)
+                tb_string = traceback.format_exc(exc_tb)
+                self.wsgi = util.make_fail_app(tb_string)
+            finally:
+                del exc_tb
 
     def init_signals(self):
         # reset signaling
--- a/gunicorn/_compat.py
+++ b/gunicorn/_compat.py
@@ -107,11 +107,8 @@ def _wrap_error(exc, mapping, key):
     new_err = new_err_cls(*exc.args)
 
     # raise a new exception with the original traceback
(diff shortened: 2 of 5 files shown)


tilgovi:
:100: 

@paulnivin there are a few other places where this happens. Want to :fire: them all?

benoitc:
+1

paulnivin:
I fix up the other places as well and update the branch.

tilgovi:
@paulnivin did you push?

paulnivin:
Not yet. Will shortly. I'm going through the codebase and fixing up all the uses of `sys.exec_info()`. The async EnvironmentError path was the only one that caused a dramatic change in gunicorn performance/resource utilization for us.

paulnivin:
Updated the branch to clean up all the calls to `sys.exc_info()` to conform to recommended best practices. We're not running this code in production, but it passes the gunicorn tests. :expressionless:

gunicorn/workers/base.py
```diff
@@ -134,11 +134,14 @@ def load_wsgi(self):
...
-            tb_string = traceback.format_exc(exc_tb)
-            self.wsgi = util.make_fail_app(tb_string)
+            try:
+                exc_type, exc_val, exc_tb = sys.exc_info()
+                self.reloader.add_extra_file(exc_val.filename)
+
+                tb_string = traceback.format_exc(exc_tb)
+                self.wsgi = util.make_fail_app(tb_string)
+            finally:
+                del exc_tb
```
berkerpeksag: We could use a comment to explain what `del exc_tb` does here (and perhaps a reference to this PR).

Could you also squash the commits into one?

Thanks!

tilgovi:
Also, sorry @paulnivin, I misread your comment and thought it said you _already_ had fixed up the other places when I asked if you had pushed. Wasn't trying to be pushy. :stuck_out_tongue_winking_eye:

gunicorn/workers/base.py
paulnivin: Added a comment and squashed the commits.

tilgovi:
Excellent. Thanks, again.

PR #1229, opened 2016-03-18
check if the header contains control characters

### LINKED ISSUE #1227: Potential HTTP Response Splitting Vulnerability
This [header injection vulnerability](http://www.zeroscience.mk/en/vulnerabilities/ZSL-2015-5267.php) was recently reported to and fixed in both [waitress (WSGI server)](https://github.com/Pylons/waitress/issues/117) and [WebOb](https://github.com/Pylons/webob/pull/229) (and waitress has [an outstanding issues](https://github.com/Pylons/waitress/issues/122) to discuss further possible vulnerabilities). I also [made changes to gevent](https://github.com/gevent/gevent/issues/775) for these issues.

gunicorn seems to be partly (but not entirely) vulnerable to these issues as well, across at least the `sync` and `gevent` workers.

Here are three apps, each producing different kinds of bad output. The `badvalue` case (invalid values in the header values) is probably the most likely injection vulnerability, with the next being the `badname` case (invalid value in the header names) and the least likely being the `badstatus` case (invalid value in the status line):

```python
def badvalue(environ, start_response):
	start_response("200 BadValue",
				   [("Test", "Hi\r\nContent-length: 0\r\nConnection: keep-alive\r\nTransfer-Encoding: identity\r\n\r\n\r\n")])
	return [b"This should be the body!"]

def badname(environ, start_response):
	start_response("200 BadName",
				   [("\r\n\r\n\r\nContent-length: 0\r\nConnection: close\r\n\r\n", "Nothing")])
	return [b"This should be the body!"]

def badstatus(environ, start_response):
    start_response("200 Evil\r\nContent-Length: 0\r\nConnection: close\r\n\r\n", [])
    return [b"This should be the body!"]
```

Here's what happens hitting `badvalue`. We get a bunch of duplicate headers that might confuse clients:
```
$ http 127.0.0.1:8000
HTTP/1.1 200 BadValue
Connection: close
Connection: keep-alive
Content-length: 0
Date: Thu, 17 Mar 2016 13:33:39 GMT
Server: gunicorn/19.4.5
Test: Hi
Transfer-Encoding: chunked
Transfer-Encoding: identity

This should be the body!
```

Here's what happens hitting `badname`. The worst I was able to do was produce malformed HTTP; on other servers I was able to cause clients to hang, but I haven't reproduced that with gunicorn (yet):
```
$ http 127.0.0.1:8000
HTTP/1.1 200 BadName
Connection: close
Connection: close: Nothing
Content-length: 0
Date: Thu, 17 Mar 2016 13:36:39 GMT
Server: gunicorn/19.4.5
Transfer-Encoding: chunked

This should be the body!
```

Finally, here's `badstatus`. Here, we completely override the rest of the response:
```
$ http 127.0.0.1:8000
HTTP/1.1 200 Evil
Connection: close
Content-Length: 0


```

Should gunicorn check for and raise exceptions for these type of malformed values in `start_response`?

benoitc:
fix #1227

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -10,6 +10,7 @@ import re
 import sys
 
 from gunicorn._compat import unquote_to_wsgi_str
+from gunicorn.http.errors import InvalidHeader
 from gunicorn.six import string_types, binary_type, reraise
 from gunicorn import SERVER_SOFTWARE
 import gunicorn.util as util
@@ -28,6 +29,7 @@ except ImportError:
 BLKSIZE = 0x3FFFFFFF
 
 NORMALIZE_SPACE = re.compile(r'(?:\r\n)?[ \t]+')
+HEADER_VALUE_RE = re.compile(r"[\x07\x1B\f\n\r\t\v]")
 
 log = logging.getLogger(__name__)
 
@@ -264,6 +266,10 @@ class Response(object):
         for name, value in headers:
             if not isinstance(name, string_types):
                 raise TypeError('%r is not a string' % name)
+
+            if HEADER_VALUE_RE.search(value):
+                raise InvalidHeader('%r' % value)
+
             value = str(value).strip()
             lname = name.lower().strip()
             if lname == "content-length":
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -10,6 +10,7 @@ import re
 import sys
 
 from gunicorn._compat import unquote_to_wsgi_str
+from gunicorn.http.message import HEADER_RE
 from gunicorn.http.errors import InvalidHeader
 from gunicorn.six import string_types, binary_type, reraise
 from gunicorn import SERVER_SOFTWARE
@@ -266,6 +267,9 @@ class Response(object):
         for name, value in headers:
             if not isinstance(name, string_types):
                 raise TypeError('%r is not a string' % name)
+
+            if HEADER_RE.search(name):
+                raise InvalidHeaderName('%r' % name)
 
             if HEADER_VALUE_RE.search(value):
                 raise InvalidHeader('%r' % value)
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -29,7 +29,7 @@ except ImportError:
 BLKSIZE = 0x3FFFFFFF
 
 NORMALIZE_SPACE = re.compile(r'(?:\r\n)?[ \t]+')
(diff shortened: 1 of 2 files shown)


benoitc:
please review.

gunicorn/http/wsgi.py
```diff
@@ -28,6 +29,7 @@
 BLKSIZE = 0x3FFFFFFF
 
 NORMALIZE_SPACE = re.compile(r'(?:\r\n)?[ \t]+')
+HEADER_VALUE_RE = re.compile(r"[\x07\x1B\f\n\r\t\v]")
```
tilgovi: Could this be simply `[\x00-\x1F]`?

tilgovi: Maybe also 127 (DEL)?

benoitc: hrm even `[\x00-\x1F\x7F]`?

benoitc:
we should probaly add a test there too ...

gunicorn/http/wsgi.py
benoitc: @tilgovi yeah :)

benoitc:
@tilgovi just added a test. let me know if it's OK for you.

tilgovi:
Code looks good but I didn't take the time yet to read about the issue. @jamadden does this look good to you?

tests/test_http.py
```diff
@@ -94,3 +95,19 @@ def test_http_header_encoding():
...
+    mocked_socket = mock.MagicMock()
+    mocked_socket.sendall = mock.MagicMock()
+
+    mocked_request = mock.MagicMock()
+    response = Response(mocked_request, mocked_socket, None)
+
+    # set umlaut header
+    try:
+        response.start_response("200 OK", [('foo', 'essai\r\n')])
+    except Exception as e:
```
berkerpeksag: Or:

```py
with pytest.raises(InvalidHeader):
    response.start_response("200 OK", [('foo', 'essai\r\n')])
```

benoitc:
@jamadden I think we are good there but let me know if we missed something. If not i will rebase and make a release during the we.

jamadden:
@benoitc Thanks, that does seem to solve the header value case, which is *probably* the most likely to be exploited. 

In descending order of priority and likelihood to be exploited, I would still suggest applying the same check to the header names and the status line. A \r or \n in any of those places causes the same or similar problems...depending on whether or not chunked transfer encoding is used in the case of header names.

benoitc:
rebased in master. 

@jamadden header names are also checked.

PR #1239, opened 2016-04-01
Use EPOLLEXCLUSIVE to avoid thundering herd problem

methane:
linux 4.5 introduced EPOLLEXCLUSIVE which can be used to avoid thundering herd problem.
http://man7.org/linux/man-pages/man2/epoll_ctl.2.html

Before linux 4.5, this flag is just ignored.


methane:
I found performance regression with this patch.
I'll investigate it and discuss on bugs.python.org.

methane:
ref https://github.com/python/asyncio/pull/330

PR #1248, opened 2016-04-29
do post_worker_init after load_wsgi

gescheit:
After 81f65c2 it became impossible to get access to application in post_worker_init, because application not loaded yet. This patch return previous behaviour.


benoitc:
Thanks!

PR #1251, opened 2016-05-03
Add access_log_format to config example

pgjones:
The example given is the current default log format.

I found myself looking through the code to figure out how to set this, so hopefully this will save other peoples time.


berkerpeksag:
Thanks!

Issue #1253, opened 2016-05-04
Gunicorn should not swallow exceptions raised in worker exit handler


jgehrcke:
The last steps before a worker process exits:

```python
        finally:
            self.log.info("Worker exiting (pid: %s)", worker_pid)
            try:
                worker.tmp.close()
                self.cfg.worker_exit(self, worker)
            except:
                pass
```
(currently: https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L538)

Instead of just `pass`ing on an exception, we should still emit the details. Otherwise people may just not realize that they have an error in their exit handler. I suggest adding `import traceback`  and then do `traceback.print_exc(file=sys.stdout)` instead of the `pass`. Swallowing exceptions is almost never a great approach.

Feedback welcome. If that seems useful, I am happy to submit a PR.

tilgovi:
Seems fine to me to log a traceback.

PR #1254, opened 2016-05-04
Do not swallow exception raised during worker exit 

jgehrcke:
This tackles #1253.

I have also added a commit proposing a documentation change: when I read the current doc for the `worker_exit()` hook, I was not sure if it's actually executed *in* the worker process (right before it terminates), or in the master process (right after it has terminated).

Please have a look, thanks!

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -11,6 +11,7 @@ import select
 import signal
 import sys
 import time
+import traceback
 
 from gunicorn.errors import HaltServer, AppImportError
 from gunicorn.pidfile import Pidfile
@@ -535,7 +536,8 @@ class Arbiter(object):
                 worker.tmp.close()
                 self.cfg.worker_exit(self, worker)
             except:
-                pass
+                self.log.warning("Exception during worker exit:\n%s",
+                                  traceback.format_exc())
 
     def spawn_workers(self):
         """\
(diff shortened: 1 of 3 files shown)


gunicorn/arbiter.py
```diff
@@ -535,7 +536,8 @@ def spawn_worker(self):
                 worker.tmp.close()
                 self.cfg.worker_exit(self, worker)
             except:
-                pass
+                self.log.error("Exception during worker exit:\n%s",
```
benoitc: can't `log.exception` be used for it instead? Or maybe `debug`?

tilgovi: I had a similar thought. I agree about `log.exception`. I think an error level is appropriate, though, because the application did error. Or at least warning, but maybe just error. I think debug is too low when there was an actual fault.

tilgovi: I thought warning only because it doesn't affect gunicorn functions, but I don't think error would be bad.

benoitc: warning should be indeed better since this error is harmless. @jgehrcke can you make the change?

jgehrcke: Thanks for feedback!

> since this error is harmless

Depends on what people do in their handler :-) Could be pretty critical to them.

However, I'm also good with level `WARNING`: in practice, people should not swallow that level, and if they do -- too bad.

Since `Logger.exception` [logs on ERROR level](https://docs.python.org/2/library/logging.html#logging.Logger.exception), this is not an option. So, I'll simply switch to `log.warning(...)`.

jgehrcke:
Updated the branch, thanks for feedback @benoitc @tilgovi!

benoitc:
Thanks!

PR #1257, opened 2016-05-10
unblock the wait loop under python 3.5

### LINKED ISSUE #1256: Gunicorn doesn't shutdown immediately on receiving SIGTERM
I am seeing a weird behavior where Gunicorn does not shutdown immediately when sent a SIGTERM. I see this with Python 3.5 on Ubuntu 16.04. The following environments (that I have access to) *do not* exhibit this issue:
- Python 2.7 on Ubuntu 16.04
- Python 2.7 on Ubuntu 14.04
- Python 3.4 on Ubuntu 14.04

Refer to https://gist.github.com/suriya/1d9a11ea9ad33a5b1df0b2ba7ed0cd0e on how to reproduce this issue.

After debugging a while, I found that the Gunicorn arbiter properly signals the child worker. The child workers handle the signal properly as well (on all versions). However, something prevents the loop in `run_for_one` function https://github.com/benoitc/gunicorn/blob/9d158bec94c91c62cd5cc28a27268a5b8804d795/gunicorn/workers/sync.py#L58 from exiting right away with Python 3.5.

Could Python 3.5's change in the behavior of `select.select()` https://docs.python.org/3/library/select.html#select.select explain this issue.

benoitc:
in python 3.5 the select is blocking when waiting for it which prevent quick
exit on SIGTERM.

The problem is described:
https://www.python.org/dev/peps/pep-0475/#backward-compatibility

This change fix it by listening for signal event on the worker pipe. Once an
event is triggered it will forcefully wake up the select and return.

fix #1256

### DIFF
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -113,6 +113,8 @@ class Worker(object):
         [util.close_on_exec(s) for s in self.sockets]
         util.close_on_exec(self.tmp.fileno())
 
+        self.wait_fds = self.sockets + [self.PIPE[0]]
+
         self.log.close_on_exec()
 
         self.init_signals()
@@ -163,6 +165,9 @@ class Worker(object):
         if hasattr(signal, 'siginterrupt'):  # python >= 2.6
             signal.siginterrupt(signal.SIGTERM, False)
             signal.siginterrupt(signal.SIGUSR1, False)
+
+        if hasattr(signal, 'set_wakeup_fd'):
+            signal.set_wakeup_fd(self.PIPE[1])
 
     def handle_usr1(self, sig, frame):
         self.log.reopen_files()
--- a/gunicorn/workers/sync.py
+++ b/gunicorn/workers/sync.py
@@ -32,7 +32,7 @@ class SyncWorker(base.Worker):
     def wait(self, timeout):
         try:
             self.notify()
-            ret = select.select(self.sockets, [], self.PIPE, timeout)
+            ret = select.select(self.wait_fds, [], [], timeout)
             if ret[0]:
                 return ret[0]
 
@@ -93,6 +93,9 @@ class SyncWorker(base.Worker):
 
             if ready is not None:
                 for listener in ready:
+                    if listener == self.PIPE[0]:
+                        continue
+
                     try:
                         self.accept(listener)
                     except EnvironmentError as e:


aolefira:
In python 3.6 this bug appeared again

benoitc:
@aolefira how do you produce it?

aolefira:
I run script ./stop.sh with next content:
kill `cat gunicorn.pid`
and waiting a lot of seconds until gunicorn processes will be killed.
Software:
gunicorn 19.7.1
python 3.6.0
ubuntu 16.04.1 LTS

benoitc:
how can I rrepoduce it? Do you have any log?

aolefira:
There are no records in logs while shutting down.
Command for start:
`
gunicorn --paste pyramid.ini --daemon
`
[server:main] part of pyramid.ini:
`
[server:main]
use = egg:gunicorn#main
host = 0.0.0.0
port = 6543
threads = 60
worker_class = gthread
errorlog = logs/error.log
loglevel = warning
capture_output = True
pidfile = gunicorn.pid
statsd_host = localhost:8125
`
Any pyramid project should be used.
To reproduce bug just enter:
`
kill `cat gunicorn.pid`
`

tilgovi:
I cannot reproduce this.

MacOS X 10.13
Python 3.6.2
Gunicorn 19.7.1
Pyramid 1.9.1

Do you have a long connection open when you do this? What happens if you use `kill -9`?

aolefira:
There are no long connections. 
If `kill -9` used, first process terminate, but second process remains hanging.

aolefira:
Yes, on my macbook with MacOS X 10.11.6 this issue not reproduced too.
This issue reproduced on server with Ubuntu 16.04.1 LTS.

aolefira:
> Do you have a long connection open when you do this?

This bug reproduce (on Ubuntu 16.04.1) even if there are no connections.

aolefira:
Problem is that pyramid runs thread at startup. In mac os x that thread die immediately on app shutdown but in ubuntu it's waiting for some time

aolefira:
setDaemon(True) for threads resolve this problem

tilgovi:
What thread does Pyramid run at startup? I'm not aware of it doing that.

tilgovi:
As far as I know, the only thread Pyramid starts are started by `pserve` command, but you said you are using the `gunicorn --paste` command.

aolefira:
> What thread does Pyramid run at startup? I'm not aware of it doing that.

It's project specific threads. Two custom threads. First - scheduler and second - mailer.

aolefira:
> As far as I know, the only thread Pyramid starts are started by pserve command, but you said you are using the gunicorn --paste command.

pserve is good for development. but for production its better to use something more advenced, like gunicorn

[https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/project.html#using-an-alternate-wsgi-server](https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/project.html#using-an-alternate-wsgi-server)

[https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/deployment/gunicorn.html](https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/deployment/gunicorn.html)

Dringho:
Also experiencing this bug on Ubuntu 16.04 / Python 3.5.2 / Gunicorn 19.7.1

PR #1261, opened 2016-05-13
don't kill ourself on reload

### LINKED ISSUE #1129: Gunicorn --reload option reloads indefinitely
Running any application, such as the following application (from gunicorn.org) with the `--reload` option and then making a trivial change leads to a never-ending reload cycle on my system.

```
      data = "Hello, World!\n"
      start_response("200 OK", [
      ])
      return iter([data])
```

```
$ gunicorn --version
gunicorn (version 19.3.0)
$ python --version
Python 3.4.3
$ gunicorn --reload app:app -b localhost:8000
[2015-10-16 13:01:58 +0200] [42184] [INFO] Starting gunicorn 19.3.0
[2015-10-16 13:01:58 +0200] [42184] [INFO] Listening at: http://127.0.0.1:8000 (42184)
[2015-10-16 13:01:58 +0200] [42184] [INFO] Using worker: sync
[2015-10-16 13:01:58 +0200] [42187] [INFO] Booting worker with pid: 42187
[2015-10-16 13:02:05 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:06 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:07 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:08 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:09 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:10 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:11 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:12 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
[2015-10-16 13:02:13 +0200] [42187] [INFO] Worker reloading: /Users/lyschoening/app.py modified
^C[2015-10-16 13:02:13 +0200] [42184] [INFO] Handling signal: int
[2015-10-16 13:02:13 +0200] [42187] [INFO] Worker exiting (pid: 42187)
[2015-10-16 13:02:13 +0200] [42184] [INFO] Shutting down: Master
```

Please let me know what other debug information I can provide.

benoitc:
Killing ourself when using the `--reload` option trigger an infinite loop under some monitoring services like the one in pycharm and don't reload the file.

Instead set self.alive as False which will trigger later the worker exit. Note that if we want to force the exit we could also use sys.exit(0) .

fix #1129

### DIFF
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -90,6 +90,10 @@ class Worker(object):
             def changed(fname):
                 self.log.info("Worker reloading: %s modified", fname)
                 self.alive = False
+                self.cfg.worker_int(self)
+                time.sleep(0.1)
+                sys.exit(0)
+
             self.reloader = Reloader(callback=changed)
             self.reloader.start()
 
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -89,7 +89,7 @@ class Worker(object):
         if self.cfg.reload:
             def changed(fname):
                 self.log.info("Worker reloading: %s modified", fname)
-                os.kill(self.pid, signal.SIGQUIT)
+                self.alive = False
             self.reloader = Reloader(callback=changed)
             self.reloader.start()
 


benoitc:
@tilgovi let me know what you think about this change :)

tilgovi:
Very weird... I don't understand why the kill causes problems. Do you?

I'd prefer if we can exit immediately, otherwise apps that use long polling or websockets will unreliably reload in development.

tilgovi:
Maybe pycharm monitors the process tree? I don't understand, though. This signal is sent to the worker. Whatever is supervising the arbiter shouldn't receive a SIGCHLD or detect the exit status. Very weird. I would like to understand the cause better.

benoitc:
@tilgovi I'm not sure. It may be that the signal is captured by pycharm in their debugger. The fix above works though :) Any idea on how to debug further?

benoitc:
@tilgovi code seems to work well there. Is this OK for you to merge it?

tilgovi:
LGTM

Issue #605, opened 2013-09-04
Body.read(), Body.readline() and Body.readlines() are broken for Transfer-Encoding: chunked


fafhrd91:
If request uses 'chunked' transfer encoding and chunk size is smaller than 1024 bytes
then Body's readxxx methods blocks forever.

genxstylez:
@benoitc 

I am having issue with nginx + gunicorn + django.

When I am using chunked transfer encoding, the file I get is a partial one, not a complete jpg.

Not really sure where my problem is, maybe you can help me out here.

client:
`curl -v -X POST  --header "Transfer-Encoding: Chunked" -d @a.jpg "http://127.0.0.1/upload/"`

nginx:
```
location / {
            proxy_pass http://localhost:8000;
            proxy_redirect off;
            proxy_set_header Host $host;
            proxy_set_header X-Real-IP $remote_addr;
            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
            client_max_body_size 20m;
            client_body_buffer_size 128k;
            proxy_connect_timeout 90;
            proxy_send_timeout 90;
            proxy_buffering off;
            proxy_temp_file_write_size 64k;
        }
```

django:
```python
@csrf_exempt
def upload(request):
    print('body:', request.body)
    # print('meta:', request.META)
    mode = 'wb'
    if os.path.exists('a.jpg'):
        mode = 'ab'
    with open('a.jpg', mode) as f:
        f.write(request.body)

    return HttpResponse('yes')
```

gunicorn:
```python
import multiprocessing

workers = (2 * multiprocessing.cpu_count()) + 1
worker_class = 'gevent'
loglevel = 'debug'
accesslog = '-'
errorlog = '-'
forwarded_allow_ips = '*'
secure_scheme_headers = {
    # 'X-FORWARDED-PROTOCOL': 'ssl',
    'X-FORWARDED-PROTO': 'https',
    # 'X-FORWARDED-SSL': 'on',
}
preload_app = False
limit_request_line = int(4094 / 4)
limit_request_fields = int(100 / 2)
keepalive = 2 * 2
```

benoitc:
hrm what's your version of django?

genxstylez:
1.9.6

I just tried with a 114 bytes text file, and the upload works, but the `\r\n` is truncated

e.g.
original 
```
first line here
second line here
```

uploaded
```
first line here second line here
```

benoitc:
@genxstylez what's your version of python?

[...]

PR #1265, opened 2016-05-14
set CONTENT_LENGTH to max integer when receiving chunked encoding req…

### LINKED ISSUE #1264: chunked encoding bug with latest version of django.
Following the discussing on #605 :
https://github.com/benoitc/gunicorn/issues/605#issuecomment-219228942

Django doesn't handle anymore chunked encoded requests. Not sure how to fix it yet.

benoitc:
set CONTENT_LENGTH to max integer when receiving chunked encoding requests

After looking at the code of Django or Werkzeug it seems that some WSGI frameworks considers that the body is empty when the `CONTENT_LENGTH` environ key is not present:

- django: https://github.com/django/django/blob/master/django/core/handlers/wsgi.py#L109-L112
- https://github.com/pallets/werkzeug/blob/master/werkzeug/wsgi.py#L205-L210

So waiting for a better  spec set the size to max since we know that Gunicorn is able to handle chunked encoding.

fix #1264

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -12,7 +12,7 @@ import sys
 from gunicorn._compat import unquote_to_wsgi_str
 from gunicorn.http.message import HEADER_RE
 from gunicorn.http.errors import InvalidHeader, InvalidHeaderName
-from gunicorn.six import string_types, binary_type, reraise
+from gunicorn.six import string_types, binary_type, reraise, MAXSIZE
 from gunicorn import SERVER_SOFTWARE
 import gunicorn.util as util
 
@@ -82,6 +82,7 @@ def base_environ(cfg):
         "wsgi.multiprocess": (cfg.workers > 1),
         "wsgi.run_once": False,
         "wsgi.file_wrapper": FileWrapper,
+        "CONTENT_LENGTH": MAXSIZE,
         "SERVER_SOFTWARE": SERVER_SOFTWARE,
     }
 


benoitc:
also fix #605. 

@tilgovi @berkerpeksag what do you think about it?

tilgovi:
I'm only worried that some framework or library might think there is an error if there is not that many bytes read.

I'll think more and read some code and write back later today.

benoitc:
@tilgovi oh good catch, I will as well read other code around. Also I asked on twitter:
https://twitter.com/benoitc/status/731529585217679360

benoitc:
@tilgovi also it seems that werkzeug is able to just use the wsgi stream if safe_fallback is False. Maybe we should rather have some configuration setting or a switch to force a "safe" behaviour depending on the framework?

GrahamDumpleton:
You can't safely assume ``MAXSIZE``. You will inevitably find some code which assumes that if it didn't read to the specified length that it is an error.

FWIW, I pushed a number of times to get the WSGI specification updated to handle chunked request encoding by way of pushing that ``CONTENT_LENGTH`` was advisory, with expectation that WSGI applications would just read to end of input, designated by being returned an empty string. Incomplete input would always be flagged by the WSGI server raising an IOError exception. Of course with the dysfunctional way the web-sig worked, never managed to get it changed.

It is for that reason that you will find unofficial support for it in mod_wsgi and Werkzeug. In the case of mod_wsgi you need to set the directive ``WSGIChunkedRequest`` to enable the behaviour which would allow it to work contrary to the WSGI specification, but how you would want it to work to handle chunked request and mutating input filters.

I have totally given up on having anything to do with the WSGI specification work now, so good luck getting anything to happen. I was the only one really championing this and no one else really cared, with some even trying to block it without trying to understand the reasons.

If you are going to support this, I would recommended adding a ``--chunked-request`` option to enable the behaviour of reading all input from a chunked request. Do not change ``CONTENT_LENGTH`` though. Best is not to set it at all if you currently fallback to setting it as 0. Setting to ``MAXSIZE`` or ``-1`` will cause problems at some point.

jamadden:
FWIW, gevent's WSGI implementation supports chunked requests *by default*, returning an empty string on EOF and raising IOErrors otherwise. If the request is chunked, the WSGI environment simply has no CONTENT_LENGTH key present. There's currently no way to disable this behavior.

benoitc:
ok, so forcing the `CONTENT_LENGHT` is a bad idea. I hear you :) 

Right now I'm clueless on the solutions we can offer for the users though. Maybe we can ship a middleware for django and wsgi users that fix the issue for them? (if it doesn't exists that it).

GrahamDumpleton:
How are you going to do it as a middleware? Support for chunked request encoding can only be implemented at the WSGI server layer. A WSGI server is not supposed to expose hop by hop transport stuff like chunking to higher layers.

benoitc:
@GrahamDumpleton well gunicorn already handle it. The issue is more that the application don't :) 

So the point of the middleware would be to trick the application when needed ...

benoitc:
i mean we already handle chunked streams, body without content length, etc.. In fact the application behind gunicorn should only have to read from the wsg.input until it get an empty string (EOF) .

benoitc:
closing the PR. We need to find a better way to provide support for legacy apps sticking on the WSGI spec for it.

PR #1266, opened 2016-05-15
lockfile improvements

benoitc:
- rename LockFile.lock  to acquire
- rename LockFile.unlock to release
- move the lockfile management in sepate functions inside the arbiter
- remove the "closed" argument from the socket.close method and add a new "destroy" function that will be called whent  the socket can be unlinked (cal release)
- fix tests

related to #1263 discussion

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -138,16 +138,7 @@ class Arbiter(object):
             self.LISTENERS, need_lock = create_sockets(self.cfg, self.log)
 
         if need_lock:
-            if not self.LOCK_FILE:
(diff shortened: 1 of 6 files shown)


benoitc:
@tilgovi does `release` sounds better? How do you find this version. We likely need to release a new version of gunicorn tomorrow since the unix sockets are broken on non linux systems. Let me know :)

tilgovi:
I won't have a chance to write until later, six hours or so.

If you need to release this, go ahead.

benoitc:
@tilgovi but about the wording? Is release better? If you have anything that come to your mind let me know :)

PR #1270, opened 2016-05-18
remove file locking

### LINKED ISSUE #1267: USR2 improvements & fixes
This ticket track a breaking change that should land asap in Gunicorn to fix #1263 and other issues that we found during the investigations. In the mean time I propose to merge #1266 to fix the current version and release Gunicorn 19.6 today.

I summarised here the current implementations and the issues. You will also find a proposal [based on the one initially done](https://github.com/benoitc/gunicorn/pull/1263#issuecomment-219258874) by @tilgovi to solve it.

Thoughts? If it's OK I can take care about the implementation. cc @tilgovi  @berkerpeksag

### goal of USR2:

Allows upgrade of Gunicorn and the application code:
http://docs.gunicorn.org/en/stable/signals.html#upgrading-to-a-new-binary-on-the-fly

### current implementation ([code](https://github.com/benoitc/gunicorn/blob/master/Gunicorn/arbiter.py#L377-L400))

We create a new process inheriting from file descriptors using a combinaison of `os.fork` followed by `os.execvpe` to replace the full process.

If a pidfile exists is it renamed to `<pidfile>.oldbin` before we fork. 

Using a lock we prevent the deletion of the unix socket.

### current issues

a. locking the unix socket file wasn't portable and was replaced in 8a6748ee653f290de37e695cd36e7c41d8ba8afd by a more generic way. We now create a lock file used to check if the arbiter actually own it and can do some actions. This solution introduce a new file on the filesystem.
b. Unix files managed by a supervisor like systemd can be deleted https://github.com/benoitc/gunicorn/pull/1263#issuecomment-219297437 . Which is not wanted
c. If the new process spawned during USR2 is killed for any reason, the pidfile of the old/current arbiter stays with the name `pidfile.oldbin`

## Solution

@tilgovi proposed to track the newly created arbiter https://github.com/benoitc/gunicorn/pull/1263#issuecomment-219258874 and to make sure that the new arbiter if a child don't allow another arbiter to be created. 

I will add to that that the following:

- old arbiter need to be aware that the new arbiter has been killed so it can takes appropriates actions like renaming the pidfile back to `pidfile` and clean any tracking metadata. 
- we should keep a list of the file descriptors under our control, ie. any sockets that have been actually created by us.
- the new arbiter check if its parent is alive. If the parent is alive, it can't unlink any sockets under the Gunicorn control. If its parent is dead then it will be able to unlink them. 
- until the parent is alive, the pidfile related to it i not renamed, and the process name stays the same. On the other hand the new spawned arbiter takes the name "Master 2" (or anything better) and its pidfile is name `pidfile.2` . The process name will be renamed to `Master` and the pidfile `pidfile` when the old arbiter die.

So the new workflow on USR2 will be

1. Before executing `os.execvpe` track the new master pid [here](https://github.com/benoitc/gunicorn/blob/master/Gunicorn/arbiter.py#L385-L387).
2. we don't rename the pidfile and process name in the current arbiter
3. in the new spawned arbiter, we record the parent pid. In the main loop we check if its parents died (like in workers).
4. the parent check if the arbiter died and mark it . Which is already done [there](https://github.com/benoitc/Gunicorn/blob/master/Gunicorn/arbiter.py#L486-L487).
5. if the parent is killed, the new arbiter renamed its pidfile and process name and mark the old parent as died (like we do for the `reexec_pid`).

Once on (5) the new arbiter can unlink any sockets. Before it, it can only close them.

To fix (b) we keep a trace of all created sockets. We create a new environment variable named `GUNICORN_FDS_CTL` that only contains the file descriptor created by Gunicorn at boot. If the fd is not in this list it won't be unlinked.

benoitc:
This changes improve the binary upgrade behaviour using USR2:

- only one binary upgrade can happen at a time: the old arbiter needs to be
  killed to promote the new arbiter.
- if a new arbiter is already spawned, until one is killed USR2 has no action
- if a new arbiter has been spawned, the unix socket won't be unlinked
- until the old arbiter have been killed the newly created pidfile has the name
  <pidfile>.2 and the name Master.2 .

Note: there is no dialog between both arbiters to handle this features.
Instead they will supervise each others until one is killed. So isolation is
still guaranted.

fix #1267


gunicorn/arbiter.py
```diff
@@ -302,6 +299,23 @@ def handle_winch(self):
         else:
             self.log.debug("SIGWINCH ignored. Not daemonized")
 
+    def if_promoted(self):
+        if self.master_pid == 0:
+            return
+
+        if self.master_pid != os.getppid():
+            self.log.info("Master have been promoted.")
```
tilgovi: Grammar: "Master has been promoted."

tests/test_sock.py
```diff
@@ -9,19 +9,10 @@
 @mock.patch('os.getpid')
 @mock.patch('os.unlink')
 @mock.patch('socket.fromfd')
-def test_unix_socket_close_delete_if_exlock(fromfd, unlink, getpid):
+def test_unix_socket_close_keep(fromfd, unlink, getpid):
```
tilgovi: Seems like this should read `test_unix_socket_close_unlink` since this test case checks that unlink _is_ called.

gunicorn/sock.py
```diff
@@ -20,6 +20,7 @@
 class BaseSocket(object):
 
     def __init__(self, address, conf, log, fd=None):
+        self.parent = os.getpid()
```
tilgovi: `self.parent` should just be removed. It's not doing anything anymore. It is always set to `os.getpid()` so it will always be equal to `os.getpid()` at close. This is because after USR2 exec these objects are still recreated in the new process.

If the intention is to prevent the destruction of systemd activated sockets then we need a separate flag here that we pass to the constructor only when called in the `LISTEN_FDS` loop, or we need to check for the environment variable here (a parameter is easier to unit test).

The arbiter is now tracking whether it's the "master" arbiter and intelligently calls `close` (or not), so the only information we need here is whether Gunicorn (any process) created the socket or whether the init system did.

It's not actually important which PID created the socket (the arbiter tracks whether to call `close()`)

gunicorn/arbiter.py
```diff
@@ -302,6 +299,23 @@ def handle_winch(self):
         else:
             self.log.debug("SIGWINCH ignored. Not daemonized")
 
+    def if_promoted(self):
```
tilgovi: I would give this a new name. Maybe `manage_arbiters` for symmetry with `manage_workers`. Right now it looks to me like a function that returns a boolean value but doesn't have side effects.

gunicorn/arbiter.py
```diff
@@ -126,28 +125,23 @@ def start(self):
         """
         self.log.info("Starting gunicorn %s", __version__)
 
+        if 'GUNICORN_PID' in os.environ:
+            self.master_pid = int(os.environ.get('GUNICORN_PID'))
+            self.proc_name = self.proc_name + ".2"
+            self.master_name = "Master.2"
```
tilgovi: It's not really a problem what we do here, but I wonder if maybe we just shouldn't rename the process. It's possible to tell which is which by the process tree and I'm just a little bit afraid the complexity isn't worth it. I can imagine bugs now or later where the process name ends up as "app.2.2.2.2.2". Maybe I'm just nervous because there was the recent tornado worker issue where the process name was incorrectly set.

I can see how it might make some shell script easier if you can grep for `.2$`, but if you're trying to automate this I think using the pid file is probably the better choice, anyway.

tilgovi:
Much better, I think.

gunicorn/arbiter.py
benoitc: I will rename it `maybe_promote_master` which is more explicit imo :)

gunicorn/sock.py
benoitc: right. removing it. About systemd, we should handle it in another commit.. The genertal idea here would be keeping in env the systemd fds that were given to us. So the `create_sockets` method would return a the listeners and external fds. I will submit a patch.

gunicorn/arbiter.py
benoitc: right now it can't happen since the string is fixed. When don't do a `+` but set the master name.

benoitc:
@tilgovi thanks for the review! fixes are coming :)

simlun:
Oops, it seems like the documentation still mentions the `.oldbin` PID file. https://github.com/benoitc/gunicorn/blob/master/docs/source/signals.rst should perhaps be updated as well? :)

berkerpeksag:
@simlun good catch! :) Would you like to send a PR?

benoitc:
so this is exactly why commenting on closed ticket is really a bad behaviour... No notifications are sent, nothing is tracked.

The documentation is still describing the old name of the the pid. Fixing it now.

benoitc:
never mind, it has been done but the current doc don't reflect it... So we just need to make a new release :) Will be done tomorrow. Sorry for the noise.

PR #1273, opened 2016-05-19
Allow to modify environment in pre_exec callback

urbaniak:
Calling pre_exec before copying the environment will allow to make environment modifications before spawning new master.


tilgovi:
I think it's okay.

benoitc:
Thanks!

Issue #376, opened 2012-07-01
Increase request related max limits or stop enforcing them


gnotaras:
Please either increase the maximum limits for --limit-request-line & --limit-request-field-size (currently hardcoded at 8190) or stop enforcing those limits once the user has used the aforementioned options in the configuration file or in the command line.

IMHO, enforcing a maximum size is not useful. I'd suggest that you provided some information in the documentation about how the application performance and security would be affected by setting those settings too high or too low, instead of hardcoding a max limit.

Thank you

janaki-sasidhar:
> What kind of line can be > 8192 ?

datatables

PR #1275, opened 2016-05-20
Bumping MAX_REQUEST_LINE to 32768

mistercrunch:
My application (https://github.com/airbnb/caravel) emits large URLs in
some rare cases. I like having stateful URLs in my app, but in some rare
corner cases: a markup widget where users can input a larger amount of
text, think documentation for the dashboard they are working on. 

I hate to refactor the entire app to use POST requests for this corner case
when it seems I can just bump the MAX_REQUEST_LINE and get more mileage.

Caravel typically runs on intranets where DOS attacks are not a concern.

From my understanding this change won't affect anyone, it will just
allow a wider spectrum when configuring their app. Note that other web
servers allow configuring this beyond 32k.

This past issue is related to this PR:
https://github.com/benoitc/gunicorn/issues/376

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -17,7 +17,7 @@ from gunicorn.http.errors import InvalidProxyLine, ForbiddenProxyRequest
 from gunicorn.six import BytesIO
 from gunicorn._compat import urlsplit
 
-MAX_REQUEST_LINE = 8190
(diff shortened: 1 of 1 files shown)


mistercrunch:
Nevermind, just found that I can use 0 as a setting for unlimited.

PR #1276, opened 2016-05-20
redtirect stdio/stderr to logfile

### LINKED ISSUE #1271: options to redirect stdout/stderr in gunicorn logging
following a discussion on twitter, I would like to rediscuss having an option to redtirect stdout/stderr to our logs:

screen shot 2016-05-18 at 13 29 17

This behaviour has been removed in 41523188bc05fcbba840ba2e18ff67cd9df638e9  3 years ago. But rediscussed last year and now.

I think we could have a `--capture-stdio` option that would allows it. Thoughts?

benoitc:
Add the new setting `--capture-stdio` to capture output from stdio/stderr to the current log file.

fix #1271

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -917,6 +917,7 @@ class WorkerTmpDir(Setting):
         If not set, the default temporary directory will be used.
         """
 
+
 class User(Setting):
     name = "user"
     section = "Server Mechanics"
@@ -1118,6 +1119,20 @@ class Loglevel(Setting):
         * warning
         * error
         * critical
+        """
+
+
+class CaptureOutput(Setting):
+    name = "capture_output"
+    section = "Logging"
+    cli = ["--capture-output"]
+    validator = validate_bool
+    action = 'store_true'
+    default = False
+    desc = """\
+        Redirect stdout/stderr to Error log.
+
+        .. versionadded:: 19.6
         """
 
 
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -12,6 +12,7 @@ from logging.config import fileConfig
 import os
 import socket
 import sys
+import threading
 import traceback
 
 from gunicorn import util
@@ -186,6 +187,8 @@ class Logger(object):
         self.access_log.propagate = False
         self.error_handlers = []
         self.access_handlers = []
+        self.logfile = None
+        self.lock = threading.Lock()
         self.cfg = cfg
         self.setup(cfg)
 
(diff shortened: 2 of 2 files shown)


benoitc:
screen shot 2016-05-20 at 23 07 34

gunicorn/config.py
```diff
@@ -1121,6 +1122,18 @@ class Loglevel(Setting):
         """
 
 
+class CaptureSTDIO(Setting):
+    name = "capture_stdio"
+    section = "Logging"
+    cli = ["--capture-stdio"]
```
berkerpeksag: Perhaps `--capture-output` would be a better name?

gunicorn/config.py
```diff
@@ -1121,6 +1122,18 @@ class Loglevel(Setting):
...
+    validator = validate_bool
+    action = 'store_true'
+    default = False
+    desc = """\
+        Redirect stdout/stderr to Error log.
```
berkerpeksag: Nedds a `.. versionadded:: 19.7` directive.

gunicorn/glogging.py
```diff
@@ -195,8 +198,16 @@ def setup(self, cfg):
         self.access_log.setLevel(logging.INFO)
 
         # set gunicorn.error handler
+        if self.cfg.capture_stdio and cfg.errorlog != "-":
```
berkerpeksag: Should we warn users when they pass `--error-log=- --capture-stdio`?

gunicorn/glogging.py
```diff
@@ -195,8 +198,16 @@ def setup(self, cfg):
...
+            for stream in sys.stdout, sys.stderr:
+                stream.flush()
+
+            self.logfile = open(cfg.errorlog, 'a+')
```
berkerpeksag: These 3 lines could be moved to a private method to avoid code duplication.

gunicorn/glogging.py
benoitc: I don't think so. if they pass `-`  then the --capture-stdio is redundant but afaik STDIO will be captured, won't it be?

gunicorn/config.py
benoitc: Isn't `stdio` more correct since it's not only stdout?

gunicorn/config.py
benoitc: fixed :)

gunicorn/glogging.py
benoitc: hrm not sure about it, it's more readable that way. But is there a rule somewhere about it?

gunicorn/config.py
berkerpeksag: IMO IO is a bit confusing since it doesn't do anything with stdin. CPython has a sort of similar (it temporarily redirects stdout and stderr to an io.StringIO instance) function named captured_output: https://github.com/python/cpython/blob/master/Lib/test/support/__init__.py#L1392

benoitc:
@berkerpeksag fixed the naming. I think we are good now :)

PR #1283, opened 2016-05-24
Add path and method to statsd

vishesh92:


tilgovi:
Haven't used statsd in a few months ... what will this do to existing stats and dashboards?

vishesh92:
@tilgovi hmm. I guess I need to keep the old stats as well to maintain the compatibility.

vishesh92:
@tilgovi I have updated the PR. Let me know if there are any other changes required.

benoitc:
@vishesh92 what this PR is about?

vishesh92:
@benoitc sorry for not adding the description.
This basically pushes metrics regarding every request to statsd. Currently only overall stats are sent to statsd.

benoitc:
PR looks ok , however i'm worried about the impact on performances of having so much logs to records. Did you ever bench it?

gunicorn/instrument/statsd.py
```diff
@@ -94,11 +94,16 @@ def access(self, resp, req, environ, request_time):
         Logger.access(self, resp, req, environ, request_time)
         duration_in_ms = request_time.seconds * 1000 + float(request_time.microseconds) / 10 ** 3
         status = resp.status
+        path = req.path.replace('/', '__')
```
berkerpeksag: Since you replaced all `/`s with `__`s, `path` doesn't look like a correct variable name here.

vishesh92:
I ran a small server with `--workers=2` option with 19.6.0. I ran ApacheBench for 100000 requests and 5 concurrency twice. Here is the output.

[with_statsd_my_changes.txt](https://github.com/benoitc/gunicorn/files/643955/with_statsd_my_changes.txt)
[with_statsd.txt](https://github.com/benoitc/gunicorn/files/643956/with_statsd.txt)
[without_statsd.txt](https://github.com/benoitc/gunicorn/files/643957/without_statsd.txt)

vishesh92:
@benoitc @berkerpeksag Should I make my changes a part of some configuration parameter?

berkerpeksag:
Perhaps we can add a `statsd_class` setting like `logger_class` to let people pass their customized `Statsd` class?

vishesh92:
@berkerpeksag The current implementation of `statsd` is done using `logger_class`. If someone wants to pass a customized `statsd` class, they can still do that using `logger_class`.

raags:
My two cents; This should probably be behind an optional flag (to avoid metric explosion) and ideally use tags instead of coding the path in the metric name. Tags would potentially require tag support to be added - dogstatsd tag format is a popular choice.

PR #1290, opened 2016-05-31
initialize  the group access list for a worker when initgroups is set

### LINKED ISSUE #1287: Incomplete privileges drop
Hello,

Seems uid/gid drop is incomplete on gunicorn. Setting uid doesn’t fetch real uid groups.
This behaviour is visible here :

```
# cat foo.py
#!/usr/bin/env python
import os
def application(env, response):
    response('200 OK', [('Content-Type', 'text/plain')])
    return [str((os.getuid(), os.getgid(), os.getgroups()))]

def nothing(status, headers):
    pass

if __name__ == '__main__':
    print(application(None, nothing))

# groups
root
# ./foo.py 
['(0, 0, [0])']  // Groups OK, [root]

# gunicorn -b 127.0.0.1:5000 foo:application
# curl localhost:5000
(0, 0, [0])  // Groups OK, [root]

# gunicorn -u www-data -g www-data -b 127.0.0.1:5000 foo:application
# curl localhost:5000
(33, 33, []) // Groups not OK, empty…

# su www-data -s /bin/sh
$ groups
www-data debian-tor
$ ./foo.py
['(33, 33, [33, 117])'] // Groups OK, [www-data, debian-tor]

$ gunicorn -b 127.0.0.1:5000 foo:application
$ curl localhost:5000
(33, 33, [33, 117])  // Groups OK, [www-data, debian-tor]
```

Running python script directly, groups are set for root (root) and www-data (www-data and debian-tor).
Running wsgi with gunicorn, ok from root and www-data.
Running wsgi with gunicorn from root and uid/gid www-data, not ok, groups are empty.

This is a problem because after privileges dropping, your app can’t access file with group restriction (in my case, unable to access file with debian-tor:debian-tor chown and 770 chmod).

Seems this is because groups are not initializated (`os.initgroups(username, groups)`) before privileges drop.
[This patch](https://github.com/aeris/gunicorn/commit/696c4bfd57ff851018687f7011a8e57608fcc367) may fix the trouble, but is not portable (UNIX only because of `pwd` package usage).

benoitc:
Allows the possibility to initialise the group access list when needed to separate privileges

fix #1287

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -987,6 +987,23 @@ class Umask(Setting):
         """
 
 
+class Initgroups(Setting):
+    name = "initgroups"
+    section = "Server Mechanics"
+    cli = ["--initgroups"]
+    validator = validate_bool
+    action = 'store_true'
+    default = False
+
+    desc = """\
+        If true, set the worker process's group access list with all of the
+        groups of which the specified username is a member, plus the specified
+        group id.
+
+        .. versionadded:: 19.7
+        """
+
+
 class TmpUploadDir(Setting):
     name = "tmp_upload_dir"
     section = "Server Mechanics"
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -54,6 +54,7 @@ hop_headers = set("""
 
 try:
     from setproctitle import setproctitle
+
     def _setproctitle(title):
         setproctitle("gunicorn: %s" % title)
 except ImportError:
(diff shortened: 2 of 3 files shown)


tilgovi:
I guess it is okay, but is this common for daemons? Wouldn't it be more common to create a group specific to the purpose of the server, like is common with a www-data group?

benoitc:
@tilgovi it's not that common I think. 

Also only the superuser or a user with sufficient privileges can also do that. Hence why I put it as an option. Generally both program that share a files are member of the same group or the correct flag is given to it.

it works here:

```sh
[~] ps -eo user,pid,tty,command |grep python                                                                                                                                           18:23:22
root            18248 ttys001  /Users/benoitc/local/venvs/py3.5/bin/python /Users/benoitc/local/venvs/py3.5/bin/gunicorn -w3 --initgroups --user=501 --group=20 test:app
benoitc         18251 ttys001  /Users/benoitc/local/venvs/py3.5/bin/python /Users/benoitc/local/venvs/py3.5/bin/gunicorn -w3 --initgroups --user=501 --group=20 test:app
benoitc         18252 ttys001  /Users/benoitc/local/venvs/py3.5/bin/python /Users/benoitc/local/venvs/py3.5/bin/gunicorn -w3 --initgroups --user=501 --group=20 test:app
benoitc         18253 ttys001  /Users/benoitc/local/venvs/py3.5/bin/python /Users/benoitc/local/venvs/py3.5/bin/gunicorn -w3 --initgroups --user=501 --group=20 test:app
benoitc         20340 ttys003  grep --color=auto --exclude-dir=.bzr --exclude-dir=CVS --exclude-dir=.git --exclude-dir=.hg --exclude-dir=.svn python
```

benoitc:
merging it, the patch work. We could eventually make it the default but it will requires more testing on different platforms.

PR #1291, opened 2016-06-02
Add environment variables to gunicorn access log format

vivianho:
Add environment variables to gunicorn access log format using `{Variable}e`.

Updated the docs to include the new identifier.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -471,31 +471,30 @@ access_log_format
 
 The access log format.
 
-===========  ===========
-Identifier   Description
-===========  ===========
-h            remote address
-l            ``'-'``
-u            user name
-t            date of the request
-r            status line (e.g. ``GET / HTTP/1.1``)
-m            request method
-U            URL path without query string
-q            query string
-H            protocol
-s            status
-B            response length
-b            response length or ``'-'`` (CLF format)
-f            referer
-a            user agent
-T            request time in seconds
-D            request time in microseconds
-L            request time in decimal seconds
-p            process ID
-{Header}i    request header
-{Header}o    response header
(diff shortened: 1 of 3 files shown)


benoitc:
thanks for the patch! But settings.rst shouldn't be edited. Instead you should update the config.py file in the AccessLog class. Can you resend your PR with that change?

vivianho:
@benoitc updated docs.

benoitc:
thanks!

tuukkamustonen:
Just saw this released into 19.7.0.

Did you consider the performance impact? Iterating the environment variable map for each log line feels unnecessary as you rarely change the environment variables in-process.

Why not populate it at the startup and re-use later on? I guess the performance impact just is negligible?

RonRothman:
+1.  I'm sorry I didn't notice this PR earlier.  I agree that the
performance impact of this change does look substantially negative.  (Our
use of Gunicorn is highly sensitive to latency and CPU usage, and we've
found during profiling that logging is particularly malperformant.)

Not sure what the best course of action here is: do nothing; fix this in a
19.7.1 release; something else?

berkerpeksag:
+1 for fixing this in a bugfix release.

benoitc:
actually do we know if it's really an issue? Im not sure having the env variable prepopulated vs using the dict will give much (except maybe by writing it to a string first...

All this access logging is fragile anyway and it's why it's not enabled by default (if you use gunicorn behind a proxy you don't need it) . IMO, while we are on it a better fix would be putting the logging in its own thread/process instead of doing it synchronously like we're doing now imo. I have the feeling that any  other fix would just distract us from the real one. Thoughts?

PR #1297, opened 2016-06-23
Update signals.rst

gfreezy:
USR2 signal does not rename master pid to .pidold any more.


tilgovi:
Good catch. Thank you.

PR #1301, opened 2016-06-29
http: large file upload improvement

hachreak:
- Implements a simple heuristic to improve the large file upload.

Signed-off-by: Leonardo Rossi leonardo.r@cern.ch

### DIFF
--- a/gunicorn/http/body.py
+++ b/gunicorn/http/body.py
@@ -208,11 +208,13 @@ class Body(object):
             self.buf.write(rest)
             return ret
 
+        read_size = 1024
         while size > self.buf.tell():
-            data = self.reader.read(1024)
+            data = self.reader.read(read_size)
             if not len(data):
                 break
             self.buf.write(data)
+            read_size = read_size * 2
 
         data = self.buf.getvalue()
         ret, rest = data[:size], data[size:]


tilgovi:
Can you explain this a bit? Is doubling every iteration a well understood idea from common practice (I've not seen this before)? Is there something simple that we can do that's better than this?

benoitc:
@hachreak do you have any figures that measure the improvement? How large are your uploads?

@tilgovi i find it uncommon too :) As i read it could be interesting maybe to offers a way to increase the number of bytes that gunicorn can read at the same time.  Not sure what could be the naming of such option yet though

hachreak:
In zenodo I have a REST API to upload file and I was testing the upload of 1gb file.
I think you can try to reproduce it with a simple flask application a try to upload big files.

In the image you can see how much cpu gunicorn consume.

uploading2

profile_before

Legend:
 - t0: when I'm starting the upload
 - t1: when the transfer is ending and gunicorn is starting to work
 - t2: when gunicorn is ending and the zenodo endpoint is starting to process the data

You can see from `t1` to `t2` the cpu usage.
I think because the hard-coded behaviour to read/write "only" 1kb at time.
It's ok for small file but it can painful for big file.

My heuristic say: the number of byte read each time is directly proportional with the number of times you read because if you read a lot of times there is a good possibility that there are a lot of bytes waiting.

Probably there is a better way to do that, but, at least, it looks decreasing the cpu usage.

After:

uploading_after2

profile_after

jamadden:
Increasing read sizes over time is also a heuristic I've used to great effect in cases like this.

I would suggest a maximum upper bound on `read_size` to prevent reading *too* much at one time and thus churning through ever larger temporary allocations (they have to be contiguous so larger and larger sizes get harder and harder to do). There's usually a sweet spot around a couple MB, depending on a bunch of factors. 

If `self.reader` implements [recv_into](https://docs.python.org/2/library/socket.html#socket.socket.recv_into) or `readinto` so that the same buffer can be reused for each call, that's potentially another massive improvement.

Another improvement might be to have `self.buf`, instead of being a `BytesIO`, be a [SpooledTemporaryFile](https://docs.python.org/2/library/tempfile.html#tempfile.SpooledTemporaryFile). It could have a suitably large `max_size` (because the current `max_size` is effectively "all available memory"), but still not one so big that large uploads (especially concurrently to the same process) explode. (This would probably need another option to control the absolute max size of an upload, to prevent using up the complete disk. This is not an uncommon option among app servers like Tomcat.)

hachreak:
> I would suggest a maximum upper bound on read_size to prevent reading too much at one time and thus churning through ever larger temporary allocations

@jamadden I was thinking the same :D 

@benoitc I think my solution can be somehow a "start point" to improve more this part but more work needs to be done because it's true that it  decreases the cpu usage, but the overhead (in terms of time) is not enough decreased :D

tilgovi:
I would be more than happy to merge this as a first effort to improve uploads as long as we cap it at a reasonable value.

benoitc:
I'm quite dubious it worth to increase the read size that way. By default we max the size read on the socket to 8192. (see the unreader code). So if you want to read more , a fix should also be done there. Did you atually notice a difference > 8192?

jamadden:
@hachreak I think @benoitc is right, it makes sense to measure, especially if there's a cap on an underlying level...that would need to be changed also/instead.

benoitc:
the code I'me referring is the following:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/unreader.py#L58-L65

diwu1989:
Why not to make this even simpler and just fix the read size to be 8KB to begin with?

tilgovi:
If we were to simply change it, I suggest maybe 4KiB because that's a typical page size.

I like the idea of the spooled temporary file the most, I think. Or simply capping it is fine.

I think anything is an improvement as long as it doesn't grow unbounded.

benoitc:
what's the status of this PR. no activity since a while....

hachreak:
hi @benoitc, actually I didn't have time to try to make something better than this.. T___T
I don't have a so deep knowledge of `gunicorn`. 
Did you manage to reproduce the problem?

benoitc:
@tilgovi for the record, we had in the past a [TeeInput](https://github.com/benoitc/gunicorn/blob/c2f12a4977836245cdbe3b4a444bc35ffbe0fb02/gunicorn/http/tee.py) to handle temporary uploads in the past. It has been removed because it could let temporary file behind in case of crashes.

benoitc:
Is there some actions to do on that PR?

tilgovi:
It seems like this solution does not have consensus.

I'd like to refer to this conversation also to my comment at https://github.com/benoitc/gunicorn/issues/1542#issuecomment-328629908

I believe we should be using `io.DEFAULT_BUFFER_SIZE` in a few place as 1024 is likely unnecessarily small.

tilgovi:
Also, some smart re-implementation of our `Body` and other reader classes in terms of the `io` module is probably overdue. Use of `readinto()` and a better honoring of the `size` parameter to `read()` should allow an application to read more efficiently if it wants to.

PR #1304, opened 2016-07-07
Add --paste-global-conf option

moriyoshi:
While bootstraping gunicorn with `paster` (PasteScript) or `pserve` (Pyramid) is now discouraged, no options are provided for PasteDeploy users to set global configuration variables from the commandline.

This patch adds the new commandline option `--paste-global-conf KEY=VALUE` so that the users can pass arbitrary global_conf values to the PasteDeploy entrypoint.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1791,7 +1791,7 @@ class PasteGlobalConf(Setting):
     name = "raw_paste_global_conf"
     action = "append"
     section = "Server Mechanics"
-    cli = ["--paste-global-conf"]
+    cli = ["--paste-global"]
     meta = "CONF"
     validator = validate_list_string
     default = []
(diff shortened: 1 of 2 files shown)


gunicorn/config.py
```diff
@@ -203,6 +204,25 @@ def sendfile(self):
...
+    def paste_global_conf(self):
+        raw_global_conf = self.settings['raw_paste_global_conf'].get()
+        if raw_global_conf is None:
+            return None
+
+        global_conf = {}
+        for e in raw_global_conf:
+            s = _compat.bytes_to_str(e)
+            try:
+                k, v = re.split(r'(?<!\\)=', s, 1)
```
tilgovi: The regular expression here is a raw string, so should there be only one backslash?

tilgovi: Ignore me. Right. The backslash here is not for character escape but for regular expression escape. Whole reason for the raw is to not have `\\\\`.

:space_invader: :moon: :rocket:

tilgovi:
LGTM

tilgovi:
Maybe we should add to the docstring that it can be specified multiple times? Also, I wonder if `--paste-global` is better than `--paste-global-conf`. When I see the `-conf` suffix I think of a full configuration, like if it were parsed as JSON syntax or something. Maybe that is just me. How about `--paste-var` to make it very explicit that it's a single variable?

moriyoshi:
* Adding docstring
  
  Makes sense to me.

* Changing `--paste-global-conf` to `--paste-global` or `--paste-var`

  Also sounds reasonable.  For me `--paste-global` looks better because there is a concept of "local configuration".

tilgovi:
Thanks!

moriyoshi:
Thanks for taking care of this!

PR #1309, opened 2016-07-17
only the arbiter should unlink the unix socket

### LINKED ISSUE #1298: unix socket gets deleted on worker restart
We have a django app deployed on heroku with gunicorn 19.6.0
It is binding to a unix socket that's used by Nginx.

These are the options that we're using:

```
  --max-requests 500 \
  --max-requests-jitter 25 \
  --preload \
  --threads 2 \
  --timeout 27 \
  --workers 3 \
  --bind unix:/tmp/nginx.socket
```

What we're seeing is that the socket file disappears after a while.
Downgraded to gunicorn 19.4.5 and things are working fine.

Is the nginx socket file being deleted accidentally?

benoitc:
make unlink explicitly done by the arbiter.

 fix #1298

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -368,6 +368,7 @@ class Arbiter(object):
         if self.reexec_pid == 0 and self.master_pid == 0:
             for l in self.LISTENERS:
                 l.close()
+                l.unlink()
 
         self.LISTENERS = []
         sig = signal.SIGTERM
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -65,7 +65,7 @@ class BaseSocket(object):
         self.sock = None
 
     def unlink(self):
-        return True
+        return
 
 
 class TCPSocket(BaseSocket):
@@ -99,7 +99,7 @@ class UnixSocket(BaseSocket):
 
     FAMILY = socket.AF_UNIX
 
-    def __init__(self, addr, conf, log, fd=None):
+    def __init__(self, addr, conf, log, fd=None, unlink=True):
         if fd is None:
             try:
                 st = os.stat(addr)
@@ -111,6 +111,7 @@ class UnixSocket(BaseSocket):
                     os.remove(addr)
                 else:
                     raise ValueError("%r is not a socket" % addr)
+        self.can_unlink = unlink
         super(UnixSocket, self).__init__(addr, conf, log, fd=fd)
 
     def __str__(self):
@@ -126,6 +127,8 @@ class UnixSocket(BaseSocket):
         super(UnixSocket, self).close()
 
     def unlink(self):
+        if not self.can_unlink:
+            return
         os.unlink(self.cfg_addr)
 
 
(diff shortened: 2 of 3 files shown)


diwu1989:
what if the unix socket was already created before gunicorn starts, for example, if another process like nginx created the socket first?

would gunicorn still try to unlink the unix socket?

benoitc:
@diwu1989 normally yoou should start gunicorn first. we don't have yet a way to handle such case except when you're using systemd. Is the current patch working for you?

tilgovi:
I'm noticing that because LISTEN_FDS doesn't get propagated this behavior is broken for systemd. If we were to more thoroughly refactor this, do you think we can make this work? I'm thinking that instead of managing this on the socket, we could a symmetric `close_sockets` function in the `sock` module with an `unlink` parameter.

In the arbiter:

```python

def reexec():
   # ... after fork ...
   if self.systemd:
       environ['LISTEN_PID'] = os.getpid()
   else:
        environ['GUNICORN_FD'] = ",".join([str(fd) for fd in fds])
    # ... and so on ...

def stop():
    unlink = self.reexec_pid == 0 and self.master_pid == 0 and not self.systemd
    close_sockets(self.LISTENERS, unlink=unlink)
```

tilgovi:
We could parse the systemd environment cleanly, isolating this into a separate thing.

```python
# gunicorn.systemd module
import os

def listen_fds():
    if ('LISTEN_PID' in os.environ
            and int(os.environ.get('LISTEN_PID')) == os.getpid()):
    # ...
```

The arbiter could then keep all the `GUNICORN_FDS` logic, instead of the `sock` module. The arbiter checks for systemd sockets, sets `self.systemd`. Otherwise, it checks for fds in `GUNICORN_FDS`. If either are found, it passes these to `create_sockets`. The `create_sockets` function can use `getsockname()` and the difference between systemd and gunicorn inheritance goes away. The arbiter is now responsible for deciding whether gunicorn ultimately unlinks, by knowing where it got the fds.

tilgovi:
Untested, but this is the sort of refactor I am thinking about: https://github.com/benoitc/gunicorn/compare/fix/1298-2

benoitc:
@tilgovi I like the idea! Do you mind to create a PR for it? So we can comment the code directly :)

tilgovi:
#1310 with tests

tilgovi:
Could use additional tests for socket creation, especially inheritance and systemd activation. I only really added tests that assert unlink is done correctly.

diwu1989:
just tested this out and it fixes our problem

tilgovi:
@benoitc I don't mind landing this patch fix and deferring the bigger refactoring for R20.

diwu1989:
hi, can this be release soon please?
thanks

kaniini:
We hit this on 19.6 and had to downgrade to 19.4.5 to solve it.  This issue is causing many people's applications to break in strange ways because workers aren't being properly terminated causing all sorts of resource exhaustion issues.

I would say a release with this fix is critical.

tilgovi:
Fixed by #1310

PR #1313, opened 2016-07-27
1306 remove upper limit on max header size config

tobgu:
I re-generated the settings.rst file to reflect the updates made to config.py. Not sure if this was a good idea. Partly because some previous changes (not related to this PR) that were previously not included now got included. Partly because some examples now reflect my personal settings rather than those of @benoitc. I can drop those changes if you want.

Apart from that I'd be happy for any comments!


benoitc:
@tobgu patch is ok. Thanks forit! :)

It would be indeed better to have the doc regeneration outside the `settings.py` update. Can you make the change?

tobgu:
Sure, no problem. I'm away from computers for a couple of days but will fix it once i get back to a sensible device.

tobgu:
Removed the generated RST.

tobgu:
Can this be merged or are there additional things you would like me to update?

gunicorn/config.py
```diff
@@ -793,8 +793,9 @@ class LimitRequestFieldSize(Setting):
     desc = """\
         Limit the allowed size of an HTTP request header field.
 
-        Value is a number from 0 (unlimited) to 8190. to set the limit
-        on the allowed size of an HTTP request header field.
+        Value is a positive number or 0. Setting it to 0 will allow unlimited header field sizes.
+
+        Setting this parameter to a very high or unlimited value can open up for DDOS attacks.
```
berkerpeksag: Since this is important, we could make it a warning:

```rst
.. warning::
   Setting [...]
```

.gitignore
```diff
@@ -13,3 +13,5 @@ examples/frameworks/pylonstest/PasteScript*
 examples/frameworks/pylonstest/pylonstest.egg-info/
 examples/frameworks/django/testing/testdb.sql
 .tox
+venv
```
berkerpeksag: Thank for updating `.gitignore`, but these can live in your global `.gitconfig`.

gunicorn/config.py
```diff
@@ -793,8 +793,9 @@ class LimitRequestFieldSize(Setting):
     desc = """\
         Limit the allowed size of an HTTP request header field.
 
-        Value is a number from 0 (unlimited) to 8190. to set the limit
-        on the allowed size of an HTTP request header field.
+        Value is a positive number or 0. Setting it to 0 will allow unlimited header field sizes.
```
berkerpeksag: Also, please keep line lengths < 80 chars

gunicorn/config.py
tobgu: Yes, fixed!

.gitignore
tobgu: Added the venv because i figured the Makefile made assumptions about the location of the venv but I've removed that now.

gunicorn/config.py
tobgu: Fixed!

berkerpeksag:
Thanks!

Issue #1305, opened 2016-07-07
workertmp -> notify -> os.fchmod performance problem


fillest:
Hello.
We're currently trying to run our web app on gunicorn under tens of thousands requests/second of live traffic on AWS. A fast sampling profiler is continuosly enabled in production so I can check out the monitoring graphs and correlate the problematic moments with the profiler per-second reports. Basically, it shows top N stack traces which were most often sampled during a second.
Periodically I see something like this (percentage varies, e.g. it may be 18%) in the log:

```
30.3% gunicorn/workers/workertmp.py notify:43 'os.fchmod(self._tmp.fileno(), self.spinner)'
      <-  gunicorn/workers/base.py notify:70
      <-  gunicorn/workers/sync.py run_for_one:59
      <-  gunicorn/workers/sync.py run:122 
      <-  gunicorn/workers/base.py init_process:128 
      <-  site-packages/gunicorn/arbiter.py spawn_worker:517 
      <-  site-packages/gunicorn/arbiter.py spawn_workers:549 
      <-  site-packages/gunicorn/arbiter.py manage_workers:484 
      <-  site-packages/gunicorn/arbiter.py run:179 
      <-  gunicorn/app/base.py run:72 
      <-  gunicorn/app/base.py run:192 
      <-  gunicorn/app/wsgiapp.py run:74 
      <-  venv/bin/gunicorn <module>:11
```

So this means that of an observed second 30% of the wall clock time of a whole web app worker process were spent in `gunicorn/workers/workertmp.py -> notify:43  -> os.fchmod(self._tmp.fileno(), self.spinner)` (and it is the hottest stack sample). The similar behavior is observed for all other workers which also run at the same time.
I would like to understand what does this part of gunicorn do and if there is an "easy" way to optimize it somehow?

We run on AWS on some large instance, Ubuntu, gunicorn==19.5.0, gunicorn is set up the following way: 4 copies of gunicorn, each has workers = 4, preload_app = true, all this behind nginx, which load balances evenly between them. Under load all CPU cores are utilized to 100% (this is mainly user cpu time of our app code).

fillest:
Oh, looks like it happens when the main disk is full (it currently gets full periodically due to nginx log spam). Our /tmp/ is on / (not memfs), maybe it could be related

tilgovi:
Oh, looks like it happens when the main disk is full (it currently gets full periodically due to nginx log spam). Our /tmp/ is on / (not memfs), maybe it could be related

[...]

PR #1329, opened 2016-08-12
docs: a warning for disk-backed --worker-tmp-dir #1305

fillest:


berkerpeksag:
Thanks for the PR! Perhaps http://docs.gunicorn.org/en/stable/faq.html#troubleshooting or somewhere else would be a better location for this. We can then add a link to the FAQ entry in `worker_tmp_dir` documentation. Also, `settings.rst` is an auto-generated file. You will need to edit `gunicorn/config.py`.

JayH5:
Maybe worth mentioning that in some environments it's not always as straightforward to mount filesystems - e.g. inside Docker containers. Would be great if the heartbeat system could not use files.

benoitc:
bump. @berkerpeksag what should we do? @JayH5 it's planned :)

berkerpeksag:
If @fillest doesn't have time to address my comments, I can do it sometime next week.

fillest:
I will try to allocate the time, sorry for the delays

benoitc:
bump

fillest:
Ok, I've moved it to faq.rst. But I'm not sure how to properly add a link to the FAQ - I mean, what url to use (or if there's some markup that should be used so that url will be generated)?

(Also it may be better to squash the commits on merging this PR.)

berkerpeksag:
> (or if there's some markup that should be used so that url will be generated)?

Thanks for the updated PR. Yes, there is a markup for this: http://www.sphinx-doc.org/en/stable/markup/inline.html#role-ref

Also, it would be great if you wrap the lines at 80 characters.

fillest:
done

tilgovi:
Thanks!

PR #1334, opened 2016-08-23
feat(import_app): print original exception on AppImportError

ptbrowne:
Useful when you have a NameError in the code that creates your app.

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -388,8 +388,7 @@ def import_app(module):
         app = eval(obj, mod.__dict__)
     except NameError:
         exc_type, exc_value, exc_traceback = sys.exc_info()
-        traceback.print_exception(exc_type, exc_value, exc_traceback,
-                                  limit=3, file=sys.stdout)
+        traceback.print_exception(exc_type, exc_value, exc_traceback)
         raise AppImportError("Failed to find application: %r" % module)
 
     if app is None:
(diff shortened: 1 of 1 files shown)


gunicorn/util.py
```diff
@@ -387,6 +387,8 @@ def import_app(module):
     try:
         app = eval(obj, mod.__dict__)
     except NameError:
+        exc_type, exc_value, exc_traceback = sys.exc_info()
+        traceback.print_exception(exc_type, exc_value, exc_traceback)
```
benoitc: this should only be done when debug mode is enabled. If not it can expose some unwanted to externals systems looking at the output.

ptbrowne: By debug mode, do you mean checking the logging level ?

benoitc: @ptbrowne yes :)

tilgovi: Please either use star args to avoid locals or explicitly delete the traceback. See #1228.

tilgovi:
@ptbrowne I think this is waiting for very minor changes from you. Please let us know if you will have time to do this.

ptbrowne:
@tilgovi, sorry for the wait, it should be ok now :)

tilgovi:
Not a long wait at all, @ptbrowne! I just happened to process this in a batch of commenting on other issues and PRs. I've been high volume, high latency in my attention to Gunicorn lately.

Looks good to me.

PR #1339, opened 2016-08-27
Add SO_REUSE socket option.

zimathias:


gunicorn/sock.py
```diff
@@ -38,6 +38,12 @@ def __getattr__(self, name):
 
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        if self.conf.socket_reuseport:
+            try:
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+            except AttributeError:
+                # not all platforms have SO_REUSEPORT.
```
tgwizard: Would it make more sense to log and/or crash here, instead of silently continuing?

benoitc: warning should be enough imo.

berkerpeksag:
#1344 also implements the same feature. I think the main question is do we need to add a new option to enable/disable (#1344 unconditionally use `SO_REUSE` if it's available so someone may want to disable it) this?

berkerpeksag:
Thanks for the PR, @zimathias. We've decided that there is no need to add a new config for this: https://github.com/benoitc/gunicorn/pull/1344#issuecomment-249186918

I'm closing this as duplicate of #1131 and #1344.

PR #1345, opened 2016-09-05
added appier frameworks example

vickysam:
added appier frameworks example.
Appier- An object-oriented Python web framework built for super fast app development.

### DIFF
--- a/examples/frameworks/appierapp.py
+++ b/examples/frameworks/appierapp.py
@@ -1,0 +1,9 @@
+import appier
+
+class HelloApp(appier.App):
+
+    @appier.route("/", "GET")
+    def hello(self):
+        return "Hello, world!"
+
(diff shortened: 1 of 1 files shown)


vickysam:
added appier frameworks example

berkerpeksag:
Thanks for the pull request! :) I don't think we should add more framework examples. It looks like appier doesn't require any specific configuration for Gunicorn and doesn't seem to be a very popular choice at the moment.

vickysam:
I agree, Thanks :)

berkerpeksag:
If you'd like work on something else, I've just opened https://github.com/benoitc/gunicorn/issues/1346 :) You may want to wait for the feedback of other maintainers before starting to work on it though.

vickysam:
Sure,

I will start working on it.

PR #1354, opened 2016-09-21
[RFC] Add a worker_id attribute to workers

### LINKED ISSUE #1352: Allow for a way to determine the current worker ID
If I run an application in multi-process mode, it would be very useful for introspection to be able to determine which worker I actually am.  And by that I don’t mean the pid but an integer that goes from 1 to `--workers=WORKERS` (if a worker dies, the new one gets the worker id of it, not a new one).

An example for that is uwsgi’s `uwsgi.worker_id()` (that starts at 0 FWIW).

hynek:
This is my own attempt to fix #1352.

The ID starts at 1 and is increased for each new worker.  If a worker terminates, its ID is reused.  Therefore the highest ID is also the number of active workers.

It doesn’t add any APIs to actually retrieve it within applications (that could be added later?), but it’s already super useful in the `post_fork` callback (and any other callback that receives the `worker` argument for that matter).

Is this something that you would accept?  My understanding is that the master process is in any case single-threaded so there’s no need for locking?  Where would I put documentation?

Let me know what you think!

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -537,11 +537,20 @@ class Arbiter(object):
                                   "value": active_worker_count,
                                   "mtype": "gauge"})
 
+    def _next_worker_id(self):
+        """
+        Find the first free worker id for the next worker.
+        """
+        candidate = 1
+        while candidate in set(w.worker_id for w in self.WORKERS.values()):
+            candidate += 1
+        return candidate
+
     def spawn_worker(self):
         self.worker_age += 1
         worker = self.worker_class(self.worker_age, self.pid, self.LISTENERS,
                                    self.app, self.timeout / 2.0,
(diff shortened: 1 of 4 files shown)


hynek:
I’ve just realized that this is at odds with how reloading works in gunicorn (first spawn all workers and then start killing the old ones).

I’m keeping it open to start a discussion, if it has any support at all.  A possible solution would be to launch the new workers with explicit IDs (pass an optional argument into `spawn_worker()`).  But I’d like to hear your thoughts about it first.

hynek:
So in the end I managed to implement it using callbacks: https://gist.github.com/hynek/ba655c8756924a5febc5285c712a7946  So I guess I’m closing this PR but leaving the issue open because a more elegant way would be really nice.

PR #1357, opened 2016-09-28
fixed incorrect gunicorn.service file for SystemD

daavve:
The ExecStart didn't point to the gunicorn executable like it should.

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -224,8 +224,8 @@ systemd:
     PIDFile=/run/gunicorn/pid
     User=someuser
     Group=someuser
-    WorkingDirectory=/home/someuser/rootofdjangoproject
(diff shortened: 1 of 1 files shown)


tilgovi:
Note the documentation is an example. We do not package gunicorn for distribution and so we do not determine its install path. Just as you have changed the application directory to say `somedjangoproject`, the path to gunicorn is an example.

I still think `/usr/bin` is a more sensible path to have in the example, so I'm happy to merge this, but could you please replace `somedjangoproject` with something not Django specific? Gunicorn does not require Django.

If you could make that change and squash the commits that would be ideal.

daavve:
Absolutely!  I'd be happy to make the documentation less 
Django-specific.  I completely forgot that other frameworks run with 
Gunicorn.

I'll make sure to change the example to something as generic as possible.


-Dave

benoitc:
bump?

PR #1363, opened 2016-10-06
Fixed mistake in systemd section

ammgws:
systemd example starts the wrong service name. Could cause confusion for beginners.


tilgovi:
Nice catch. Thanks!

PR #1369, opened 2016-10-14
Add InotifyReloader and updated '--reload {poll,inotify}' configuration option

### LINKED ISSUE #1368: Support using inotify with --reload
`--reload` is a very useful option when debugging and working on an application. However, in some situations (such as running gunicorn inside of a VM or a Docker container on Mac with host mounted directories), the current file polling method can cause high CPU load due to the overhead of dealing with a virtualized filesystem.

Many of these filesystems now have support for the kernel's inotify API which allows applications to monitor for filesytem events instead of polling. This is much less resource intensive and can improve CPU usage of the Reloader in many situations. It would be nice to either have gunicorn support inotify by default (on supporting platforms) or via an extra argument.

mark-adams:
Currently, `--reload` uses FS polling to find out when files have changed. For some time, the Linux kernel has had a feature called inotify that allows applications to monitor for FS events without polling.

This commit adds a new `use-inotify` configuration option that will cause gunicorn to use the new `InotifyReloader` instead of the default `Reloader` when `reload` is enabled.

Using inotify can result in lower CPU consumption by gunicorn especially when working with virtualized filesystems or environments with a large number of watched files / directories.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -162,3 +162,4 @@ Wolfgang Schnerring <wosc@wosc.de>
 Jason Madden <jason@nextthought.com>
 Eugene Obukhov <irvind25@gmail.com>
 Jan-Philip Gehrcke <jgehrcke@googlemail.com>
+Mark Adams <mark@markadams.me>
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -255,8 +255,8 @@ Debugging
 reload
 ~~~~~~
 
-* ``--reload``
-* ``False``
+* ``--reload RELOADER_TYPE``
+* ``None``
 
 Restart workers when code changes.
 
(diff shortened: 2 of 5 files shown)


benoitc:
Thanks for the patch!

See my comment in #1368 I think it would be better to augment the `--reload` setting instead of having another setting in gunicorn. Can you do the change?

mark-adams:
@benoitc Absolutely! I'll take a look.

If we modify the usage of `--reload` to now require either `poll` or `inotify`, that will break existing scripts. I'm not super familiar with how your settings work in gunicorn but is there an existing method where we could say that `--reload` would be equivalent to `--reload=poll` while still allowing you to use `--reload=inotify`

mark-adams:
@benoitc Alright, I think I got it figured out. I've reworked things to work like this now:

* `--reload=poll` enables reloading and forces FS polling
* `--reload=inotify` enables reloading and forces inotify
* `--reload` enables reloading and uses inotify (if available) and falls back on FS polling.

Do you mind taking another look?

benoitc:
@mark-adams patch looks good. Can you eventually add some doc for it or at least specify somewhere in the doc that inotify is needed?

mark-adams:
@benoitc Absolutely! I added it and also ran `make html` and included the appropriate changes. :thumbsup:

benoitc:
merged, thanks!

mark-adams:
Thanks @benoitc! Any idea when the next release is planned? Just wondering when I'll see this hit PyPI. Just curious. 😃

benoitc:
@mark-adams i will drop a new release tonight or tomorrow :)

PR #1371, opened 2016-10-16
Add some more knobs to the setproctitle functionality

jsferrei:
-  the existing implementation falls short because start-stop-daemon is unable to work with gunicorn if setproctitle is used or even
installed
 - allow users to set the format of process names
 - allow users to override the "gunicorn" prefix
 - allow users to override the "master" and "worker" identifiers

Closes https://github.com/benoitc/gunicorn/issues/1372


jsferrei:
It looks like my doc updates caught some stale changes from previous pull requests.

PR #1373, opened 2016-10-17
Add a configurable amount of jitter between spawning new Workers.

rowillia:
We've been using this pathc in order to prevent thundering herds of new workers when hot
restarting some of our highest QPS services.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1832,6 +1832,7 @@ if sys.version_info >= (2, 7):
         Ciphers to use (see stdlib ssl module's)
         """
 
+
 class PasteGlobalConf(Setting):
     name = "raw_paste_global_conf"
     action = "append"
@@ -1858,7 +1859,7 @@ class SpawnWorkerSleepTime(Setting):
     section = "Worker Processes"
     cli = ['--spawn-worker-sleep-time']
     validator = validate_float
-    default = 0.1
+    default = 0.05
     desc = """\
     Minimum time to sleep between spawning workers in seconds.
 
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1854,15 +1854,15 @@ class PasteGlobalConf(Setting):
         .. versionadded:: 20.0
         """
 
-class SpawnWorkerSleepTime(Setting):
-    name = "spawn_worker_sleep_time"
+class SpawnJitterTime(Setting):
+    name = "spawn_jitter_time"
     section = "Worker Processes"
-    cli = ['--spawn-worker-sleep-time']
+    cli = ['--spawn-jitter-time']
     validator = validate_float
     default = 0.05
     desc = """\
     Minimum time to sleep between spawning workers in seconds.
 
     The amount of time spent sleeping between spawning new workers will jitter
-    from spawn-worker-sleep-time to 2 * spawn-worker-sleep-time.
+    from spawn-jitter-time to 2 * spawn-jitter-time.
     """
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -346,6 +346,15 @@ def validate_pos_int(val):
     return val
 
 
+def validate_float(val):
+    if not val:
+        return 0.0
+    try:
+        return float(val)
+    except TypeError:
+        raise TypeError("Value is not a float: %s" % val)
+
+
 def validate_string(val):
     if val is None:
         return None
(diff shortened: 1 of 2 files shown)


rowillia:
@benoitc @berkerpeksag  Anything I can change with this PR and the other one I've opened?  I work with @paulnivin at Lyft, we've been running both of these with very high loads for a while.  Happy to walk though this with you folks!

gunicorn/config.py
```diff
@@ -346,6 +346,15 @@ def validate_pos_int(val):
     return val
 
 
+def validate_float(val):
+    if not val:
+        return 0.0
```
benoitc: this would make the default to 0, where the current behaviour expect `0.1 * random.random()` .  There should be a decent default to have all the workers accepting  simultaneously. maybe `0.05` if we don't want to change the function. Thoughts?

gunicorn/config.py
```diff
@@ -1811,3 +1819,16 @@ class PasteGlobalConf(Setting):
 
         .. versionadded:: 20.0
         """
+
+class SpawnWorkerSleepTime(Setting):
+    name = "spawn_worker_sleep_time"
+    section = "Worker Processes"
+    cli = ['--spawn-worker-sleep-time']
```
benoitc: small nitpicking but this quite a long option name. Maybe something shorter would be better? like `--spawn-worker-sleep` or something more native?

benoitc:
@rowillia sorry i'm quite busy these days. I commented your patch. Otherwise I like the idea, so let's go further :)

gunicorn/config.py
RonRothman: FWIW I prefer descriptive names over short names, so a slight +1 for `--spawn-worker-sleep-time`.  (`--spawn-worker-sleep` sounds like it could be a boolean flag, in any case.)

gunicorn/config.py
```diff
@@ -1836,3 +1845,16 @@ class PasteGlobalConf(Setting):
...
+    validator = validate_float
+    default = 0.05
```
rowillia: @benoitc This should ensure that we have a default closer to the existing behavior, right?

gunicorn/config.py
rowillia: I think this is OK b/c I specify a default in `SpawnWorkerSleepTime`, right?  Adjusted the default value to more closely match the existing behavior.

rowillia:
Thanks for the feedback @benoitc !  If you feel strongly about the flag name I'm happy to change it 😄

tilgovi:
Maybe simply `spawn_jitter`? Our other jitter related option is `max_requests_jitter`. I think `spawn` implies worker spawn (we don't spawn anything else). So `spawn_jitter` would seem to be to be a descriptive name.

tilgovi:
Otherwise, LGTM.

rowillia:
Thanks @tilgovi !  Rebased and renamed that flag.

gunicorn/config.py
benoitc: no the  default should be the current one , so it doesn't break the current expectations ie 0.1 * random, which is still random where 0.05 isn't :)

tilgovi: Default value `0.05` and `* random.uniform(0, 2)` should give the same default expected value.

Old:
`0.1 * 0.5 = 0.05` average
`0.1 * 0 = 0` min
`0.1 * 1 = 0.1` max

New:
`0.05 * 1 = 0.05` average
`0.05 * 0 = 0` min
`0.05 * 2 = 0.1` max

No?

tilgovi:
I approved the changes, pending tweak to the default value.

gunicorn/config.py
tilgovi: Or keep it closer to what we have and just do `0.1` as the default and use `* random.random()` like before. Will that not work?

benoitc:
@rowillia ping?

tilgovi:
I think jitter is maybe not the right word here. This is not a period function. That is my fault. I will try to finish this soon. I am sorry for my bad reviewing :(.

gunicorn/config.py
benoitc: should work indeed. rethinking of it we need to keep the current default anyway, as some app (like one i have around is expecting it :)

benoitc:
`spawn_delay_ms` or just `spawn_delay` should be OKish, shouldn't it ?

tilgovi:
I feel bad we never merged this. It's a useful feature for high volume deployments. Can we agree on a name and a behavior for R20?

benoitc:
why did you closed itt? not useful anymore in gunicorn?

PR #1374, opened 2016-10-17
Add --delay-pidfile option for on the fly upgrades

rowillia:
When running very high QPS services, we've ran into race conditions where the pidfile gets created before all of the workers have finished booting, which then leads the wrapper we have for Gunicorn to handle hot restarts (Rainbow Saddle) killing the old master before the new master is fully booted.

This flag will delay pidfile creation until all of the workers have booted, ensuring we don't tear down the old master until the new one is ready.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -78,6 +78,13 @@ class Arbiter(object):
             0: sys.executable
         }
 
+    def _create_pidfile(self):
+        pidname = self.cfg.pidfile
+        if self.master_pid != 0:
+            pidname += ".2"
+        self.pidfile = Pidfile(pidname)
+        self.pidfile.create(self.pid)
+
     def _get_num_workers(self):
         return self._num_workers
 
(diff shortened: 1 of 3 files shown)


rowillia:
@benoitc Ping!  Anything I can do for this PR?

tilgovi:
I worry a little bit that different users may have different expectations around when gunicorn will start handling requests, and also how failure to start one or more new workers after USR2 might impact rollback if the pid file isn't updated.

Is there anything you can say to clarify or think of cases where we might have issues? In other words, I'm not opposed but we should handle with care.

benoitc:
pidfile should be created before workers are spawned. If not how would you know a new master process is created? There shouldn't be any expectation that the workers are already spawned when it's created, the pidfile is there to find the newly created master.

Normal boot flow is the following:
1. boot the master
2. create the pidfile
3. processeses of the new worker are spawned
4. Old workers are gracefully shutdown. while old connections should go on the backlog until a new worker accept it.

Why are you using rainbow-saddle ? Symlinking?

benoitc:
In other word I don't like the idea to have an option that break the usual expectation. If something is missing in term of upgrade, then please open a ticket and we will find a way to improve it.

tilgovi:
I agree with @benoitc. Let's not change the pidfile behavior but please open a ticket to discuss ways we might expose that new workers are "ready". Or maybe `post_worker_init` could be useful to you as well as a high listen backlog. Touching a file or notifying some monitor process, maybe by sending the new application code version, should let you build a custom solution to detect when new workers are mostly booted. If you think there's something we cold add to core, let's discuss on an issue for that topic.

benoitc:
@rowillia ping

benoitc:
@rowillia thanks for the patch. Following the comments above, i'm closing the PR though. Let's discuss of the other topics in a new issue or on irc.

PR #1375, opened 2016-10-17
Fix leak of duplicate file descriptor for bound sockets.

rowillia:
`socket.fromfd` does not close the original file descriptor, so we have to explicitly close it to avoid a leak.

See also:
http://bugs.python.org/issue10099


berkerpeksag:
Thanks!

PR #1376, opened 2016-10-19
Docs/system d change

daavve:
Fixed security issue in gunicorn.socket.  Also added information about how to set up Nginx to forward incoming traffic to Gunicorn using Systemd.


benoitc:
@daavve can you change the text to a 80 chars column so it's consistent with the rest of the doc? Otherwise the change looks good.

daavve:
Will do!

Thanks for the feedback.

-Dave

benoitc:
thanks!

PR #1378, opened 2016-10-27
Fix validate_reload by returning the validated value

mark-adams:
When '--reload=RELOADER_TYPE' was implemented, `validate_reload()` was
added but in one of the last refactorings, it lost the return statement
at the end of the function. As a result, the '--reload' config value was
totally broken.

This resolves the issue by adding the missing return.

My first commit failed because the default value of the setting (`None`) was not the same as the output of its validator when `None` was passed in. At this point, I realized that trying to keep ``--reload`` (with no argument) is virtually impossible because of the ambiguity when we allow `--reload` by itself and `--reload RELOADER_TYPE`. For instance, running `gunicorn --reload myapp:app` now results in an error because `myapp:app` is now seen as the argument to `--reload` instead of the app module name.

I think the best thing to do here is to eliminate the ambiguity by always requiring an argument to `--reload`, set the default to `off`, and add the `auto` option which does the inotify (if available) and fallback to polling.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -256,7 +256,7 @@ reload
 ~~~~~~
 
 * ``--reload RELOADER_TYPE``
-* ``None``
+* ``off``
 
 Restart workers when code changes.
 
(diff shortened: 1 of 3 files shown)


mark-adams:
@benoitc Probably should address this before doing a release.

benoitc:
thanks for the fix!

mark-adams:
You're welcome @benoitc! Thanks for the merge! Looking super forward to this making it into the next release! :smile:

PR #1381, opened 2016-11-03
Reject requests like "GET / HTTP/1a1"

mlgh:
Numbers must be separated by dot. This makes life
a little bit harder for attackers who would like to inject specially crafted packets after GET / (e.g. in nginx there are sometimes regular expressions like (?P<action>[^.]).html


berkerpeksag:
Could you please add a test case? Thanks!

mlgh:
Sure! Added a test and checked that it fails after removing backslash before "." in regex.

berkerpeksag:
Great patch, thanks! :)

mlgh:
Thanks! :)

PR #1384, opened 2016-11-06
Add --server_name and --no_server_name options #825

mathiasverhoeven:
Set `--server_name` to return a different HTTP Server response header than `gunicorn/19.6.0`.
Set `--no_server_name` to no longer return the header.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -463,7 +463,7 @@ server_name
 ~~~~~~~~~~~~~~~~~~~
 
 * ``--server_name STRING``
-* ``gunicorn\X.Y.Z``
+* ``gunicorn/X.Y.Z``
 
 Server name to use in the ``Server`` HTTP Header on responses.
 
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -459,6 +459,22 @@ Set to ``*`` to disable checking of Front-end IPs (useful for setups
 where you don't know in advance the IP address of Front-end, but
 you still trust the environment)
 
+server_name
+~~~~~~~~~~~~~~~~~~~
+
+* ``--server_name STRING``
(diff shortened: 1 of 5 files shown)


benoitc:
Thanks for the patch. I'm not sure I like the idea to add another option for gunicorn.  Maybe just removing the signature is enough? Why would like someone replace the signature of Gunicorn?

tilgovi:
I'm fine for both options, but the CLI flags should be hyphenated, not snake_case, to be consistent with the existing flags.

tilgovi:
My mistake. The CLI flags _are_ hyphenated, just not in the documentation.

docs/source/settings.rst
```diff
@@ -459,6 +459,22 @@ Set to ``*`` to disable checking of Front-end IPs (useful for setups
 where you don't know in advance the IP address of Front-end, but
 you still trust the environment)
 
+server_name
+~~~~~~~~~~~~~~~~~~~
+
+* ``--server_name STRING``
```
tilgovi: Here and below, is what I am referring to when I talk about hyphenation.

```
``--server-name STRING``
```

No?

mathiasverhoeven:
@benoitc I mainly wanted to add the option to choose between no header, `gunicorn` and `gunicorn/19.6.0`. But I agree that just being able to remove the header would address #825.
Would you be ok with just adding `--no_server_name`?

@tilgovi Thanks I will adjust the docs!

benoitc:
t

benoitc:
why are you closjng the issue? has it been solved?

RKursatV:
It's been 4 years and still no solution? Too bad

benoitc:
@RKursatV it has been decided to not do it. The version has been removed.

Issue #1360, opened 2016-10-01
Ability to pass in all config values as environment variables.


jmagnusson:
Would be nice to be able to pass in settings as environment variables. Perhaps prefixed with `GUNICORN_`, i.e. `bind = '127.0.0.1:5002'` becomes `export GUNICORN_BIND='127.0.0.1:5001'`

Perfect when using docker.

tilgovi:
I don't see any reason not to support it. Pull requests welcome.

tilgovi:
Feel free to give PRs for any or all options you would like to see supported as environment variables.

hramezani:
PR #1385

benoitc:
fixed in latest master

PR #1385, opened 2016-11-06
add ability to pass in all config values as environment variables.

hramezani:
add environment variables that starts with `GUNICORN_` like `GUNICORN_WORKERS` as settings.
skip variables that does not start with `GUNICORN_`, and also skip variable that not in `settings` like `GUNICORN_FOO`

### DIFF
--- a/gunicorn/app/base.py
+++ b/gunicorn/app/base.py
@@ -154,6 +154,17 @@ class Application(BaseApplication):
             if default_config is not None:
                 self.load_config_from_file(default_config)
 
+        # Load up environment configuration
+        env_vars = self.cfg.get_cmd_args_from_env()
+        if env_vars:
+            env_args = parser.parse_args(env_vars)
+            for k, v in vars(env_args).items():
+                if v is None:
+                    continue
+                if k == "args":
+                    continue
+                self.cfg.set(k.lower(), v)
+
         # Lastly, update the configuration with any command line
         # settings.
         for k, v in args.__dict__.items():
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -18,6 +18,7 @@ import re
 import ssl
 import sys
 import textwrap
+import shlex
 
 from gunicorn import __version__
 from gunicorn import _compat
@@ -68,6 +69,11 @@ class Config(object):
         if name not in self.settings:
             raise AttributeError("No configuration setting for: %s" % name)
         self.settings[name].set(value)
+
+    def get_cmd_args_from_env(self):
+        if 'GUNICORN_CMD_ARGS' in self.env_orig:
+            return shlex.split(self.env_orig['GUNICORN_CMD_ARGS'])
+        return []
 
     def parser(self):
         kwargs = {
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -285,3 +285,24 @@ def test_always_use_configured_logger():
     c.set('statsd_host', 'localhost:12345')
     # still uses custom logger over statsd
     assert c.logger_class == MyLogger
+
+
+def test_load_enviroment_variables_config(monkeypatch):
+    monkeypatch.setenv("GUNICORN_CMD_ARGS", "--workers=4")
+    with AltArgs():
+        app = NoConfigApp()
+        assert app.cfg.workers == 4
+
+
+def test_invalid_enviroment_variables_config(monkeypatch):
+    monkeypatch.setenv("GUNICORN_CMD_ARGS", "--foo=bar")
+    with AltArgs():
+        with pytest.raises(SystemExit):
+            NoConfigApp()
+
+
+def test_cli_overrides_enviroment_variables_module(monkeypatch):
+    monkeypatch.setenv("GUNICORN_CMD_ARGS", "--workers=4")
+    with AltArgs(["prog_name", "-c", cfg_file(), "--workers", "3"]):
+        app = NoConfigApp()
+        assert app.cfg.workers == 3
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -291,7 +291,7 @@ def test_load_enviroment_variables_config(monkeypatch):
     monkeypatch.setenv("GUNICORN_CMD_ARGS", "--workers=4")
     with AltArgs():
         app = NoConfigApp()
-        assert app.cfg.workers == 4
+    assert app.cfg.workers == 4
 
 
 def test_invalid_enviroment_variables_config(monkeypatch):
@@ -305,4 +305,4 @@ def test_cli_overrides_enviroment_variables_module(monkeypatch):
     monkeypatch.setenv("GUNICORN_CMD_ARGS", "--workers=4")
     with AltArgs(["prog_name", "-c", cfg_file(), "--workers", "3"]):
         app = NoConfigApp()
-        assert app.cfg.workers == 3
+    assert app.cfg.workers == 3
--- a/THANKS
+++ b/THANKS
@@ -163,3 +163,4 @@ Jason Madden <jason@nextthought.com>
 Eugene Obukhov <irvind25@gmail.com>
 Jan-Philip Gehrcke <jgehrcke@googlemail.com>
 Mark Adams <mark@markadams.me>
+Hasan Ramezani <hasan.r67@gmail.com>


gunicorn/config.py
```diff
@@ -69,6 +69,19 @@ def set(self, name, value):
             raise AttributeError("No configuration setting for: %s" % name)
         self.settings[name].set(value)
 
+    def parse_env_vars(self):
+        env_vars = {}
+        for k,v in self.env_orig.items():
+            if k.startswith('GUNICORN_'):
+                try:
+                    name = k.split('GUNICORN_', 1)[1].lower()
```
tilgovi: I don't think this can ever fail if `k.startswith('GUNICORN_')`.

gunicorn/app/base.py
```diff
@@ -154,6 +154,11 @@ def load_config(self):
             if default_config is not None:
                 self.load_config_from_file(default_config)
 
+        # Load up environment configuration
+        env_vars = self.cfg.parse_env_vars()
+        for k,v in env_vars.items():
```
tilgovi: small linting nit here, space between `,v`.

gunicorn/config.py
```diff
@@ -69,6 +69,19 @@ def set(self, name, value):
             raise AttributeError("No configuration setting for: %s" % name)
         self.settings[name].set(value)
 
+    def parse_env_vars(self):
+        env_vars = {}
+        for k,v in self.env_orig.items():
```
tilgovi: Same nitpick here about spacing.

gunicorn/config.py
```diff
@@ -69,6 +69,19 @@ def set(self, name, value):
...
+            if k.startswith('GUNICORN_'):
+                try:
+                    name = k.split('GUNICORN_', 1)[1].lower()
+                except:
+                    continue
+                if not name or name not in self.settings:
```
tilgovi: `name not in self.settings` should be sufficient, no?

hramezani:
@tilgovi  thanks, i fix problems.

berkerpeksag:
I also like the idea, thanks! I haven't reviewed the code yet, but we should document all `GUNICORN_` env variables before merging this.

tests/test_config.py
```diff
@@ -285,3 +285,25 @@ def test_always_use_configured_logger():
...
+    os.environ["GUNICORN_WORKERS"] = "4"
+    with AltArgs():
+        app = NoConfigApp()
+        assert app.cfg.workers == 4
+
+def test_not_loading_enviroment_variables_config():
+    os.environ["GUNICORN_FOO"] = "BAR"
+    with AltArgs():
+        app = NoConfigApp()
+        try:
```
berkerpeksag: This can be written as:

```py
with pytest.raises(AttributeError):
    app.cfg.too
```

tests/test_config.py
```diff
@@ -285,3 +285,25 @@ def test_always_use_configured_logger():
     c.set('statsd_host', 'localhost:12345')
     # still uses custom logger over statsd
     assert c.logger_class == MyLogger
+
+def test_load_enviroment_variables_config():
+    os.environ["GUNICORN_WORKERS"] = "4"
```
berkerpeksag: We need to unset these environment variables. I don't know what's the best practice to do this in pytest though (I'd suggest writing a context manager if we were using unittest -- perhaps you need a fixture?)

hramezani:
@berkerpeksag Thanks, 
this PR accept `KNOWN_SETTINGS` in this format `GUNICORN_setting_name`. setting names are in [Settings](http://docs.gunicorn.org/en/stable/settings.html). for example `GUNICORN_bind`, `GUNICORN_worker_class`, ...
Is this a good idea that a write a paragraph in settings page and mention that we can change all of settings in this page by setting environment  variables with abov format?

benoitc:
What is the use case for this change? Can you elaborate about it?

hramezani:
@benoitc i saw [`Ability to pass in all config values as environment variables`](https://github.com/benoitc/gunicorn/issues/1360) in issue list and @tilgovi`  commented:

> I don't see any reason not to support it. Pull requests welcome.

finally i decide to do this.

with this change we can change gunicorn settings with environment variables.

tilgovi:
There are times, especially in these days of containerization, where it is much easier to change an environment variable than to change the command line invocation or a configuration file.

tilgovi:
We can do this in a separate PR, but I'm wondering if there is any reason to clear these keys from the environment when we spawn workers. I'm just wondering if any settings might contain information that we should try to prevent from leaking to workers and possibly then into a response.

hramezani:
@benoitc @tilgovi 
Is there anything that I do?

berkerpeksag:
> Is this a good idea that a write a paragraph in settings page and mention that we can change all of settings in this page by setting environment variables with abov format?

It would be nice to add something like "Environment variable: `GUNICORN_FOO`" (or just plain `GUNICORN_FOO`) to every setting. For example, let's take a look at `bind`: http://docs.gunicorn.org/en/stable/settings.html#bind With "-b ADDRESS, --bind ADDRESS", we know that we can pass `-b` or `--bind` to `gunicorn`. We can document the env variable in the same way:

```
bind

* -b ADDRESS, --bind ADDRESS (CLI)
* GUNICORN_BIND (env variable)
* ['127.0.0.1:8000'] (default value)
```

hramezani:
@berkerpeksag thanks for reply, I add documents.

berkerpeksag:
Thanks, `docs/source/settings.rst` is automatically generated. I prefer modifying https://github.com/benoitc/gunicorn/blob/master/docs/gunicorn_ext.py#L39 instead of manually adding these names. You can then run `make -C docs html` to update `settings.rst` (note that `make html` will make some unrelated changes in `settings.rst` -- please discard them before committing your changes)

docs/source/settings.rst
```diff
@@ -17,6 +17,7 @@ config
 ~~~~~~
 
 * ``-c CONFIG, --config CONFIG``
+* ``GUNICORN_CONFIG (env variable)``
```
berkerpeksag: "(env variable)" can be removed. I used it as an annotation in my example :)

hramezani:
@berkerpeksag Thanks, whta is your opinion about adding a property like `cli` with name `env_var` to Setting class? this can be used for following purpose:

1. change the environment variable name(now variables name  are the same as the `name` property).
2. add variables to config if this propery has valid value.
3. disable this functionality for some setting.
4. generate document based on this property.

benoitc:
I'm not sure i'm comfortable with this change, it introduces an extra level of complexity in the configuration, having to document the variables to use is a sign. What about instead having a simple `GUNICORN_CMD_ARGS` (or whatever its name) where you pass the command line arguments that could be parsed like usual? Ie:

`GUNICORN_CMD_ARGS ="--bind=127.0.0.1 --workers=3"` ...

So we wouldn't have to document each changes and could reuse the current code to handle it. Thoughts?

hramezani:
@benoitc Thanks, I change PR as your previous comment.

decal:
Looks even better than the Apache 2.4 envvars support..

gunicorn/config.py
```diff
@@ -69,6 +69,12 @@ def set(self, name, value):
             raise AttributeError("No configuration setting for: %s" % name)
         self.settings[name].set(value)
 
+    def get_gunicorn_env_var(self):
+        env_vars = []
```
berkerpeksag: Unless I'm missing something obvious, `env_vars` is not needed here:

```py
if 'GUNICORN_CMD_ARGS' in self.env_orig:
    return self.env_orig['GUNICORN_CMD_ARGS'].split()
return []
```

gunicorn/config.py
```diff
@@ -69,6 +69,12 @@ def set(self, name, value):
...
+        if 'GUNICORN_CMD_ARGS' in self.env_orig:
+            env_vars = self.env_orig['GUNICORN_CMD_ARGS'].split()
```
berkerpeksag: Did you consider using `shlex.split()` here?

gunicorn/config.py
```diff
@@ -69,6 +69,12 @@ def set(self, name, value):
             raise AttributeError("No configuration setting for: %s" % name)
         self.settings[name].set(value)
 
+    def get_gunicorn_env_var(self):
```
berkerpeksag: I'm not a native English speaker, but perhaps we can rename it to something like `get_cmd_args_from_env`? `get_gunicorn_env_var` sounds too generic to me. Ping @tilgovi :)

gunicorn/app/base.py
```diff
@@ -154,6 +154,17 @@ def load_config(self):
             if default_config is not None:
                 self.load_config_from_file(default_config)
 
+        # Load up environment configuration
+        env_vars = self.cfg.get_gunicorn_env_var()
+        if env_vars:
+            env_args = parser.parse_args(env_vars)
+            for k, v in env_args.__dict__.items():
```
berkerpeksag: Wouldn't `vars(env_args).items()` be better?

tests/test_config.py
```diff
@@ -285,3 +285,15 @@ def test_always_use_configured_logger():
...
     # still uses custom logger over statsd
     assert c.logger_class == MyLogger
+
+def test_load_enviroment_variables_config():
+    os.environ["GUNICORN_CMD_ARGS"] = "--workers=4"
+    with AltArgs():
+        app = NoConfigApp()
+        assert app.cfg.workers == 4
+
+def test_cli_overrides_enviroment_variables_module():
```
berkerpeksag: Tests are look great, thank you! :) Could you also add a test that passes an invalid config option (e.g `os.environ["GUNICORN_CMD_ARGS"] = "--monty-python"`?

hramezani:
@berkerpeksag Thanks, i fix all your comments.

berkerpeksag:
Looks great, thanks! I think you forgot to address two of my previous comments:

* We need to cleanup the env variables we set in tests. We can do this in the following way:

  ```py
  def test_cli_overrides_enviroment_variables_module(monkeypatch):
      monkeypatch.setenv("GUNICORN_CMD_ARGS", "--workers=4")
      # test code
  ```
* `GUNICORN_CMD_ARGS` needs to be documented. We can mention it in `docs/source/run.rst` and `docs/gunicorn_ext.py`. We can leave this to a separate PR if you want.

tests/test_config.py
```diff
@@ -285,3 +285,21 @@ def test_always_use_configured_logger():
...
+    os.environ["GUNICORN_CMD_ARGS"] = "--workers=4"
+    with AltArgs():
+        app = NoConfigApp()
+        assert app.cfg.workers == 4
+
+def test_invalid_enviroment_variables_config():
+    os.environ["GUNICORN_CMD_ARGS"] = "--foo=bar"
+    with AltArgs():
+        with pytest.raises(SystemExit):
+            app = NoConfigApp()
```
berkerpeksag: Style nit: No need to set `NoConfigApp()` to `app`. Just call `NoConfigApp()`.

tests/test_config.py
```diff
@@ -285,3 +285,21 @@ def test_always_use_configured_logger():
     c.set('statsd_host', 'localhost:12345')
     # still uses custom logger over statsd
     assert c.logger_class == MyLogger
+
```
berkerpeksag: Style nit: Please add two empty lines after each tests.

hramezani:
@berkerpeksag thanks!, comments fixed.
can you help me for documentation? is this ok to add a note or paragraph in `Settings` and ``

berkerpeksag:
Thanks for the PR, @hramezani! :)

> is this ok to add a note or paragraph in Settings and Running Gunicorn page, and describe that we can use all command line config in env vars in bellow format?
> GUNICORN_CMD_ARGS ="--bind=127.0.0.1 --workers=3"

Yes, that sounds good to me. Let's open another pull request for documentation.

hramezani:
@berkerpeksag  thanks very much for your help. i open another pull request for documentation soon.

PR #1387, opened 2016-11-08
threadWorker cancelled request should decrease nr_conns

zhipcui:


benoitc:
@tilgovi why not, let's do this in 2 pass though ? :)

PR #1394, opened 2016-11-18
Add child_exit() callback to configuration

jonashaag:
This adds a `child_exit()` callback to the Gunicorn configuration that works similar to `worker_exit()` but is called in the *master* process instead of the *child process*.

I also added some basic unit tests for the both.

A change to how `Worker.pid` is implemented was necessary in order for it to be meaningful in the master process (it would always be the PID of the master process, no information given about the worker process).

Our need for this is prometheus/client_python#115

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -498,8 +498,9 @@ class Arbiter(object):
                 if self.reexec_pid == wpid:
                     self.reexec_pid = 0
                 else:
-                    # A worker said it cannot boot. We'll shutdown
-                    # to avoid infinite start/stop cycles.
+                    # A worker was terminated. If the termination reason was
+                    # that it could not boot, we'll shut it down to avoid
+                    # infinite start/stop cycles.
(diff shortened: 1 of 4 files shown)


benoitc:
I will look at the patch. For your case why don't you clean the data when the worker is spawned though?

jonashaag:
> For your case why don't you clean the data when the worker is spawned though?

We'll need the old worker's PID to do that

jonashaag:
Rebased to latest master

berkerpeksag:
Looks good to me too. It would be nice to add a `versionadded` directive. @benoitc and @tilgovi should it be 19.7 or 20?

(We also need to use the "squash and merge" button when we merge this.)

jonashaag:
> (We also need to use the "squash and merge" button when we merge this.)

I deliberately split the changes into three commits to make them easier to understand

jonashaag:
Hey, what's the state of this? Is there anything I can do to have this merged?

PR #1395, opened 2016-11-18
Fix Python 3.5 on Travis typo

jonashaag:


jonashaag:
Somehow PyPy 3 is still broken but at least it fixes Python 3.5

berkerpeksag:
Good catch, thanks! I think we should just remove pypy3 from .travis.yml at this point.

PR #1396, opened 2016-11-18
Fix argument mistake

t10471:
datetime.timedelta([days[, seconds[, microseconds[, milliseconds[, minutes[, hours[, weeks]]]]]]])

### DIFF
--- a/gunicorn/workers/_gaiohttp.py
+++ b/gunicorn/workers/_gaiohttp.py
@@ -21,7 +21,7 @@ from aiohttp.wsgi import WSGIServerHttpProtocol as OldWSGIServerHttpProtocol
 
 class WSGIServerHttpProtocol(OldWSGIServerHttpProtocol):
     def log_access(self, request, environ, response, time):
-        self.logger.access(response, request, environ, datetime.timedelta(0, 0, time))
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Good catch, but the question is why we need a timedelta instance at all? It looks like aiohttp uses `self._loop.time() - now` all the time. I never used this worker so I might be wrong here.

t10471:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/sync.py#L184

Other workers use timedelta, so this worker convert float to timedelta.

gunicorn/workers/_gaiohttp.py
```diff
@@ -21,7 +21,7 @@
 
 class WSGIServerHttpProtocol(OldWSGIServerHttpProtocol):
     def log_access(self, request, environ, response, time):
-        self.logger.access(response, request, environ, datetime.timedelta(0, 0, time))
+        self.logger.access(response, request, environ, datetime.timedelta(0, time))
```
berkerpeksag: Wouldn't `datetime.timedelta(seconds=time)` be more readable here?

Issue #1210, opened 2016-02-16
'Response' object has no attribute 'status_code' in wsgi.py with websockets


bclark8923:
I start my app using

`gunicorn --worker-class eventlet -w 1 server:app --bind="127.0.0.1:5000"`

And when using Flack-SocketIO for websockets I often get an error with Gunicorn not returning from a function properly

```
[2016-01-30 10:20:53 -0800] [7330] [ERROR] Error handling request
Traceback (most recent call last):
  File "/Users/brianclark/Desktop/Projects/HDP/application/api/venv/lib/python2.7/site-packages/gunicorn/workers/async.py", line 52, in handle
    self.handle_request(listener_name, req, client, addr)
  File "/Users/brianclark/Desktop/Projects/HDP/application/api/venv/lib/python2.7/site-packages/gunicorn/workers/async.py", line 114, in handle_request
    resp.close()
  File "/Users/brianclark/Desktop/Projects/HDP/application/api/venv/lib/python2.7/site-packages/gunicorn/http/wsgi.py", line 423, in close
    self.send_headers()
  File "/Users/brianclark/Desktop/Projects/HDP/application/api/venv/lib/python2.7/site-packages/gunicorn/http/wsgi.py", line 316, in send_headers
    tosend = self.default_headers()
  File "/Users/brianclark/Desktop/Projects/HDP/application/api/venv/lib/python2.7/site-packages/gunicorn/http/wsgi.py", line 297, in default_headers
    elif self.should_close():
  File "/Users/brianclark/Desktop/Projects/HDP/application/api/venv/lib/python2.7/site-packages/gunicorn/http/wsgi.py", line 230, in should_close
    if self.status_code < 200 or self.status_code in (204, 304):
AttributeError: 'Response' object has no attribute 'status_code'
```

Any advice on configurations I could use to fix this or is it a bug?

benoitc:
@bclark8923 do you have any application that could help to reproduce the issue?

bclark8923:
Yup! What would you like me to do?

On Monday, March 28, 2016, Benoit Chesneau <notifications@github.com> wrote:

> @bclark8923 <https://github.com/bclark8923> do you have any application
> that could help to reproduce the issue?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/benoitc/gunicorn/issues/1210#issuecomment-202484667>
>


-- 
Thanks,
Brian Clark
(248) 990 5616
[...]

PR #1406, opened 2016-12-06
add support for eventlets _AlreadyHandled object

stefaang:
fixes https://github.com/benoitc/gunicorn/issues/1210

### DIFF
--- a/gunicorn/workers/async.py
+++ b/gunicorn/workers/async.py
@@ -27,6 +27,10 @@ class AsyncWorker(base.Worker):
     def timeout_ctx(self):
         raise NotImplementedError()
 
+    def is_already_handled(self, respiter):
+        # some workers will need to overload this function to raise a StopIteration
+        return respiter == ALREADY_HANDLED
+
     def handle(self, listener, client, addr):
         req = None
         try:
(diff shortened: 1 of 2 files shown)


Hum4n01d:
Does this just silence the error?

benoitc:
thanks for the patch.

but such changes must be done in the eventlet worker code since it's
specific to it.

What about catching an optional value set by the worker?

btw Wondering why the lib can't use the gunicorn way that provide a
consistent behaviour across worker and doesn't force the usage of eventlet
though...has it been discussed?

stefaang:
@Hum4n01d The error was caused by an empty wsgi response on disconnect, due to a slightly different implementation of the _AlreadyHandled object. The goal of _AlreadyHandled is to raise a StopIteration exception to close the socket.
@benoitc I reworked this into a wrapper around self.wsgi by overloading self.load_wsgi in the geventlet worker. 
I'm not that familiar with the all the eventing projects... but this error was messing up my logs so I wanted a quick fix.

Hum4n01d:
I don't totally get what you're saying but please merge it @benoitc

benoitc:
@stefaang the change can be simpler imo,

so we check `ALREADY_HANDLED` there:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/async.py#L104

Instead we could check it using a function `is_already_handled`  that could be overriden by the worker. In the case of Eventlet it would check for both the gunicorn value and the eventlet value. Do you want to make the change?

stefaang:
You're right, that's a better approach. I'll get around with another fix.

Hum4n01d:
So can it be merged?

tilgovi:
This looks good to me.

benoitc:
merged. Thanks!

Hum4n01d:
Yay! Now my logs won't be clogged up

qwexvf:
Thank you for this fix <3

PR #1410, opened 2016-12-16
Add hook just called after master received a signal

MrKiven:
It's useful if fork in main process

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -202,6 +202,8 @@ class Arbiter(object):
                     self.log.info("Ignoring unknown signal: %s", sig)
                     continue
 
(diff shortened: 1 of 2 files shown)


benoitc:
what is the use case?

MrKiven:
For example:
```
class MyApp(Application):
    ...

    def run():
        fork_a_process()  # may do something background
       super(Myapp, self).run()
```

I want deliver signals(TERM/INT/QUIT or any other signals) to `a` process so that it can be graceful  exit or just do something when master received..

tilgovi:
For the use case described, you can use the `on_exit` hook.

Please re-open if you think that does not address your use case.

Issue #1091, opened 2015-07-31
gunicorn.pidfile.validate crashes gunicorn when PID exists but is from a different user


ski-csis:
```
Traceback (most recent call last):
  File "/opt/python2.7/bin/gunicorn", line 11, in <module>
    sys.exit(run())
  File "/opt/python2.7/lib/python2.7/site-packages/gunicorn/app/wsgiapp.py", line 74, in run
    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
  File "/opt/python2.7/lib/python2.7/site-packages/gunicorn/app/base.py", line 189, in run
    super(Application, self).run()
  File "/opt/python2.7/lib/python2.7/site-packages/gunicorn/app/base.py", line 72, in run
    Arbiter(self).run()
  File "/opt/python2.7/lib/python2.7/site-packages/gunicorn/arbiter.py", line 171, in run
    self.start()
  File "/opt/python2.7/lib/python2.7/site-packages/gunicorn/arbiter.py", line 125, in start
    self.pidfile.create(self.pid)
  File "/opt/python2.7/lib/python2.7/site-packages/gunicorn/pidfile.py", line 23, in create
    oldpid = self.validate()
  File "/opt/python2.7/lib/python2.7/site-packages/gunicorn/pidfile.py", line 75, in validate
    os.kill(wpid, 0)
OSError: [Errno 1] Operation not permitted
```

This happens because the process identified by the pid-file exists, but belongs to a different user than the one starting gunicorn.

(This is with gunicorn 19.3.0)

benoitc:
how did you start gunicorn? Is the process still associated to the arbiter? does the user starting gunicorn have the rights to kill its children?

ski-csis:
We run gunicorn with supervisord, nothing really fancy.

I believe what happened was that the server was rebooted, and a few gunicorns did not shutdown in time, so when the server came up again their old pid-files were still there, but now the PIDs belonged to gunicorns from other users (the server runs a bunch of projects as different users.)

benoitc:
@ski-csis but each pidfile is created per projetcs right ? So if the same suer is launching the project each time it shouldn't happen. Can you share your supervisor conf?

ski-csis:
There is a pid-file per gunicorn, yes. That's not the problem.

The problem is that the pid in the pid-file no longer belongs to the user running the gunicorn, because gunicorn was not shutdown cleanly before a reboot.

[...]

PR #1414, opened 2016-12-22
Fix/1091 pidfile eperm

tilgovi:
Tests (including failing test) for `Pidfile.validate` and the fix for #1091 that makes the test pass in a subsequent commit.

### DIFF
--- a/tests/test_pidfile.py
+++ b/tests/test_pidfile.py
@@ -1,0 +1,59 @@
+# -*- coding: utf-8 -
+#
+# This file is part of gunicorn released under the MIT license.
+# See the NOTICE for more information.
+
+import errno
+import sys
+
+try:
+    import unittest.mock as mock
+except ImportError:
+    import mock
+
+import gunicorn.pidfile
+
+
+def builtin(name):
+    if sys.version_info >= (3, 0):
+        module = 'builtins'
+    else:
+        module = '__builtin__'
+
+    return '{0}.{1}'.format(module, name)
+
+
+@mock.patch(builtin('open'), new_callable=mock.mock_open)
+def test_validate_no_file(_open):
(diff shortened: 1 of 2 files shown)


tilgovi:
I'll have to fix this for py3

tests/test_pidfile.py
```diff
@@ -0,0 +1,83 @@
...
+
+try:
+    import unittest.mock as mock
+except ImportError:
+    import mock
+
+import gunicorn.pidfile
+
+
+def mock_patch_builtin(name):
```
berkerpeksag: I haven't tried this idea, but perhaps using `mock_open()` would make the test code simpler.

tilgovi: It would, but I didn't want to introduce a requirement on a newer version of mock. I thought it was you who once had me change `assert_not_called` for this reason.

tilgovi: On the other hand, I'm happy to introduce a version requirement for mock.

tilgovi: Just let me know which you would prefer.

berkerpeksag: I thought we already use the latest version available of mock for Python 2. `mock_open()` has been in the stdlib version since Python 3.3 and in the PyPI version since 2012: https://github.com/testing-cabal/mock/commit/afc98d4e4eab7d66b09483fb3da1ac30e766d026 So unless I'm missing something, we are good to go without introducing a version requirement :)

benoitc: let just use the latest mock version. There is no need to be conservative for tests imho :)

tilgovi:
Changed to use `mock_open`. It's a bit cleaner now.

PR #1418, opened 2016-12-27
use GunicornWebWorker from aiohttp if available

fafhrd91:
we are planing to deprecate `aiohttp.wsgi` module, so i'd like to deprecate AiohttpWorker as well.
`GunicornWebWorker` supports both `aiohttp.web.Application` and `aiohttp.swgi` applications

we are planing to remove `aiohttp.wsgi` in 9-12 months.

### DIFF
--- a/gunicorn/workers/gaiohttp.py
+++ b/gunicorn/workers/gaiohttp.py
@@ -4,14 +4,23 @@
 # See the NOTICE for more information.
 
 import sys
+from gunicorn import utils
 
-if sys.version_info >= (3, 3):
+if sys.version_info >= (3, 4):
     try:
         import aiohttp  # NOQA
     except ImportError:
         raise RuntimeError("You need aiohttp installed to use this worker.")
     else:
-        from gunicorn.workers._gaiohttp import AiohttpWorker
+        try:
+            from aiohttp.worker import GunicornWebWorker as AiohttpWorker
+        except ImportError:
+            from gunicorn.workers._gaiohttp import AiohttpWorker
+
+            utils.warn(
+                "AiohttpWorker is deprecated please install aiohttp 1.2+ "
+                "and set aiohttp.worker.GunicornWebWorker as a custom worker ")
+
         __all__ = ['AiohttpWorker']
 else:
-    raise RuntimeError("You need Python >= 3.3 to use the asyncio worker")
+    raise RuntimeError("You need Python >= 3.4 to use the asyncio worker")
--- a/tests/test_gaiohttp.py
+++ b/tests/test_gaiohttp.py
@@ -6,9 +6,8 @@
 import unittest
 import pytest
 aiohttp = pytest.importorskip("aiohttp")
-
-
-from aiohttp.wsgi import WSGIServerHttpProtocol
+WSGIServerHttpProtocol = pytest.importorskip(
+    "aiohttp.wsgi.WSGIServerHttpProtocol")
 
 import asyncio
 from gunicorn.workers import gaiohttp


gunicorn/workers/gaiohttp.py
```diff
@@ -4,14 +4,25 @@
...
-        from gunicorn.workers._gaiohttp import AiohttpWorker
-        __all__ = ['AiohttpWorker']
+        try:
+            from aiohttp.worker import WSGIServer  # NOQA
+            from aiohttp.worker import GunicornWebWorker as AiohttpWorker
+        except ImportError:
+            from gunicorn.workers._gaiohttp import AiohttpWorker
+
+            warnings.warn(
+                "AiohttpWorker is deprecarted please install "
```
tilgovi: "deprecated"

fafhrd91: fixed

gunicorn/workers/gaiohttp.py
```diff
@@ -4,14 +4,25 @@
...
+            from aiohttp.worker import GunicornWebWorker as AiohttpWorker
+        except ImportError:
+            from gunicorn.workers._gaiohttp import AiohttpWorker
+
+            warnings.warn(
+                "AiohttpWorker is deprecated please install "
+                "aiohttp >= 1.2.0 ",
+                DeprecationWarning)
+
+        __all__ = ('AiohttpWorker',)
```
berkerpeksag: If I remember correctly, pydoc doesn't like `tuple`s. It might be better to keep `__all__` a list.

gunicorn/workers/gaiohttp.py
```diff
@@ -4,14 +4,25 @@
...
+        try:
+            from aiohttp.worker import WSGIServer  # NOQA
+            from aiohttp.worker import GunicornWebWorker as AiohttpWorker
+        except ImportError:
+            from gunicorn.workers._gaiohttp import AiohttpWorker
+
+            warnings.warn(
+                "AiohttpWorker is deprecated please install "
+                "aiohttp >= 1.2.0 ",
+                DeprecationWarning)
```
berkerpeksag: I think we do use https://github.com/benoitc/gunicorn/blob/master/gunicorn/util.py#L550 for this kid of use cases. With, `util.warn`, people can see that the worker has been deprecated without passing `-Wd` or `-Wall` to `python`.

gunicorn/workers/gaiohttp.py
```diff
@@ -4,14 +4,25 @@
...
-        __all__ = ['AiohttpWorker']
+        try:
+            from aiohttp.worker import WSGIServer  # NOQA
+            from aiohttp.worker import GunicornWebWorker as AiohttpWorker
+        except ImportError:
+            from gunicorn.workers._gaiohttp import AiohttpWorker
+
+            warnings.warn(
+                "AiohttpWorker is deprecated please install "
+                "aiohttp >= 1.2.0 ",
```
berkerpeksag: Perhaps we should also say something like "install aiohttp 1.2.0+ and set aiohttp.worker.GunicornWebWorker as a custom worker"?

benoitc:
bump?

fafhrd91:
@benoitc do you want me to deprecate worker or replace it with new one?

fafhrd91:
Updated.

fafhrd91:
I can replace AiohttpWorker with GunicornWebWorker as well.

gunicorn/workers/gaiohttp.py
```diff
@@ -4,14 +4,23 @@
...
     except ImportError:
         raise RuntimeError("You need aiohttp installed to use this worker.")
     else:
-        from gunicorn.workers._gaiohttp import AiohttpWorker
+        try:
+            from aiohttp.worker import GunicornWebWorker as AiohttpWorker
+        except ImportError:
+            from gunicorn.workers._gaiohttp import AiohttpWorker
+
+            utils.warn(
```
berkerpeksag:
```py
utils.warn(
    "AiohttpWorker is deprecated. Please install aiohttp 1.2+ "
    "and set aiohttp.worker.GunicornWebWorker as a custom worker."
)
```

fafhrd91:
I asked question regarding deprecation month ago, you guys should decide. right now gunicorn does not work with aiohttp 2.x at all.

benoitc:
@fafhrd91 is there any eplacement to support asgi/wsgi? 

since you have the commit bit you could likely replace the current implementation by a new one btw ;)

benoitc:
I would like to understand what it means as it is not clear at all? is the AiohttpWorker inside  can still work or not with the latest version of aiohttp? If not why the worker hasn't been updated there?

fafhrd91:
AiohttpWorker does not work with aiohttp2.0. solution was siggested 5 months ago.

benoitc:
well my question was , why the worker hasn't been updated then but ok, then we will rather deprecate the current one.

bart3005:
Is there a time frame for aiohttp2.0 support?

tilgovi:
I think we just drop it for R20 and don't worry about deprecating, but we need to update the docs.

bart3005:
So is there any plan to support aiohttp? In my real world use case with a Flask app, Gunicorn with the gaiohttp worker is 3rd faster than the other fastest worker.

Just wondering what I should do.

tilgovi:
@bart3005 you can still set `-k aiohttp.worker.GunicornWebWorker` if you install aiohttp. My understanding is that this is just about removing the worker from being bundled with Gunicorn by default.

bart3005:
@tilgovi Hmmm OK does not seem to work, with latest of both:

RuntimeError: aiohttp.wsgi is not supported anymore, consider to switch to aiohttp.web.Application

Works OK with aiohttp (1.3.5) but not with latest aiohttp.

berkerpeksag:
I think what they meant with that exception message is that you should use the `web.Application` API instead of a plain WSGI app. Here's a minimal working example with using aiohttp 2.2.5 and Gunicorn:

```py
import asyncio

from aiohttp import web


# I have Python 3.4 on my current system so I didn't use the 'async' keyword.
@asyncio.coroutine
def handle(request):
    name = request.match_info.get("name", "Anonymous")
    text = "Hello, " + name
    return web.Response(text=text)

app = web.Application()
app.router.add_get('/', handle)
app.router.add_get('/{name}', handle)
```

And run it:

```sh
$ gunicorn foo:app  -b 127.0.0.1:8585 -k aiohttp.worker.GunicornWebWorker
[2017-08-07 15:28:13 +0300] [18979] [INFO] Starting gunicorn 19.7.1
[2017-08-07 15:28:13 +0300] [18979] [INFO] Listening at: http://127.0.0.1:8585 (18979)
[2017-08-07 15:28:13 +0300] [18979] [INFO] Using worker: aiohttp.worker.GunicornWebWorker
[2017-08-07 15:28:13 +0300] [18984] [INFO] Booting worker with pid: 18984
```

And test it:

```sh
$ curl http://127.0.0.1:8585
Hello, Anonymous
```

And I got the same exception when I used [`examples/test.py`](https://github.com/benoitc/gunicorn/blob/master/examples/test.py) (tweaked a bit) with aiohttp 2.2.5:

```sh
cd examples/
$ gunicorn test:app  -b 127.0.0.1:8585 -k aiohttp.worker.GunicornWebWorker
...
[2017-08-07 15:22:27 +0300] [18852] [INFO] Worker exiting (pid: 18852)
[2017-08-07 15:22:28 +0300] [18853] [INFO] Booting worker with pid: 18853
[2017-08-07 15:22:28 +0300] [18853] [ERROR] Exception in worker process
  File "/home/berker/projects/gunicorn/gunicorn/arbiter.py", line 579, in spawn_worker
    worker.init_process()
  File "/home/berker/projects/gunicorn/venv/lib/python3.4/site-packages/aiohttp/worker.py", line 41, in init_process
    super().init_process()
  File "/home/berker/projects/gunicorn/gunicorn/workers/base.py", line 133, in init_process
    self.run()
  File "/home/berker/projects/gunicorn/venv/lib/python3.4/site-packages/aiohttp/worker.py", line 49, in run
    self.loop.run_until_complete(self._runner)
  File "/usr/local/lib/python3.4/asyncio/base_events.py", line 341, in run_until_complete
    return future.result()
  File "/usr/local/lib/python3.4/asyncio/futures.py", line 276, in result
    raise self._exception
  File "/usr/local/lib/python3.4/asyncio/tasks.py", line 243, in _step
    result = coro.send(None)
  File "/home/berker/projects/gunicorn/venv/lib/python3.4/site-packages/aiohttp/worker.py", line 105, in _run
    handler = self.make_handler(self.wsgi)
  File "/home/berker/projects/gunicorn/venv/lib/python3.4/site-packages/aiohttp/worker.py", line 68, in make_handler
    "aiohttp.wsgi is not supported anymore, "
RuntimeError: aiohttp.wsgi is not supported anymore, consider to switch to aiohttp.web.Application
```

berkerpeksag:
> I think what they meant with that exception message is that you should use the web.Application API instead of a plain WSGI app.

To be clear, a Flask app is also a plain WSGI app so you'll either stay with aiohttp 1 or switch your application to use aiohttp's `web.Application` API.

berkerpeksag:
Turns out, there is a third-party project called `aiohttp_wsgi` which lets you wrap your WSGI application with `aiohttp_wsgi.WSGIHandler` and pass it into `aiohttp.web.Application`: https://aiohttp-wsgi.readthedocs.io/en/stable/wsgi.html#run-a-simple-web-server

I'm going to open a new PR to document the current situation.

bart3005:
@berkerpeksag thanks so much for this. Works like a charm and performance is very good.

in the __init__.py of the flask app:

from aiohttp import web
from aiohttp_wsgi import WSGIHandler

app = Flask(__name__)
wsgi_handler = WSGIHandler(app)
aioapp = web.Application()
aioapp.router.add_route("*", "/{path_info:.*}", wsgi_handler)

And gunicorn:

gunicorn app:aioapp -b 127.0.0.1:9000 \
  --worker-class aiohttp.worker.GunicornWebWorker

asvetlov:
@bart3005 aiohttp doesn't support `aiohttp_wsgi` but I hope next aiohttp releases will not break the library (or author will publish bugfix version)

bart3005:
@asvetlov understood.

Fover23:
@bart3005
You can try aiohttp 1.3.5 with uvloop
is very fast and work well with my django project 

```
pip install aiohttp==1.3.5
pip install uvloop
```
and gunicorn:

```
gunicorn -k aiohttp.worker.GunicornUVLoopWebWorker -w 1 yourproject.wsgi:application -b 127.0.0.1:8000
```

berkerpeksag:
Superseded by https://github.com/benoitc/gunicorn/pull/1569. I've added an example app, document the current situation, and deprecated the gaiohttp worker there.

PR #1428, opened 2017-01-11
Update arbiter.py to not leave tmp files around (fix #1327)

### LINKED ISSUE #1327: (deleted) tmp files left around
Env:
gunicorn==19.6.0
OS = CentOS
Running under systemd.

A bunch of temp files are still owned by my gunicorn worker processes. The number of temp files is varying across workers, that worker has 45 left around, another one has 67, some seems to be shared. The naming of those temp files seems to be mean that they are used/owned by gunicorn. I ran into resource problems (max number of opened files > 1024) and one of our Ops guy noticed that gunicorn had many opened files.

Thanks for any hints on debugging this !!

I start/stop the service through systemd, and I have not configured how systemd should stop and reload the app. My upgrade process make it stop and start. For the record I tried to implement custom restart by sending USR2, then WINCH and TERM to the old process, but systemd was ignoring that, detecting it as non handled workflow and would restart gunicorn on its own.

```
gunicorn: 3378 crashzone    7u      REG                8,3         0 269227083 /tmp/wgunicorn-WUef6V (deleted)
gunicorn: 3378 crashzone    8u      REG                8,3         0 269227084 /tmp/wgunicorn-z78nQi (deleted)
gunicorn: 3378 crashzone    9u      REG                8,3         0 269227085 /tmp/wgunicorn-yBPZFy (deleted)
gunicorn: 3378 crashzone   10u      REG                8,3         0 269227116 /tmp/wgunicorn-IZ8yLv (deleted)
gunicorn: 3378 crashzone   11u      REG                8,3         0 269227118 /tmp/wgunicorn-b7RadI (deleted)
gunicorn: 3378 crashzone   12u      REG                8,3         0 269231113 /tmp/wgunicorn-QTigaF (deleted)
gunicorn: 3378 crashzone   13u      REG                8,3         0 269231114 /tmp/wgunicorn-2Vthud (deleted)
gunicorn: 3378 crashzone   14u      REG                8,3         0 269231116 /tmp/wgunicorn-8CHx_z (deleted)
gunicorn: 3378 crashzone   15u      REG                8,3         0 269231144 /tmp/wgunicorn-OIXCeX (deleted)
gunicorn: 3378 crashzone   16u      REG                8,3         0 269231145 /tmp/wgunicorn-saa7Vq (deleted)
gunicorn: 3378 crashzone   17u      REG                8,3         0 269231146 /tmp/wgunicorn-jX9a4S (deleted)
gunicorn: 3378 crashzone   18u      REG                8,3         0 269231154 /tmp/wgunicorn-Dm1KAE (deleted)
gunicorn: 3378 crashzone   19u      REG                8,3         0 269231155 /tmp/wgunicorn-RBV92D (deleted)
gunicorn: 3378 crashzone   20u      REG                8,3         0 269231157 /tmp/wgunicorn-VUD025 (deleted)
gunicorn: 3378 crashzone   21u      REG                8,3         0 269231158 /tmp/wgunicorn-VxoyId (deleted)
gunicorn: 3378 crashzone   22u      REG                8,3         0 269231159 /tmp/wgunicorn-h65JSk (deleted)
gunicorn: 3378 crashzone   23u      REG                8,3         0 269231160 /tmp/wgunicorn-yPNANx (deleted)
gunicorn: 3378 crashzone   24u      REG                8,3         0 269231161 /tmp/wgunicorn-P0YYso (deleted)
gunicorn: 3378 crashzone   25u      REG                8,3         0 269231162 /tmp/wgunicorn-ykwUxS (deleted)
gunicorn: 3378 crashzone   26u      REG                8,3         0 269231163 /tmp/wgunicorn-GhL7Yl (deleted)
gunicorn: 3378 crashzone   27u      REG                8,3         0 269231164 /tmp/wgunicorn-gV2KLt (deleted)
```

I have a wrapper for running gunicorn, which I ran entirely through python:

```
'''Start the gunicorn HTTP server
'''

import gunicorn.app.base
from gunicorn.six import iteritems


class StandaloneApplication(gunicorn.app.base.BaseApplication):
    '''A gunicorn application'''

    def __init__(self, app, options=None):
        '''Construct a gunicorn standalone app'''

        self.options = options or {}
        self.application = app
        super(StandaloneApplication, self).__init__()

    def init(self, parser, opts, args):
        '''init the Gunicorn app (the base class method is not implemented)'''

        super(StandaloneApplication, self).init(parser, opts, args)

    def load_config(self):
        '''Load our config file'''

        config = dict([(key, value) for key, value in iteritems(self.options)
                       if key in self.cfg.settings and value is not None])
        for key, value in iteritems(config):
            self.cfg.set(key.lower(), value)

    def load(self):
        '''Load our application'''

        return self.application
```

I run the application like that:

```
StandaloneApplication(app, options).run()
```

bsergean:
(deleted) tmp files left around #1327

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -565,6 +565,10 @@ class Arbiter(object):
             self.WORKERS[pid] = worker
             return pid
 
+        # Do not inherit the temporary files of other workers
+        for sibling in self.WORKERS.values():
+            sibling.tmp.close()
+        
         # Process Child
         worker_pid = os.getpid()
         try:


tilgovi:
This patch helped us diagnose the problem, but I don't think it is the best solution.

It would be less costly to set the close on exec flag for the temp file after it's inherited by its worker, right after the `fork`. This way, we never have to worry about it again and it won't be inherited by any other process that might get spawned for any reason.

bsergean:
Found those references:

http://stackoverflow.com/questions/6125068/what-does-the-fd-cloexec-fcntl-flag-do

```
flags = fcntl.fcntl(fd, fcntl.F_GETFD)
fcntl.fcntl(fd, fcntl.F_SETFD, flags | fcntl.FD_CLOEXEC)
```

bsergean:
Oh it looks like there's already a utility for that: close_on_exec

bsergean:
There's a worker.tmp in the code, is it that temp file we're talking about ?

tilgovi:
I am wrong, `close_on_exec` won't work because the workers don't exec.

tilgovi:
@benoitc @berkerpeksag I'd like to merge this, but would appreciate one other review. I don't see a better way right now and I've been putting this off for weeks.

gunicorn/arbiter.py
```diff
@@ -565,6 +565,10 @@ def spawn_worker(self):
             self.WORKERS[pid] = worker
             return pid
 
+        # Do not inherit the temporary files of other workers
+        for sibling in self.WORKERS.values():
+            sibling.tmp.close()
```
berkerpeksag: Perhaps a more elegant way would be:

```py
import os
import resource

fd_high = resource.getrlimit(resource.RLIMIT_NOFILE)[0]
os.closerange(3, fd_high)
```

Or we could store FDs of temp files of workers in a list and pass the highest number as `fd_high`.

benoitc:
I'm confused we already set close on exec there : https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L106

and we already unlink it there:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/workertmp.py#L32

Normally these files shouldn't be let on the fs then. So the actual question for me is on which system is it reproducible? How do you launch gunicorn ?

Anyway why not using `os.set_inheritable(fd, inheritable)` when creating a temporary file on python 3 with `inheritable=false`? Wouldn't it solve the issue?

gunicorn/arbiter.py
tilgovi: I'm not sure it would be safe to assume that all open FDs are `WorkerTmp` files. We could have any number of log files open at this point, no?

tilgovi:
@benoitc:

> I'm confused we already set close on exec

The workers don't call `exec`.

> and we already unlink it there

We don't close it, though.

> Normally these files shouldn't be let on the fs then. So the actual question for me is on which system is it reproducible? How do you launch gunicorn ?

As far as I understand, this is not OS-specific and it doesn't matter how you launch Gunicorn.

> Anyway why not using os.set_inheritable(fd, inheritable) when creating a temporary file on python 3 with inheritable=false? Wouldn't it solve the issue?

We need the file to be inheritable. The arbiter unlinks the file, so passing the fd through the fork is the only way the worker can access it.

Also, per the documentation [1], "On UNIX, non-inheritable file descriptors are closed in child processes at the execution of a new program, other file descriptors are inherited." The workers do not call `exec`.

[1] https://docs.python.org/3/library/os.html#inheritance-of-file-descriptors

gunicorn/arbiter.py
berkerpeksag: You have a point :) So unless we keep tracking file descriptors of those `WorkerTmp` files I don't see a better way to solve this problem either.

benoitc:
That's true. We want them inheritable, like i said i'm confused by that ticket :). I don't reproduce the issue there. I'm not sure how you can still have tmp file on disk if they are unlinked. So I guess it's OS specific, or maybe just another systemd crap as it's apparently used there. I would be interrested in knowing more what does systemd or whatever. Anyway the patch is OK, let's merge it.

tilgovi:
> I don't reproduce the issue there. I'm not sure how you can still have tmp file on disk if they are unlinked.

They are not in the filesystem, but in the list of open files for the worker processes. You have to call `lsof` on a worker to see that it will have some other workers tmp files open, too.

berkerpeksag:
Note that I think we can use [`os.register_at_fork()`](https://docs.python.org/3.7/library/os.html#os.register_at_fork) under Python 3.7+. Unlike the close-on-exec flag, this doesn't require the child to call `exec*` functions to work.

benoitc:
Still i was not reproducing it when the process is restarted using HUP or USR2 in my systems so something is triggering it.... I guess a better fix would be finally replacing that temporary file trick :) I need to focus on that.

@berkerpeksag you mean to have a function cleaning on exit?

berkerpeksag:
> @berkerpeksag you mean to have a function cleaning on fork?

Yes, `os.register_at_fork()` may help us to close temp files.

boazin:
Is this fix merged into a release? When is it expected in an official release?

PR #1430, opened 2017-01-13
Use the best available reloader when --reload is passed

### LINKED ISSUE #1422: --reload flag changed semantics
I just upgraded from the 19.6.0 release to the latest commit c54426c1e4b26b30c5feee2cec4450656ae147a3

I previously ran gunicorn like `gunicorn --reload mymodule:app` during development. However now I get the following error:
```
$ gunicorn --reload mymodule:app
usage: gunicorn [OPTIONS] [APP_MODULE]
gunicorn: error: No application module specified.
```
If I put an option after the reload flag it works.
```
$ gunicorn --reload -w=1 mymodule:app
[2017-01-05 11:02:33 +0000] [102] [INFO] Starting gunicorn 19.6.0
[2017-01-05 11:02:33 +0000] [102] [INFO] Listening at:  (102)
[2017-01-05 11:02:33 +0000] [102] [INFO] Using worker: sync
[2017-01-05 11:02:33 +0000] [105] [INFO] Booting worker with pid: 105
```

Calling it like `gunicorn --reload auto mymodule:app` also works, but then the argument isn't really optional. Perhaps the reload type should be non-optional to avoid confusion?

berkerpeksag:
Fixes #1422

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -266,8 +266,8 @@ Debugging
 reload
 ~~~~~~
 
-* ``--reload RELOADER_TYPE``
-* ``off``
+* ``--reload``
+* ``False``
 
 Restart workers when code changes.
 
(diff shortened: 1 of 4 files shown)


mark-adams:
The biggest problem that I see with this change is that the `preferred_reloader` may not always be desirable. For instance, inotify may be supported by the OS but not by the underlying filesystem. In that case, the library would try to use inotify but would never receive notification of changes because the filesystem driver doesn't support inotify. Currently, the user could simply force polling via the `--reload` argument and they'd be good to go. With this change, they'd be out of luck and the reloader wouldn't work.

PR #1431, opened 2017-01-14
Document limitations of async workers and preload_app

### LINKED ISSUE #1056: monkey.patch_all and gunicorn with more than 1 worker

berkerpeksag:
Fixes #1056

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -418,8 +418,9 @@ restarting workers.
 
 .. note::
    Using this setting with the ``'gevent'`` and ``'eventlet'`` workers
-   is discouraged. When ``--reload`` is used, these workers will try to
-   monkeypatch the standard library *after* the application is loaded.
+   is discouraged. When ``--preload`` is used, these workers will try
+   to monkeypatch the standard library *after* the application is
+   loaded.
 
 .. _sendfile:
 
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -125,6 +125,14 @@ A string referring to one of the following bundled classes:
 * ``gthread``  - Python 2 requires the futures package to be installed
 * ``gaiohttp`` - Deprecated.
 
+.. note::
+   If you want to use ``'gevent'`` and ``'eventlet'`` workers, avoid
+   monkeypatching (e.g. calling the ``gevent.monkey.patch_all()`` and
+   ``eventlet.monkey_patch()`` functions respectively) in your
+   application. Gunicorn will do it for you after the ``on_starting``
+   hook ran. See also the documentation for the :ref:`preload-app`
+   setting.
+
 Optionally, you can provide your own worker by giving Gunicorn a
 Python path to a subclass of ``gunicorn.workers.base.Worker``.
 This alternative syntax will load the gevent class:
@@ -407,6 +415,11 @@ By preloading an application you can save some RAM resources as well as
 speed up server boot times. Although, if you defer application loading
 to each worker process, you can reload your application code easily by
 restarting workers.
+
+.. note::
+   Using this setting with the ``'gevent'`` and ``'eventlet'`` workers
+   is discouraged. When ``--reload`` is used, these workers will try to
+   monkeypatch the standard library *after* the application is loaded.
 
 .. _sendfile:
 
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -939,8 +939,9 @@ class PreloadApp(Setting):
 
         .. note::
            Using this setting with the ``'gevent'`` and ``'eventlet'`` workers
-           is discouraged. When ``--reload`` is used, these workers will try to
-           monkeypatch the standard library *after* the application is loaded.
+           is discouraged. When ``--preload`` is used, these workers will try
+           to monkeypatch the standard library *after* the application is
+           loaded.
         """
 
 
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -612,6 +612,14 @@ class WorkerClass(Setting):
         * ``gthread``  - Python 2 requires the futures package to be installed
         * ``gaiohttp`` - Deprecated.
 
+        .. note::
+           If you want to use ``'gevent'`` and ``'eventlet'`` workers, avoid
+           monkeypatching (e.g. calling the ``gevent.monkey.patch_all()`` and
+           ``eventlet.monkey_patch()`` functions respectively) in your
+           application. Gunicorn will do it for you after the ``on_starting``
+           hook ran. See also the documentation for the :ref:`preload-app`
+           setting.
+
         Optionally, you can provide your own worker by giving Gunicorn a
         Python path to a subclass of ``gunicorn.workers.base.Worker``.
         This alternative syntax will load the gevent class:
@@ -928,6 +936,11 @@ class PreloadApp(Setting):
         speed up server boot times. Although, if you defer application loading
         to each worker process, you can reload your application code easily by
         restarting workers.
+
+        .. note::
+           Using this setting with the ``'gevent'`` and ``'eventlet'`` workers
+           is discouraged. When ``--reload`` is used, these workers will try to
+           monkeypatch the standard library *after* the application is loaded.
         """
 
 


gunicorn/config.py
```diff
@@ -894,6 +902,11 @@ class PreloadApp(Setting):
         speed up server boot times. Although, if you defer application loading
         to each worker process, you can reload your application code easily by
         restarting workers.
+
+        .. note::
+           Using this setting with the ``'gevent'`` and ``'eventlet'`` workers
+           is discouraged. When ``--reload`` is used, these workers will try to
```
berkerpeksag: I made a silly typo here. `--reload` should have been `--preload`.

docs/source/settings.rst
```diff
@@ -359,6 +367,11 @@ speed up server boot times. Although, if you defer application loading
 to each worker process, you can reload your application code easily by
 restarting workers.
 
+.. note::
+   Using this setting with the ``'gevent'`` and ``'eventlet'`` workers
+   is discouraged. When ``--reload`` is used, these workers will try to
+   monkeypatch the standard library *after* the application is loaded.
```
benoitc: typo?

docs/source/settings.rst
```diff
@@ -359,6 +367,11 @@ speed up server boot times. Although, if you defer application loading
 to each worker process, you can reload your application code easily by
 restarting workers.
 
+.. note::
```
benoitc: maybe being more generic and only making gevent and eventlet as examples?

docs/source/settings.rst
berkerpeksag: I couldn't find the typo. Could you elaborate?

docs/source/settings.rst
berkerpeksag: The note explicitly mentions the gevent and eventlet workers. Or did you mean remove the note directive and make it a normal paragraph?

docs/source/settings.rst
benoitc: shouldn't it be `--preload.`?

docs/source/settings.rst
benoitc: i'm not sure we should discourage people to se async with --preload but rather warn than any monkey patching will be done once in the worker. So afaik what is discouraged is to monkey-patch globally inside a module. Thoughts?

docs/source/settings.rst
benoitc: cc @berkerpeksag

docs/source/settings.rst
berkerpeksag: I'm happy to change the wording but both gevent and eventlet advise calling `monkey_patch()` functions early as possible (and calling them in module level (in `project/__init__.py` for example) is one of the ways to do it) What is the use cause of calling those functions later?

jamadden: I can mention that a large gevent application we deploy *must* use `preload-app` due to a heavy startup time, and in order for that to work we *must* monkey-patch the entire gunicorn arbiter process, which we do by calling `gevent.monkey.patch_all()` in a script before loading and invoking gunicorn's entry point.

Also, gevent does strongly recommend monkey-patching ASAP and that very often means in a module-global location. That's common during development, when a module might look like:
```python
from gevent import monkey
monkey.patch_all()

from myproject.app import wsgi_app # safely separated to another module away from the monkey
from gevent.pywsgi import WSGIServer

server = WSGIServer(('localhost', 0), wsgi_app)
server.serve_forever()
```

But it's not unusual for it to also look like this:
```python
from gevent import monkey
monkey.patch_all()

from gevent.pywsgi import WSGIServer

def wsgi_app(environ, start_response):
   "We can be used with gunicorn or in development."
   "Note how we're in the same module as our monkey patch"
   
if __name__ == '__main__':
    server = WSGIServer(('localhost', 0), wsgi_app)
    server.serve_forever()
```

tilgovi: If we want to document using a custom application [1] for preloading with gevent/eventlet, I would be fine with that. Otherwise, enough things have been imported by the time gunicorn reads the configuration that I'm not sure we can monkey patch early enough. It might be interesting to try, though, by seeing how much we can factor out the loading steps into different modules. At this point, though, without a custom application I don't think we can advise using preload with gevent.

[1] http://docs.gunicorn.org/en/latest/custom.html

benoitc:
file is conflicting, do we still want it?

PR #1438, opened 2017-01-17
add empty line after some tests.

hramezani:
i check all tests and fixed them.


berkerpeksag:
I'm normally not fan of PEP 8-only commits, but in this case it doesn't cause any harm (e.g. cluttering `git blame`) so I just merged it. Thanks!

hramezani:
@berkerpeksag sorry and thanks for approving.

Issue #1045, opened 2015-06-11
Gunicorn's settings - workers vs threads


jkozlowicz:
Could someone clarify what workers and threads in gunicorn's settings are? The advised number of workers/threads in the docs is the same (with the same reasoning).

jasich:
:+1:

tilgovi:
Workers are the number of processes that get spawned. Threads are the number of threads in each process.

Threads is _only_ meaningful with the threaded worker. Every other worker type ignores that setting and runs one thread per process.

The advise and reasoning are the same because the effective concurrency is the same.

nadermx:
So to just clarify you can run threads with workers?  Still having a bit of a issue understanding the abstraction.  

If I have a function with say 5 workers and 5 threads, that means I could have a lot of i/o waiting and have 5 cpu intense tasks running at the same time while waiting for that worker?

heticor915:
Hi After reading the documentation it is not pretty clear for me. What would be the number of workers to run normally .

heticor915:
Hi everybody, after searching for a while I just found  a cool[ tutorial](http://michal.karzynski.pl/blog/2013/06/09/django-nginx-gunicorn-virtualenv-supervisor/) for Setting up Django with Nginx, Gunicorn, virtualenv,and  supervisor.  maybe this could help:

> 
As a rule-of-thumb set the **--workers (NUM_WORKERS)** according to the following formula: _**2 * CPUs + 1.**_ The idea being, that at any given time half of your workers will be busy doing I/O. For a single CPU machine it would give you 3.

tuukkamustonen:
> Threads is only meaningful with the threaded worker. Every other worker type ignores that setting and runs one thread per process.

Might make sense to add this to docs for `--threads` option.

Also should clarify that *threaded worker* refers to *gthread* worker in specific (*gthread* worker is described under "AsyncIO workers" section in http://docs.gunicorn.org/en/stable/design.html because it works on a main loop). Just stating this may be a bit confusing.

Also, does `--workers` have any effect with async(IO) workers? I would assume no, or at least user should not increase them?

tilgovi:
@tuukkamustonen `--workers` affects every type of worker.

[...]

PR #1446, opened 2017-01-26
Doc tweaks around threads setting

RyPeck:
In response to #1045 minor tweaks to the documentation to emphasize that
the `--threads` setting is only applied to the `gthread` worker.

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -113,7 +113,7 @@ How Many Threads?
 ===================
 
 Since Gunicorn 19, a threads option can be used to process requests in multiple
-threads. Using threads assumes use of the sync worker. One benefit from threads
+threads. Using threads assumes use of the gthread worker. One benefit from threads
 is that requests can take longer than the worker timeout while notifying the
 master process that it is not frozen and should not be killed. Depending on the
 system, using multiple threads, multiple worker processes, or some mixture, may
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -148,6 +148,8 @@ application's work load.
 
 If it is not defined, the default is ``1``.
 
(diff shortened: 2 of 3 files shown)


RyPeck:
@berkerpeksag am I reading this snippet right? Gunicorn will switch to a `gthread` worker type if it would have been sync if the `threads` setting is used greater than 1? Looking at code I don't see the sync worker or base worker class using thread.

RyPeck:
There aren't any existing docs for this but I can add them if I understand it correctly.

berkerpeksag:
> Gunicorn will switch to a gthread worker type if it would have been sync if the threads setting is used greater than 1?

Yes, that's correct. Here's a demo:

```py
$ gunicorn --threads=3 echo:app
[2017-01-26 10:05:17 +0000] [28146] [INFO] Starting gunicorn 19.7.0
[2017-01-26 10:05:17 +0000] [28146] [INFO] Listening at: http://127.0.0.1:8000 (28146)
[2017-01-26 10:05:17 +0000] [28146] [INFO] Using worker: threads
```

benoitc:
Why should people know about the worker used? Do we really want it?

berkerpeksag:
IMO implicitly switching to a different worker should be documented.

benoitc:
ok. Thanks for the feedback. :)

berkerpeksag:
BTW, https://github.com/benoitc/gunicorn/pull/1446#pullrequestreview-18561283 hasn't been addressed yet :)

benoitc:
@berkerpeksag  mmm what did I missed, should I revert it?

berkerpeksag:
No need to revert, but it would be nice to mention the fact that the worker will implicitly changed to gthread. @RyPeck could you send another PR?

PR #1455, opened 2017-02-06
move pytest cli args to setup.cfg

aconrad:
This makes the experience of running pytest consistent whether pytest was called from the command line using `py.test` or via `tox`.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ skipsdist = True
 
 [testenv]
 usedevelop = True
-commands = py.test {posargs:tests/} --assert=plain
+commands = py.test {posargs}
 deps =
   -rrequirements_test.txt
   py26: unittest2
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -4,7 +4,7 @@ skipsdist = True
 
 [testenv]
 usedevelop = True
-commands = py.test {posargs:tests/} --assert=plain
```
berkerpeksag: Would `tox -e py35 -- some/test/path/test_foo.py` still work with this change? We might need to keep `{posargs}`.

aconrad: Hum... I only use `tox` to run the full test suite (or to reproduce what the CI system does). If I want to run a specific test, I use pytest directly and specify the filename, such as `py.test some/test/path/test_foo.py`.

berkerpeksag: Right, but it's a bit hard to do that if you want to run the same test on different Python versions :)

aconrad: Ah, I see what you mean. :) Ok I'll add `{posargs}` in there.

aconrad: Updated.

berkerpeksag:
Thanks! :)

PR #1469, opened 2017-02-23
Remove dead code found using vulture

edmorley:
https://pypi.python.org/pypi/vulture

In particular the removal of `get_maxfd()` means the `resource` module is no longer required (which is not available on Windows) and so helps with #524.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -470,22 +470,6 @@ def validate_chdir(val):
     return path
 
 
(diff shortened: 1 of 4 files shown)


edmorley:
Can this be merged? (The failing couverture-io test is a false positive)

tilgovi:
@berkerpeksag approved it and I just double-checked it, so I merged it. Thanks, @edmorley.

PR #1473, opened 2017-02-27
Code Climate config with pep8 and duplication

larkinscott:
Hi Gunicorn team! I'm the Open Source Evangelist at Code Climate! Benoit suggested that I open a PR so that the dev team can discuss the config I created. It currently uses pep8 and our duplication engine (though we can further customize both of these engines). 

If this seems useful, I'd be more than happy to discuss further or tweak this config to make it fit your project better!

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,14 @@
+engines:
+ duplication:
+   enabled: true
+   config:
+     languages:
+     - python
+ pep8:
+   enabled: true
+
+ratings:
+ paths:
+ - "**.py"
+
+exclude_paths:


tilgovi:
We are already linting with prospector in CI.

I guess this would check code duplication and maybe give us a badge for code quality. Are there other benefits?

larkinscott:
Hey @tilgovi! Thanks for taking a look! 

A few additional benefits to running Code Climate along with your CI: 

- A comparison view to identify newly introduced and fixed issues 
- Ratings for each file to see where quality has changed
- Inline content to help resolve the issue
- Results displayed directly in GitHub alongside line-by-line test coverage

tnir:
On #1600, pycodestyle (pep8) is activated through prospector.

PR #1481, opened 2017-03-12
Set CWD and Python path before and after config

### LINKED ISSUE #1349: "--config python:foo" module search path
What I'm trying to do:

```
# myapp/gunicorn.py
from . import something

def on_starting(...): ...
```

If I try to run it with `gunicorn -c file:myapp/gunicorn.py ...` if fails with "cannot perform relative import" because this file is not loaded as a module. OK, I can accept that.

If I try to run it with `gunicorn -c python:myapp.gunicorn ...` it fails with "Error: no module named 'myapp'".

Digging deeper, I find that when I run `python`, I get following for `sys.path`:
`['', '/usr/local/lib/python35.zip', '/usr/local/lib/python3.5', '/usr/local/lib/python3.5/plat-linux', '/usr/local/lib/python3.5/lib-dynload', '/usr/local/lib/python3.5/site-packages']`

However, when I run `gunicorn`, it has following for `sys.path`:
`['/usr/local/bin', '/usr/local/lib/python35.zip', '/usr/local/lib/python3.5', '/usr/local/lib/python3.5/plat-linux', '/usr/local/lib/python3.5/lib-dynload', '/usr/local/lib/python3.5/site-packages']`

(nevermind system-wide paths, that's because docker)

What's the canonical way?
- install myapp into system path?
- teach gunicorn to add "." to system path? (is there a command-line argument?)
- ugly hack in my config? (e.g. `-c file:xxx` which will chain-load real config)

tilgovi:
The config may be specified as a Python module, in which case we want
to ensure that the Python path is fixed up properly before we try to
load it. That means we should follow symlinks and add the current working
directory before and after the configuration is loaded.

Fix #1349

### DIFF
--- a/gunicorn/app/base.py
+++ b/gunicorn/app/base.py
@@ -75,7 +75,17 @@ class BaseApplication(object):
             sys.stderr.flush()
             sys.exit(1)
 
+
 class Application(BaseApplication):
+
+    def chdir(self):
+        # chdir to the configured path before loading,
+        # default is the current dir
+        os.chdir(self.cfg.chdir)
+
+        # add the path to sys.path
+        if self.cfg.chdir not in sys.path:
+            sys.path.insert(0, self.cfg.chdir)
 
     def get_config_from_filename(self, filename):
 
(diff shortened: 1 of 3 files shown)


tilgovi:
I don't understand the couverture report. It seems to be reporting that coverage improved, but the status hooks say it worsened.

aconrad:
There might be a bug in the message saying that coverage has "worsened". What it really means in this case is that not all changes were tested, even though the overall coverage increased. (couverture encourages that all changes should be covered and doesn't satisfy itself by just looking at the overall code coverage)

PR #1483, opened 2017-03-14
Catch errors setting SO_REUSEPORT

### LINKED ISSUE #1480: Upgrading to 19.7.0 means can't bind anymore
Upgraded to latest gunicorn and could not bind ip address anymore (simple wsgi app, ipv4 or ipv6 same story).

```
foo@xxx:~$ cd current
foo@xxx:~/current$ . env/bin/activate
foo@xxx:~/current$ gunicorn my.wsgi.package
[2017-03-10 18:50:46 +0000] [9242] [INFO] Starting gunicorn 19.7.0
[2017-03-10 18:50:46 +0000] [9242] [ERROR] Retrying in 1 second.
[2017-03-10 18:50:47 +0000] [9242] [ERROR] Retrying in 1 second.
[2017-03-10 18:50:48 +0000] [9242] [ERROR] Retrying in 1 second.
[2017-03-10 18:50:49 +0000] [9242] [ERROR] Retrying in 1 second.
[2017-03-10 18:50:50 +0000] [9242] [ERROR] Retrying in 1 second.
[2017-03-10 18:50:51 +0000] [9242] [ERROR] Can't connect to ('127.0.0.1', 8000)
```

Reverting to 19.6.0 fixes the problem.

System version : Linux xxx 3.2.0-4-amd64 #1 SMP Debian 3.2.68-1+deb7u1 x86_64 GNU/Linux

Let me know if I can provide more info, 19.6.0 is acceptable solution for me but I guess that's a regression you wanna at least understand ?

tilgovi:
Fix #1480

### DIFF
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -39,8 +39,12 @@ class BaseSocket(object):
 
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        if hasattr(socket, 'SO_REUSEPORT'):
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        if hasattr(socket, 'SO_REUSEPORT'):  # pragma: no cover
+            try:
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+            except socket.error as err:
+                if err[0] not in (errno.ENOPROTOOPT, errno.EINVAL):
+                    raise
         if not bound:
             self.bind(sock)
         sock.setblocking(0)


tilgovi:
The py36 coverage seems totally spurious.

aconrad:
Would it be possible that the tests are not guaranteed to run in a deterministic way where the code path for this branching logic could be different for some test runs? That's usually what the issue is. I've seen this same exact coverage report in other PRs of gunicorn.

aconrad:
@tilgovi Just so you know, I put a breakpoint just before `raise LimitRequestLine(len(data), limit)` and ran `tox -e py35` in a loop and indeed, sometimes the breakpoint hits, and sometimes it doesn't (most of the time in my experience). My guess is that a dictionary is at play somewhere, the ordering of the keys is less deterministic in Python 3 than it is in Python 2 (but it might be something else too).

aconrad:
Here's what I actually ran (so I don't have to wait so long for all the tests to run):

```bash
while true; do tox -e py35 -- tests/test_invalid_requests.py; done
```

tilgovi:
I think it's more deliberate than that. Search `random` in https://github.com/benoitc/gunicorn/blob/master/tests/treq.py

aconrad:
Hah, there you go. I think it would be a good opportunity to fix the tests so they are deterministic (and also write additional tests to cover both sides of the coin)

gunicorn/sock.py
```diff
@@ -39,8 +39,14 @@ def __getattr__(self, name):
 
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        if hasattr(socket, 'SO_REUSEPORT'):
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        if hasattr(socket, 'SO_REUSEPORT'):  # pragma: no cover
+            try:
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+            except socket.error as err:
+                if err[0] in (errno.ENOPROTOOPT, errno.EINVAL):
```
berkerpeksag: Or a simpler version:

```py
if err[0] not in (errno.ENOPROTOOPT, errno.EINVAL):
    raise
```

tilgovi:
Updated.

gunicorn/sock.py
```diff
@@ -39,8 +39,12 @@ def __getattr__(self, name):
 
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
-        if hasattr(socket, 'SO_REUSEPORT'):
-            sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+        if hasattr(socket, 'SO_REUSEPORT'):  # pragma: no cover
+            try:
+                sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
+            except socket.error as err:
+                if err[0] not in (errno.ENOPROTOOPT, errno.EINVAL):
```
benoitc: i'm wondering if instead of raising an error, we couldn't simply try to start without this option and log a warning. Thoughts?

tilgovi: Looking at the errors that setsockopt(2) can return, I think anything else is probably very bad.

benoitc:
@aconrad what do you mean by more deterministic?

gunicorn/sock.py
benoitc: fair enough :)

aconrad:
@benoitc Couverture.io showed us that the code path execution of each test run may be different each time (not guaranteed, non-deterministic), likely due to the `random` module used in tests https://github.com/benoitc/gunicorn/blob/master/tests/treq.py, as @tilgovi suggested.

As we can see from the coverage report, line 215 or line 218 of file https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/message.py#L215-L218 may be executed during test runs (unpredictable):

https://app.couverture.io/coverage/github/benoitc/gunicorn/9e076f4246c80f2091163242e7c985ad6ccb78c7/py36

So if the culprit is indeed the `random` module, we should set a proper seed value so we can guarantee repeatable outcomes of the tests. We might want to set a `PYTHONHASHSEED` env variable which tox describes here (even though it's a Python feature, not a tox one): https://tox.readthedocs.io/en/latest/example/basic.html#special-handling-of-pythonhashseed

Once fixed (e.g. line 215 is guaranteed to be the code path taken) then we should write additional tests to make sure that line 218 is also covered eventually.

I hope this helps.

gunicorn/sock.py
abitduck: It's OK with this patch~

hartym: Fixes #1480 on my system.

PR #1491, opened 2017-03-28
Fix "TypeError: 'OSError' object is not subscriptable" in py3

fernandezcuesta:
In python3 the exception object is not iterable and must use either the `args` attribute or in this case `errno` (equivalent to `.args[0]`).


berkerpeksag:
Thanks!

tilgovi:
Thank you!

CaselIT:
@berkerpeksag I'm facing the same bug fixed in this pr in the current version. Is it possible to publish this update on pypi?
Thanks

PR #1495, opened 2017-04-05
Add --statsd-rate to set the sampling rate for statsd

jamescarr:
For our usage we need to tune the sampling rate down a bit as the default of `1.0` pushes way too much udp traffic at our internal monitoring infrastructure. As it stands only a much smaller percentage is needed for us to accurately monitor the bits in gunicorn that interest us. 

This PR simply adds a new option (`--statsd-rate`) with a default sampling rate set to what it is now (`1.0`). 

I'm not too proud of the tests here but I'm not 100% sure about how to capture that value without digging too deep, so the tests simply verify the sampling rate that is set.


gunicorn/config.py
```diff
@@ -1384,6 +1384,18 @@ class StatsdPrefix(Setting):
     .. versionadded:: 19.2
     """
 
+class StatsdRate(Setting):
+    name = "statsd_rate"
+    section = "Logging"
+    cli = ["--statsd-rate"]
+    meta = "STATSD_RATE"
+    default = "1.0"
+    validator = validate_string
```
bryanhelmig: Is there a validate float option?

jamescarr: Hmmm doesn't look like it... 

```
330:def validate_bool(val):
346:def validate_dict(val):
352:def validate_pos_int(val):
363:def validate_string(val):
371:def validate_list_string(val):
382:def validate_string_to_list(val):
391:def validate_class(val):
399:def validate_callable(arity):
423:def validate_user(val):
437:def validate_group(val):
452:def validate_post_request(val):
466:def validate_chdir(val):
480:def validate_hostport(val):
491:def validate_reload_engine(val):
```

jamescarr: I should add that too since this PR is actually using a string but neglecting to cast it.

charettes:
not sure about the test failures but I assume this will need to be documented in `docs/source/settings.rst` as well.

jamescarr:
retest

gunicorn/instrument/statsd.py
```diff
@@ -23,11 +23,13 @@
 class Statsd(Logger):
     """statsD-based instrumentation, that passes as a logger
     """
+
```
berkerpeksag: Style nit: We usually avoid doing unnecessary style changes.

PR #1499, opened 2017-04-11
Fallback to the original behavior if prefix is not available (#1068)

yyuu:
I needed to restore the original behavior because some external applications was actually relying on the behavior.

https://github.com/apache/incubator-airflow/blob/1.8.0/airflow/bin/cli.py#L775

I believe this needs to be fixed in gunicorn side because the caller of `gunicorn` cannot know how gunicorn behaves on `-c` before it actually invokes `gunicorn`.

### DIFF
--- a/gunicorn/app/base.py
+++ b/gunicorn/app/base.py
@@ -121,12 +121,14 @@ class Application(BaseApplication):
         if location.startswith("python:"):
             module_name = location[len("python:"):]
             cfg = self.get_config_from_module_name(module_name)
-        else:
-            if location.startswith("file:"):
-                filename = location[len("file:"):]
-            else:
-                filename = location
+        elif location.startswith("file:"):
+            filename = location[len("file:"):]
             cfg = self.get_config_from_filename(filename)
+        else:
+            try:
+                cfg = self.get_config_from_module_name(location)
+            except ImportError:
+                cfg = self.get_config_from_filename(location)
 
         for k, v in cfg.items():
             # Ignore unknown names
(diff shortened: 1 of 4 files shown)


ToGoBananas:
+1

benoitc:
@berkerpeksag  the change is quite old though (2 years in #1068)  . I guess we need to cover it by a test before merging it. Just to make sure we don't forget anything.

 Or maybe we can check we try to load a module? Like checking if it has an extension at least. Thoughts?

berkerpeksag:
@yyuu sorry, I overwrote your original change in `gunicorn/app/base.py` while I was trying to push a test to your branch.

gunicorn/app/base.py
```diff
@@ -121,12 +121,14 @@ def load_config_from_module_name_or_filename(self, location):
...
-            if location.startswith("file:"):
-                filename = location[len("file:"):]
-            else:
-                filename = location
+        elif location.startswith("file:"):
+            filename = location[len("file:"):]
             cfg = self.get_config_from_filename(filename)
+        else:
+            try:
+                cfg = self.get_config_from_module_name(location)
```
benoitc: if we re-add that feature, wouldn’it be better to add the warning based kn the principle of least astonishment? Espacially since this chang re-add a behaviour removed 2 years ago. thoughts?

benoitc:
@berkerpeksag bump.

berkerpeksag:
In #1634, we decided not to merge this. We've changed the behavior in 19.4 and our latest bugfix release will be 19.8. Ideally it would be nice to keep backward compatibility in bugfix releases but unfortunately that ship has sailed two years ago. Thank you for your work, @yyuu.

PR #1504, opened 2017-04-21
optionally send datadog tags with statsd

dblackdblack:
The datadog hosted metrics service uses a slightly forked statsd protocol in order to add arbitrary key/value tags to metrics. This PR allows the user to set a static list of datadog tags to send along with statsd metrics. I would like to be able to e.g. send the tags "app_name:my_awesome_web_app" and "stage:prod" along with my stats. This lets you easily create dashboard that only show stats from the prod stage.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -47,6 +47,7 @@ Dan Callaghan <dcallagh@redhat.com>
 Dan Sully <daniel-github@electricrain.com>
 Daniel Quinn <code@danielquinn.org>
 Dariusz Suchojad <dsuch-github@m.zato.io>
+David Black <github@dhb.is>
 David Vincelli <david@freshbooks.com>
 David Wolever <david@wolever.net>
 Denis Bilenko <denis.bilenko@gmail.com>
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -376,6 +376,17 @@ if not provided).
 
 .. versionadded:: 19.2
 
+dogstatsd_tags
+~~~~~~~~~~~~~~
+
+* ``--dogstatsd-tags DOGSTATSD_TAGS``
+* ``(empty string)``
+
+Comma-delimited list of static dogstatsd (datadog statsd) tags sent with all statsd metrics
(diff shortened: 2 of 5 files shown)


dubbl:
Hi, what is the general opinion regarding this feature?
I would also be interested, even though I don't use datadog at all. 
Instead I use the open source projects telegraf, influxdb and grafana to record and display telemetry, and [telegraf also supports datadogs tags extension to statsd](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/statsd).

tilgovi:
Seems fine to me, needs a quick rebase. @dblackdblack interested?

tilgovi:
I also think renaming it `--dogstatsd-tags` might make sense, since that's how they refer to the protocol: https://docs.datadoghq.com/developers/dogstatsd/

tilgovi:
Needs a rebase to resolve conflicts, but otherwise looks good to me.

dblackdblack:
rebased.

dblackdblack:
any else I need to do here?  rebase again?  @tilgovi

tilgovi:
Looks good! Thank you for your patience.

tilgovi:
Want to squash up the commits at all?

dblackdblack:
@tilgovi  all squashed up and rebased

tilgovi:
One last change! Can we make the version added 20? I think master has gone far enough that we'll be looking at more than a patch release. If we do make a patch release to 19.9, we'd have to fork off from earlier on and back-port this.

dblackdblack:
done.

tilgovi:
Thank you, @dblackdblack!

dnlserrano:
Hey guys, thanks for the all your hard work in gunicorn! ❤️
Is this getting released any time soon? Thanks in advance! 🤞

tilgovi:
Please follow #2022 or watch the repository for releases. I think everything is in place, but in case I forget to post back here, you should follow in those ways.

PR #1505, opened 2017-04-23
Introduce --startup-timeout option

messense:
One of my app requires several minutes to startup(loads a lot of data), the `timeout` option seems not quite right for allowing workers to startup within some time.

I think adding a `--startup-timeout` option would be better to allow workers to startup in a certain amount of time while booted workers silent for more than `--timeout` seconds are still get killed and restarted.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -210,6 +210,16 @@ you're sure of the repercussions for sync workers. For the non sync
 workers it just means that the worker process is still communicating and
 is not tied to the length of time required to handle a single request.
 
+.. _startup-timeout:
+
+startup_timeout
+~~~~~~~~~~~~~~~
+
+* ``--startup-timeout INT``
+* ``30``
+
+Timeout for workers startup.
+
 .. _graceful-timeout:
 
 graceful_timeout
@@ -319,6 +329,24 @@ because it consumes less system resources.
 .. note::
    In order to use the inotify reloader, you must have the ``inotify``
    package installed.
+
+.. _reload-engine:
+
+reload_engine
+~~~~~~~~~~~~~
+
+* ``--reload-engine STRING``
+* ``auto``
+
+The implementation that should be used to power :ref:`reload`.
+
+Valid engines are:
+
+* 'auto'
+* 'poll'
+* 'inotify' (requires inotify)
+
+.. versionadded:: 19.7
 
 .. _spew:
 
@@ -1138,7 +1166,7 @@ ssl_version
 ~~~~~~~~~~~
 
 * ``--ssl-version``
-* ``3``
+* ``_SSLMethod.PROTOCOL_TLS``
 
 SSL version to use (see stdlib ssl module's)
 
@@ -1152,7 +1180,7 @@ cert_reqs
 ~~~~~~~~~
 
 * ``--cert-reqs``
-* ``0``
+* ``VerifyMode.CERT_NONE``
 
 Whether client certificate is required (see stdlib ssl module's)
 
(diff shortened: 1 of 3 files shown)


messense:
I am not sure how to add test cases for it. 

Did not find any test cases for `--timeout` or `--graceful-timeout` options.

docs/source/settings.rst
```diff
@@ -210,6 +210,16 @@ you're sure of the repercussions for sync workers. For the non sync
...
 
+.. _startup-timeout:
+
+startup_timeout
+~~~~~~~~~~~~~~~
+
+* ``--startup-timeout INT``
+* ``30``
+
+Timeout for workers startup.
```
vvk: It would be nice if you specify in the description what time units the value has (milliseconds/seconds/minutes?).
Reasonable question from people who will use it: is it timeout per worker or aggregation of all workers startup time? Need to specify it clearly in the docs.

Also you better to point out in the docs explicitly how `--startup-timeout` correlates with `--timeout` if they both are specified.

docs/source/settings.rst
```diff
@@ -320,6 +330,24 @@ because it consumes less system resources.
    In order to use the inotify reloader, you must have the ``inotify``
    package installed.
 
+.. _reload-engine:
```
vvk: How does this relates to this PR?

gunicorn/config.py
```diff
@@ -715,6 +715,19 @@ class Timeout(Setting):
...
+class StartupTimeout(Setting):
+    name = 'startup_timeout'
+    section = "Worker Processes"
+    cli = ["--startup-timeout"]
+    meta = "INT"
+    validator = validate_pos_int
+    type = int
+    default = 30
+    desc = """\
+        Timeout for workers startup.
```
vvk: The same about the docs as in the comment above.

docs/source/settings.rst
messense: `settings.rst` is auto generated from `config.py`, this means someone forget to generate them before.

https://github.com/benoitc/gunicorn/blob/master/CONTRIBUTING.md

> If you are adding a new configuration option or updating an existing one, please do it in gunicorn/config.py, then run make -C docs html to update docs/source/settings.rst.

gunicorn/arbiter.py
```diff
@@ -491,6 +491,8 @@ def murder_workers(self):
         workers = list(self.WORKERS.items())
         for (pid, worker) in workers:
             try:
+                if not worker.booted and time.time() - worker.tmp.last_update() <= self.cfg.startup_timeout:
```
tilgovi: It could be maybe a bit more clear here and make this clause not depend on ordering with the one after it.

```python
timeout = self.cfg.timeout if worker.booted else self.cfg.startup_timeout
if time.time() - worker.tmp.last_update() <= timeout:
    continue
```

Is that worth doing?

messense:
@tilgovi 
This PR has been stalled here for a long time without any feedback until recently, so before anymore work on this, I want to make sure we all understand why we need this feature and are in favor of adding this to gunicorn. Thanks!

tilgovi:
@messense I agree!

The motivation makes sense to me; some apps take a long time to boot. I do not think this PR works as intended, though.

Before you change it we should agree on what to do, though. You are right about this.

Is the PR necessary? Is there another way to accomplish this? I think `--preload` and USR2 might handle your use case. Perhaps you could say more about your issue.

messense:
`--preload` and USR2 signal works for me. I'll close this, Thanks!

tilgovi:
Thanks! If it turns out that it does not work for you for some reason, please open an issue. I would be happy to help.

PR #1511, opened 2017-05-11
Do not strip leading slash from path

YuppY:


benoitc:
why this PR? What are you trying to fix?

YuppY:
Commented in https://github.com/benoitc/gunicorn/issues/1512

tilgovi:
This may be relevant:

    A URL-path-segment string must be one of the following:

    * zero or more URL units, excluding U+002F (/) and U+003F (?), that together are not a single-dot path segment or a double-dot path segment.

Source: https://url.spec.whatwg.org/#url-path-segment-string

Note "zero or more", so "//" is valid in a URL and I guess distinct from "/". We probably shouldn't strip it unless we have a good reason to.

tilgovi:
@benoitc what do you think?

tilgovi:
This needs a rebase, but I think this should move forward.

YuppY:
Rebased.

By the way, it fixes security issue: if we configure reverse proxy for gunicorn app that denies access to '/admin/', then an attacker can bypass this rule by requesting 'https://example.org//admin/'.

We patched our proxy configurations to redirect requests from '//+(.+)' to '/$1' but there could be more apps with such vulnerability.

requirements_test.txt
```diff
@@ -1,3 +1,4 @@
 coverage>=4.0,<4.4  # TODO: https://github.com/benoitc/gunicorn/issues/1548
+mock; python_version < "3"
```
tilgovi: This is already handled in `setup.py`.

tilgovi: I think this form is cleaner, but maybe we do that for mock and unittest2 in a separate change.

YuppY: ok. removed this commit from pr

gunicorn/util.py
```diff
@@ -543,3 +543,14 @@ def app(environ, start_response):
...
 
     return app
+
+
+def split_request_uri(uri):
+    if uri.startswith("//"):
+        # When the path starts with //, urlsplit considers it as a
+        # relative uri while the RFC says we should consider it as abs_path
+        # http://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html#sec5.1.2
+        parts = _compat.urlsplit("." + uri)
```
benoitc: not sure i understand this line. What trigger the "." here ?

tilgovi: I'm not sure the comment makes sense, but a quick test in the console shows that this forces `urlsplit` to consider the string as a path—with no scheme or netloc—rather than parsing this as a scheme-relative URL with the first path component as the netloc.

YuppY: "." forces urlsplit to parse path part as RFC says:
```python
In [5]: urlsplit('//path?query#fragment')
Out[5]: SplitResult(scheme='', netloc='path', path='', query='query', fragment='fragment')

In [6]: urlsplit('.//path?query#fragment')
Out[6]: SplitResult(scheme='', netloc='', path='.//path', query='query', fragment='fragment')
```

gunicorn/http/message.py
```diff
@@ -312,18 +312,10 @@ def parse_request_line(self, line_bytes):
         self.method = bits[0].upper()
 
         # URI
```
berkerpeksag: Can you delete this line too? New code is descriptive enough so  we don't need it anymore.

gunicorn/util.py
berkerpeksag: Can you add a test case in `tests/test_util.py`? It's easy to introduce a regression when fixing another issue in the urllib module. This way we can know early if something has changed in the development version of Python.

gunicorn/http/message.py
YuppY: It is consistent with other comments in this function: [Method](https://github.com/benoitc/gunicorn/blob/cb2f47183153368b70b9c06d14e43a4bf6ac7d92/gunicorn/http/message.py#L309) and [Version](https://github.com/benoitc/gunicorn/blob/cb2f47183153368b70b9c06d14e43a4bf6ac7d92/gunicorn/http/message.py#L333).

gunicorn/util.py
benoitc: @tilgovi thanks. imo that worlrth a comment for readability.

YuppY:
made test and comment in `split_request_uri`

YuppY:
do I need to fix something to get it merged?

berkerpeksag:
Thanks!

PR #1516, opened 2017-05-24
Use getfullargspec instead of getargspec (where available) to silence deprecation warning.

### LINKED ISSUE #1515: Config calls inspect.getargspec() which is deprecated, resulting in a warning
The warning is as follows:

```
  /Users/riceb2/.local/share/virtualenvs/guvnor-iF69-5_x/lib/python3.6/site-packages/gunicorn/config.py:417: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
    if arity != -1 and arity != len(inspect.getargspec(val)[0]):
```

jeamland:
Fixes #1515.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -414,7 +414,7 @@ def validate_callable(arity):
                     "" % (obj_name, mod_name))
         if not six.callable(val):
             raise TypeError("Value is not six.callable: %s" % val)
-        if arity != -1 and arity != len(inspect.getargspec(val)[0]):
+        if arity != -1 and arity != len(_compat.getargspec(val)[0]):
             raise TypeError("Value must have an arity of: %s" % arity)
         return val
     return _validate_callable
--- a/gunicorn/_compat.py
+++ b/gunicorn/_compat.py
@@ -262,3 +262,7 @@ if PY26:
 
 else:
     from gunicorn.six.moves.urllib.parse import urlsplit
+
+import inspect
+
+getargspec = getattr(inspect, 'getfullargspec', inspect.getargspec)


gunicorn/config.py
```diff
@@ -396,6 +396,12 @@ def validate_class(val):
     return validate_string(val)
 
 
+if hasattr(inspect, 'getfullargspec'):
```
berkerpeksag: `getattr(inspect, 'getfullargspec', inspect.getargspec)` would be shorter.

jeamland: Doing that now.

jeamland:
Are the lint tests failing on master too? They appear to be based on what I can see...

berkerpeksag:
It looks like for some reason prospector ignored the `ignore-paths` setting at https://github.com/benoitc/gunicorn/blob/master/.prospector.yaml#L3

gunicorn/config.py
```diff
@@ -396,6 +396,9 @@ def validate_class(val):
     return validate_string(val)
 
 
+getargspec = getattr(inspect, 'getfullargspec', inspect.getargspec)
```
berkerpeksag: Sorry, I forgot to say this in my earlier comment, but could you please move this to `gunicorn/_compat.py`?

jeamland: Done.

jeamland:
lint is something to do with prospector 0.12.5. Locking the version to 0.12.4 in tox.ini makes the lint failures stop. I'm digging in to that now.

PR #1482, opened 2017-03-13
Remove gunicorn_paster and paste serve integration

tilgovi:
The `--paste` option still works, but using egg:gunicorn in the server
block of an ini file will no longer work and the `gunicorn_paster`
script is gone.

These features were deprecated in 19.0.

Close #1189

PR #1519, opened 2017-05-24
Unlock pylint and deal with a large swathe of lint failures.

### LINKED ISSUE #1518: lint fails with prospector >= 0.12.5
Locking the version of prospector to 0.12.4 in tox.ini makes lint pass. Letting it install 0.12.5 but unlocking pylint doesn't fix the problem. Still digging.

jeamland:
This is mostly trivial stuff. There was at least one bug found, being bad format string in statsd decrement. This will want a significant amount of review though as I'm not amazingly familiar with the codebase so I may have made a bad change in fixing things.

The main area I'd like reviewed is the removal of the imports in the run functions in `gunicorn.app.wsgiapp` and `gunicorn.app.pasterapp`.

I've left it as a bunch of smaller commits for now but I can squash them down if desired.

Fixes #1518

### DIFF
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -31,6 +31,7 @@ pylint:
   disable:
     - import-error
     - too-many-locals
+    - wrong-import-position
 
 mccabe:
   run: false
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -27,5 +27,9 @@ pyflakes:
     # Access to a protected member $X of a client class
     - W0212
 
+pylint:
+  disable:
+    - too-many-locals
+
 mccabe:
   run: false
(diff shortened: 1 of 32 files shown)


berkerpeksag:
Thanks again!

Some quick comments:

You're right that some of these are valid bugs. `def __str__(self, name):` for example.

I'm not sure about renaming `foo` to `_foo` though. We shouldn't change function and method signatures. It might be better to turn the warning off.

I'm also not sure about `# pylint: ...` comments. It would be better if we could keep all pylint configuration in one place.

I wouldn't touch and ignore paster stuff since we are going to remove most of the code in #1482.

> The main area I'd like reviewed is the removal of the imports in the run functions in gunicorn.app.wsgiapp and gunicorn.app.pasterapp.

They are imported in run() functions to avoid circular import errors. Also, those functions are main entrypoints for CLIs. See https://github.com/benoitc/gunicorn/blob/master/setup.py#L96 for details.

PR #1520, opened 2017-05-25
Unlock pylint and fix lint errors

### LINKED ISSUE #1515: Config calls inspect.getargspec() which is deprecated, resulting in a warning

jeamland:
This is a much better version of #1519 and also supersedes #1516. Commits are now better separated based on the category of fault they're fixing. There are no `# pylint` comments. More categories are unilaterally disabled in the prospector configuration as there is no way in the prospector configuration to disable a category for a specific file.

Fixes #1515 and #1518.

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -109,12 +109,10 @@ class SafeAtoms(dict):
             kl = k.lower()
             if kl in self:
                 return super(SafeAtoms, self).__getitem__(kl)
-            else:
-                return "-"
+            return "-"
         if k in self:
             return super(SafeAtoms, self).__getitem__(k)
-        else:
-            return '-'
+        return '-'
 
 
 def parse_syslog_address(addr):
--- a/gunicorn/instrument/statsd.py
+++ b/gunicorn/instrument/statsd.py
@@ -80,7 +80,7 @@ class Statsd(Logger):
                         pass
 
             # Log to parent logger only if there is something to say
-            if msg is not None and len(msg) > 0:
+            if msg:
                 Logger.log(self, lvl, msg, *args, **kwargs)
         except Exception:
             Logger.warning(self, "Failed to log to statsd", exc_info=True)
--- a/gunicorn/instrument/statsd.py
+++ b/gunicorn/instrument/statsd.py
@@ -108,7 +108,7 @@ class Statsd(Logger):
         self._sock_send("{0}{1}:{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
 
     def decrement(self, name, value, sampling_rate=1.0):
-        self._sock_send("{0){1}:-{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
+        self._sock_send("{0}{1}:-{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
 
     def histogram(self, name, value):
         self._sock_send("{0}{1}:{2}|ms".format(self.prefix, name, value))
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -61,11 +61,6 @@ except ImportError:
     has_inotify = False
 
 
-class InotifyReloader():
-    def __init__(self, callback=None):
-        raise ImportError('You must have the inotify module installed to use '
-                          'the inotify reloader')
-
 if has_inotify:
 
     class InotifyReloader(threading.Thread):
@@ -112,6 +107,13 @@ if has_inotify:
                 filename = event[3]
 
                 self._callback(filename)
+
+else:
+
+    class InotifyReloader():
+        def __init__(self, callback=None):
+            raise ImportError('You must have the inotify module installed to '
+                              'use the inotify reloader')
 
 
 preferred_reloader = InotifyReloader if has_inotify else Reloader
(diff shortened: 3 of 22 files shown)


gunicorn/reloader.py
```diff
@@ -113,6 +108,13 @@ def run(self):
 
                 self._callback(filename)
 
+else:
+
+    class InotifyReloader():
```
berkerpeksag: Could you change this to

```py
```

while we are here.

gunicorn/glogging.py
```diff
@@ -109,12 +109,10 @@ def __getitem__(self, k):
             kl = k.lower()
             if kl in self:
                 return super(SafeAtoms, self).__getitem__(kl)
-            else:
```
berkerpeksag: While I prefer your style, there is nothing wrong with this code (and other similar usages of the else clause) and it would be great if we could avoid code churn.

gunicorn/instrument/statsd.py
```diff
@@ -108,7 +108,7 @@ def increment(self, name, value, sampling_rate=1.0):
         self._sock_send("{0}{1}:{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
 
     def decrement(self, name, value, sampling_rate=1.0):
-        self._sock_send("{0){1}:-{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
+        self._sock_send("{0}{1}:-{2}|c|@{3}".format(self.prefix, name, value, sampling_rate))
```
berkerpeksag: Good catch!

gunicorn/workers/ggevent.py
```diff
@@ -143,13 +143,15 @@ def run(self):
...
             # Force kill all active the handlers
             self.log.warning("Worker graceful timeout (pid:%s)" % self.pid)
-            [server.stop(timeout=1) for server in servers]
+            for server in servers:
+                server.stop(timeout=1)
         except:
             pass
 
-    def handle_request(self, *args):
+    def handle_request(self, listener_name, req, sock, addr): #arguments-differ
```
berkerpeksag: This comment is a bit cryptic. Arguments differ from base Worker class or Sync or Async? It looks like base class of this worker is Async and it has the same signature as this one:

```py
def handle_request(self, listener_name, req, sock, addr):
    ...
```

So perhaps we don't need a comment after all?

tests/test_config.py
```diff
@@ -77,7 +77,7 @@ def test_property_access():
     assert os.getegid() == c.gid
 
     # Proc name
-    assert "gunicorn" == c.proc_name
+    assert c.proc_name == "gunicorn" # misplaced-comparison-constant
```
berkerpeksag: While I think the old version looks like a PHP code, I think it's perfectly fine as-is. I'd prefer disabling misplaced-comparison-constant check.

berkerpeksag:
I went ahead and cherry-picked the following commits into make master green again:

* Deal fully with deprecated functions. - https://github.com/benoitc/gunicorn/commit/72dd0199d7d1d8b3a792ccf2f4e691c1c9174486
* Don't use a list comprehension as shorthand for a for loop. - https://github.com/benoitc/gunicorn/commit/5b89fe1beb3d6e287fbe56a0f0a2d302cbaddb7e
* Empty sequences are false, there's no need to check the len. - https://github.com/benoitc/gunicorn/commit/032271a030a6bbe39d22e9ebadf2a86ddf3b6705
- Remove some errant trailing commas. - https://github.com/benoitc/gunicorn/commit/e6edcb454abbd6c2059e39b5bbfd6f7a5ccbc1f2
- Fix a broken format string. - https://github.com/benoitc/gunicorn/commit/1f42421762913912422abf5074db73469dd99186
- __str__ doesn't take any extra parameters. - https://github.com/benoitc/gunicorn/commit/e81c654c2d0116ca371fdf40a02e335ced29c45a
- RequestParser takes two parameters. - https://github.com/benoitc/gunicorn/commit/a4461ba1586808603bbbf989c024d2c55ecceb12

berkerpeksag:
Move a default class definition into an else block so we're not redefining. - https://github.com/benoitc/gunicorn/commit/20616d50fcdb066adba3a5afa9a6a692dbc7cc4b (I've addressed my review comment locally.)

berkerpeksag:
I prefer not to cherry-pick the following commits (see my review above):

* Put things in the preferred order. - https://github.com/benoitc/gunicorn/pull/1520/commits/8bd402e07346bdf8ad6c5bb46deb7625287c34b1
* Don't put the return in an else if there's one in the if. - https://github.com/benoitc/gunicorn/pull/1520/commits/24ac9642860faca56a553ad3a45c06cb31d231af

berkerpeksag:
Here are more cherry-picked commits:

* Ensure that overriding methods match the signature they've overriden. - https://github.com/benoitc/gunicorn/commit/0d116917186700ae681297113bddb86634cc52eb (I've removed the comment locally.)
* Unlock pylint as prospector now works with >=1.7. Use Python 3.6 for lint. - https://github.com/benoitc/gunicorn/commit/f69c17f8e29c88faf3d7c6cd13bec9abbdf78ba3 (I've removed outdated comment locally.)

berkerpeksag:
0ffd232bc740a485e559093dc58c811b36f77b8e ("Disable a number of lint categories.") is the only commit that is not merged into master. I've opened #1543 to tweak lint categories of prospector. Thank you for the PR again!

PR #1527, opened 2017-06-13
new --reload-extra-file option

garbas:
I quickly added an option which is implemented but not exposed in the command line.
Would something like this be excepted? If yes I'll add the documentation for the command line option.

/cc @tilgovi (author of ``gunicorn/reloader.py``)

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -368,6 +368,14 @@ def validate_string(val):
     return val.strip()
 
 
+def validate_file_exists(val):
+    if val is None:
+        return None
+    if not os.path.exists(val):
+        raise ValueError("File %s does not exists." % val)
+    return val
+
+
 def validate_list_string(val):
     if not val:
         return []
@@ -377,6 +385,10 @@ def validate_list_string(val):
         val = [val]
 
     return [validate_string(v) for v in val]
+
+
+def validate_list_of_existing_files(val):
+    return [validate_file_exists(v) for v in validate_list_string(val)]
 
 
 def validate_string_to_list(val):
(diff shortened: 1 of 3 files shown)


tilgovi:
Seems okay to me!

tilgovi:
As far as I can tell, nothing will break if you don't validate that the files exist, but maybe it's the right idea so that users don't silently watch the wrong thing?

garbas:
@tilgovi I usually follow the "fail early" rule and if somebody provides an option that points to a file which we should watch and that file does not exists it is very likely to be a typo. I would assume 99% of cases is like this. In the other 1%, you can always use ``reloader.py`` module programmaticaly and skip the validation.

I added the description for new option and squashed the work in one commit. As far as I can tell this looks ready to be merged, if not let me know wha

gunicorn/config.py
```diff
@@ -379,6 +388,17 @@ def validate_list_string(val):
     return [validate_string(v) for v in val]
 
 
+def validate_list_of_existing_files(val):
+    if not val:
```
tilgovi: Maybe remove these few lines and replace with a call to `validate_list_string`?

gunicorn/config.py
```diff
@@ -368,6 +368,15 @@ def validate_string(val):
     return val.strip()
 
 
+def validate_file_exists(val):
+    val = validate_string(val)
+    if val is None:
```
tilgovi: I'm not sure we want this here. For the string options, I think it's there for when the default is `None` if the option wasn't provided at all. In the case of this option, the default is an empty list.

gunicorn/config.py
```diff
@@ -857,6 +877,20 @@ class ReloadEngine(Setting):
...
+    name = "reload_extra_files"
+    action = "append"
+    section = "Debugging"
+    cli = ["--reload-extra-file"]
+    meta = "FILES"
+    validator = validate_list_of_existing_files
+    default = []
+    desc = """\
+        Extends --reload option to also watch and reload on additional files
+        (eg. templates, configurations, specifications, ...).
```
tilgovi: nit: "(e.g., templates, configurations, specifications, etc.)."

tilgovi:
Thanks, @garbas. Merged. 👍

PR #1528, opened 2017-06-14
Redirect accesslog issue 1403

hramezani:


tilgovi:
If we do this we should probably have the default to be `True` so that it's not changing the defaults. What do you think?

tilgovi:
Also, I think it would be preferable if the option made it clear from its name that it's redirecting to syslog.

tilgovi:
Oh, I see now that the issue that motivated this. Maybe we should change the default. We could do that for R20. What do you think, @benoitc?

hramezani:
@benoitc @tilgovi, I am waiting for your decision for default value

berkerpeksag:
I agree that we should keep the default behavior in 19.x. I agree with @tilgovi's plan:

In 19.x:

* Add a `--disable-access-log-redirection` (we should probably include "syslog" to the option name) to disable current behavior.

In 20.0:

* Change default behavior not to send access logs to syslog.
* Add a new option to send access logs to syslog.
* Remove `--disable-access-log-redirection`.

hramezani:
@berkerpeksag thanks, please check my changes

docs/source/settings.rst
```diff
@@ -609,6 +609,16 @@ The Access log file to write to.
...
+.. _disable-access-log-redirection:
+
+disable_access_log_redirection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* ``--disable-access-log-redirection``
+* ``False``
+
+Disable redirect access logs to syslog.
+
```
berkerpeksag: Needs 

```rst
.. versionadded:: 19.8
```

gunicorn/glogging.py
```diff
@@ -314,7 +315,8 @@ def access(self, resp, req, environ, request_time):
         for format details
         """
 
-        if not (self.cfg.accesslog or self.cfg.logconfig or self.cfg.syslog):
+        if not (self.cfg.accesslog or self.cfg.logconfig or \
```
berkerpeksag: `\` is not needed here since we are already in `if not (...)`.

docs/source/settings.rst
berkerpeksag: Also, it would be nice to add a note to `syslog` documentation at http://docs.gunicorn.org/en/stable/settings.html#syslog

```rst
.. versionchanged:: 19.8

   You can now disable sending access logs by using the
   :ref:`disable-access-log-redirection` setting.
```

hramezani:
@berkerpeksag thanks, requested changes are ready

benoitc:
hrm the option name is misleading somehow. maybe `--disable-redirect-access-to-syslog`  or `--redirect-access-to-syslog=no`  ? I quite like the second.

hramezani:
@benoitc, thanks, I rename the config with the first one `--disable-redirect-access-to-syslog`

benoitc:
@hramezani thanks!

PR #1529, opened 2017-06-15
doc: Keepalive missing in sync worker and value recommendation

tuukkamustonen:
Related to: https://github.com/benoitc/gunicorn/issues/1194

As per https://github.com/benoitc/gunicorn/issues/1194#issuecomment-239917723 am I'm correct in stating that `sync` worker just doesn't support keep-alive, even if you try setting the header manually?

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -34,6 +34,10 @@ as any errors will affect at most a single request. Though as we describe below
 only processing a single request at a time requires some assumptions about how
 applications are programmed.
 
+``sync`` worker does not support persistent connections - each connection is
+closed after response has been sent (even if you manually add ``Keep-Alive``
+or ``Connection: keep-alive`` header in your application).
+
 Async Workers
 -------------
 
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -234,7 +234,14 @@ keepalive
 
 The number of seconds to wait for requests on a Keep-Alive connection.
 
-Generally set in the 1-5 seconds range.
+Generally set in the 1-5 seconds range for servers with direct connection
+to the client (e.g. when you don't have separate load balancer). When
+Gunicorn is deployed behind a load balancer, it often makes sense to
+set this to a higher value.
+
+.. note::
+   ``sync`` worker does not support persistent connections and will
+   ignore this option.
 
 Security
 --------
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -760,7 +760,14 @@ class Keepalive(Setting):
     desc = """\
         The number of seconds to wait for requests on a Keep-Alive connection.
 
-        Generally set in the 1-5 seconds range.
+        Generally set in the 1-5 seconds range for servers with direct connection
+        to the client (e.g. when you don't have separate load balancer). When
+        Gunicorn is deployed behind a load balancer, it often makes sense to
+        set this to a higher value.
+
+        .. note::
+           ``sync`` worker does not support persistent connections and will
+           ignore this option.
         """
 
 


docs/source/settings.rst
```diff
@@ -234,7 +234,13 @@ keepalive
 
 The number of seconds to wait for requests on a Keep-Alive connection.
 
-Generally set in the 1-5 seconds range.
+Generally set in the 1-5 seconds range for servers with direct connection
```
berkerpeksag: Could you make this change in `gunicorn/config.py` instead and then run `make html` in `docs/` directory? (After running `make html` you will need to strip unrelated changes in `docs/source/settings.rst`.)

tuukkamustonen: Ah, it gets copied from there... alright, thanks for the guidance... I adjusted PR accordingly.

tilgovi:
Seems fine to me.

docs/source/design.rst
```diff
@@ -34,6 +34,10 @@ as any errors will affect at most a single request. Though as we describe below
 only processing a single request at a time requires some assumptions about how
 applications are programmed.
 
+The sync worker does not support keep-alive - each connection is closed after
```
berkerpeksag: As a non-native speaker, I'd prefer "persistent connection" instead of keep-alive. On another hand, we already have a  `--keep-alive` setting so perhaps it's OK. @tilgovi, what do you think?

docs/source/design.rst
```diff
@@ -34,6 +34,10 @@ as any errors will affect at most a single request. Though as we describe below
...
+response has been sent (even if you manually add ``Keep-Alive`` header in
```
berkerpeksag: Is `Keep-Alive` a standard now? We could probably mention `Connection: keep-alive` too.

docs/source/settings.rst
```diff
@@ -234,7 +234,13 @@ keepalive
...
+to the client (=when you don't have separate load balancer). When
```
berkerpeksag: Perhaps replace '=' with 'i.e.'?

docs/source/settings.rst
```diff
@@ -234,7 +234,13 @@ keepalive
...
+gunicorn is deployed behind a load balancer, it often makes sense to
```
berkerpeksag: gunicorn -> Gunicorn

docs/source/settings.rst
```diff
@@ -234,7 +234,13 @@ keepalive
...
+set this to a higher value.
+
+Note that ``sync`` worker does not support keep-alive and will ignore
```
berkerpeksag: We could use a note directive here:

```rst
.. note::
   ...
```

docs/source/design.rst
tilgovi: I defer to you, since English is my first language. Your judgement is more helpful here.

docs/source/design.rst
tilgovi: That's correct.

tilgovi: It is `Connection: keep-alive` as far as I know.

docs/source/settings.rst
tilgovi: Or 'e.g.', since it might be something other than a load balancer (like an app server that has a connection to another backend service that uses gunicorn).

tilgovi:
Great comments, @berkerpeksag!

tuukkamustonen:
@tilgovi @berkerpeksag Sorry for the quite drastic delay... the PR has been updated accordingly.

docs/source/design.rst
```diff
@@ -34,6 +34,10 @@ as any errors will affect at most a single request. Though as we describe below
 only processing a single request at a time requires some assumptions about how
 applications are programmed.
 
+The sync worker does not support peristent connections - each connection is
```
berkerpeksag: peristent -> persistent

gunicorn/config.py
```diff
@@ -760,7 +760,14 @@ class Keepalive(Setting):
...
         The number of seconds to wait for requests on a Keep-Alive connection.
 
-        Generally set in the 1-5 seconds range.
+        Generally set in the 1-5 seconds range for servers with direct connection
+        to the client (e.g. when you don't have separate load balancer). When
+        Gunicorn is deployed behind a load balancer, it often makes sense to
+        set this to a higher value.
+
+        .. note::
+           ``sync`` worker does not support keep-alive and will ignore this
```
berkerpeksag: We could use the same wording in design.rst and say 

```rst
``sync`` worker does not support persistent connections [...]
```

tuukkamustonen:
Oopsie. Maybe now?

berkerpeksag:
Looks great, thank you for improving the documentation! :)

PR #1541, opened 2017-07-11
Only use inotify on linux

pjvandehaar:


gunicorn/reloader.py
```diff
@@ -53,12 +53,14 @@ def run(self):
...
             time.sleep(self._interval)
 
-try:
-    from inotify.adapters import Inotify
-    import inotify.constants
-    has_inotify = True
-except ImportError:
-    has_inotify = False
+has_inotify = False
+if sys.platform == 'linux':
```
berkerpeksag: We still support Python 3.2 so

```py
if sys.platform.startswith('linux'):
    ...
```

would be better.

pjvandehaar: Good catch.

tilgovi:
LGTM

berkerpeksag:
Thanks!

Issue #1530, opened 2017-06-19
syslog unix domain socket type assumption


guj-zz:
Using syslog_addr='unix:///dev/log' on Linux leads to a crash due to unhandled exception in _connect_unixsocket self.socket.connect:

> socket.error: [Errno 91] Protocol wrong type for socket

This is caused by the assumption AF_UNIX -> SOCK_STREAM in parse_syslog_address:
`if addr.startswith("unix://"): sock_type = socket.SOCK_STREAM`
At least on Linux, Unix domain sockets may be, and in the case of syslog, actually are served by daemons expecting SOCK_DGRAM. Python 2.7 SysLogHandler does the right thing:

> For Unix sockets, you can also specify a socktype of None, in which case socket.SOCK_DGRAM will be used, falling back to socket.SOCK_STREAM.

I suggest defaulting sock_type to None, see https://github.com/guj-zz/gunicorn/commit/af92742143605e095d080e776579852394393937 (can send PR). Thanks.

tilgovi:
Sounds great. I would be happy to review a PR.

guj-zz:
Thank you. Here is #1547,  a trivial change that should avoid the problem.

guj-zz:
Hello Randall.

Thank you for your reply.

On 14/07/2017 20:22, Randall Leeds wrote:
> Sounds great. I would be happy to review a PR.

Thanks. I created PR #1547 and added a comment on Issue #1530.

Cheers,
Steve

tilgovi:
Closed by #1547

PR #1547, opened 2017-07-17
parse_syslog_addr sock_type default

guj-zz:
Default to sock_type None, unless explictly overridden, and explain how this covers both options

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -119,10 +119,12 @@ class SafeAtoms(dict):
 
 def parse_syslog_address(addr):
 
+    # unix domain socket type depends on backend
+    # SysLogHandler will try both when given None
     if addr.startswith("unix://"):
-        sock_type = socket.SOCK_STREAM
+        sock_type = None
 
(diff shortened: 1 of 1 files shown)


tilgovi:
It sounds like this is correct, but I don't know. Can you point me to any documentation or code links in the syslog implementation?

tilgovi:
Oh, right, this is #1530. I think this looks good.

guj-zz:
Hello, Randall.

On 8/7/2017 2:14 AM, Randall Leeds wrote:
> tilgovi approved this pull request.

Thank you. In case this is needed for documentation purposes, the
comment I was referring to is in logging.py, currently at
https://github.com/python/cpython/blob/master/Lib/logging/handlers.py#L806
HTH

Sincerely,
Stephan

-- 

Gruner+Jahr AG & Co KG
Informationstechnologie
IT-Betrieb / Anwendungsservices
20459 Hamburg
Am Baumwall 11
Telephon +49-40-3703-3177

tilgovi:
Thanks so much!

PR #1551, opened 2017-07-22
Remove redundant 'is not None'

tomjaguarpaw:


tilgovi:
I take that back. We want to be sure it's non-empty, too. So, I'll go with what you sent. Thank you!

Issue #1567, opened 2017-08-08
Why restrict access logging?


tomjaguarpaw:
I want to create a new access logger with

    handler = ...
    logger = logging.getLogger('gunicorn.access')
    logger.addHandler(handler)

But this doesn't work!  The reason it doesn't work is that none of `self.cfg.accesslog`, `self.cfg.logconfig` or `self.cfg.syslog` are set.  I'm very puzzled by [the check for the existence of these attributes in the `Logger` class](https://github.com/benoitc/gunicorn/blob/fc0d9734168ee675723ce09af26b3c8c37fec534/gunicorn/glogging.py#L319).  In fact, this check has existed in some form since [the beginning of this class](https://github.com/benoitc/gunicorn/blame/824801d01768c984d7f6cbd2968cecffc46b0472/gunicorn/glogging.py#L85).

Why restrict access logging to only the cases where one of those attributes is set?  Is there a good reason?  I can't think of one.  I propose deleting that check.

tomjaguarpaw:
Specifically, I propose https://github.com/benoitc/gunicorn/pull/1568

tilgovi:
I think this makes sense to me. The check does seem unnecessary. The only thing I can think that we would want to check is that we still set the access log not to propagate even if no options configure access logging. Otherwise, we will bubble into the `gunicorn` logger, and that seems to be against the intent of the way the access log is used. In other words,` we use the `gunicorn.error` and `gunicorn.access` loggers as purpose-named rather than module-named loggers.

I am not a big fan of a logger with an `access` method, though, and I hesitate to encourage using it as public interface. Can you say more about why you are trying to log to the access log from somewhere else other than Gunicorn internals?

tomjaguarpaw:
> Can you say more about why you are trying to log to the access log from somewhere else other than Gunicorn internals?

What do you mean?  I'm not calling `Logger.access` myself.  I just  want to set up a `logging` logger that logs access to my Gunicorn server.  How else would I do that?

tilgovi:
Gunicorn sets up that logger and does that logging internally. You should be able to use the logging settings to get what you want.

http://docs.gunicorn.org/en/latest/settings.html#logging

tomjaguarpaw:
Yeah but I want more control over logging than that.  I want to be able to log to `sys.stdout`, for example. Indeed I want as much control as the `logging` module provides.  Why would GUnicorn restrict me in that?  I cannot see a reason.

[...]

PR #1568, opened 2017-08-08
Remove check for logging attributes

tomjaguarpaw:
We may still want to log access even if none of `self.cfg.accesslog`, `self.cfg.logconfig` or `self.cfg.syslog` are set.

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -316,9 +316,6 @@ class Logger(object):
         for format details
         """
 
-        if not (self.cfg.accesslog or self.cfg.logconfig or self.cfg.syslog):
-            return
-
         # wrap atoms:
         # - make sure atoms will be test case insensitively
         # - if atom doesn't exist replace it by '-'


tomjaguarpaw:
I don't know what to propose regarding the test failure.  Perhaps use `environ.get` instead of `environ[]`?  In any case, this seems to be an artefact of the test setup, not of my patch.

tilgovi:
Can you explain more what you mean? Are you calling the access method yourself somewhere?

tilgovi:
I'm sorry. I see your issue in #1567 now. I will look there and comment if I need to.

berkerpeksag:
> I don't know what to propose regarding the test failure. Perhaps use environ.get instead of environ[]? In any case, this seems to be an artefact of the test setup, not of my patch.

I think the problem is in the test code here. We should pass a proper `environ` dict instead of `{}` in line 92:

```py
logger.access(SimpleNamespace(status="200 OK"), None, {}, timedelta(seconds=7))
```

You can take a look at `tests/test_logger.py` for an example:

https://github.com/benoitc/gunicorn/blob/fc0d9734168ee675723ce09af26b3c8c37fec534/tests/test_logger.py#L40-L45

tilgovi:
@tomjaguarpaw lets get the tests passing and we can definitely merge this.

benoitc:
access *must* be disabled by default. 

If the idea is to allow someone to add its own way to output the access log which makes sense, then should rather have a way tell to gunicorn how to do it.  but by default it should always be disabled.

I also think we shouldn't test for each logger, that was an hack for the time where we only supported 2 log systems. Let's discuss that on #1572 ?

benoitc:
to be clear this patch can't be merged as is since it will let access logs go threw the logging system which has a performance impact on gunicorn. 

People using it behind a proxy  generally don't need them

NicolasLM:
> to be clear this patch can't be merged as is since it will let access logs go threw the logging system which has a performance impact on gunicorn.

The Python docs give a nice solution to this problem:

```python
    def access(self, resp, req, environ, request_time):
        """ See http://httpd.apache.org/docs/2.0/logs.html#combined
        for format details
        """

        if not self.access_log.isEnabledFor(logging.INFO):
            return
```

The result of `isEnabledFor()` can even be cached to avoid the extra overhead. See  https://docs.python.org/3/howto/logging.html#optimization

aconrad:
Please rebase off of master for coverage tests to pass.

tomjaguarpaw:
I'm not sure my PR makes sense any more.  I haven't really been able to understand what the proposed solution is, or even if there is one, although [@NicholasLM's suggestion](https://github.com/benoitc/gunicorn/pull/1568#issuecomment-348451726) sounds good.

tilgovi:
I think the suggestion from @NicolasLM is good, too.

tilgovi:
I propose we re-consider this for master only (Python 3 only) and replace the check with a call to `self.access_log.hasHandlers`[1]. Since we set `propagate` to `False`, this check will only check for handlers on the access logger itself. Since we really only use one level (`INFO`) for access log messages, I think we can expect that if there is a handler at all it will handle `INFO` and expect that users that do not wish to have an access log will not set up handlers for it.

[1] https://docs.python.org/3/library/logging.html#logging.Logger.hasHandlers

PR #1578, opened 2017-08-25
http/message: Split request line as bytes to avoid splitting on 0x0A. Fixes #1577

### LINKED ISSUE #1577: Bad Request when sending utf-8 encoded http path under python3
This is the sent data:
```
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.connect(('127.0.0.1', 8080))
print(s.send('GET /à%20k HTTP/1.0\r\n\r\n'.encode('utf-8')))
print(s.recv(100000).decode('ascii'))
```

This is the response
```
HTTP/1.1 400 Bad Request
Connection: close
Content-Type: text/html
Content-Length: 181

<html>
  <head>
    <title>Bad Request</title>
  </head>
  <body>
    <h1><p>Bad Request</p></h1>
    Invalid HTTP Version 'Invalid HTTP Version: '%20k HTTP/1.0''
  </body>
</html>
```

This is because the request line is first decoded as latin1 using the _compat:bytes_to_str this causes the "à" to be returned as "\xc3\xa0", then the request line is split using line.split(None, 2) which will consider the \xa0 (non breaking space) as whitespace and strip it, thus rendering the request line invalid.

A first attempt would be to use line.split(' ', 2) but then the split will no longer eat up all consecutive whitespaces and may introduce other bugs.

I'm not sure what would be the best solution here.

rslinckx:


tests/treq.py
```diff
@@ -78,8 +78,11 @@ def send_lines(self):
             yield lines
 
     def send_bytes(self):
-        for d in str(self.data.decode("latin1")):
-            yield bytes(d.encode("latin1"))
+        for d in self.data:
+            if six.PY3:
+                yield bytes([d])
```
tilgovi: I don't think the cast is necessary anymore, since we open the file in binary mode.

tilgovi:
Looks good. Thank you!

tests/treq.py
rslinckx: I'm not sure what was the intention there, but it's reading the data byte-by-byte and yielding byte-by-byte. 

In py2, iterating the data str will yield individual bytes as str
In py3, iterating the data bytes will yield individual bytes as int, which then need to be returned as new bytes using bytes([d])

Previous code was just plain wrong under py2 as str(self.data.decode("latin1")) will try to re-encode an unicode string in ascii

tilgovi: Thanks for the explanation.

tilgovi:
@berkerpeksag, @benoitc do you see any problem with this?

benoitc:
@rslinckx thanks

@tilgovi merged :)

PR #1581, opened 2017-09-02
Fix #1576 - config file/module in GUNICORN_CMD_ARGS

### LINKED ISSUE #1576: --config doesn't work in GUNICORN_CMD_ARGS
Specifying `--config` in the `GUNICORN_CMD_ARGS` environment variable quietly fails as config file loading only happens when it is passed as an argument:
https://github.com/benoitc/gunicorn/blob/328e509260ae70de6c04c5ba885ee17960b3ced5/gunicorn/app/base.py#L137-L175

Code0x58:
I also sorted THANKS.md to follow CONTRIBUTING.md.

### DIFF
--- a/gunicorn/app/base.py
+++ b/gunicorn/app/base.py
@@ -147,26 +147,26 @@ class Application(BaseApplication):
             for k, v in cfg.items():
                 self.cfg.set(k.lower(), v)
 
+        env_args = parser.parse_args(self.cfg.get_cmd_args_from_env())
+
         if args.config:
             self.load_config_from_file(args.config)
+        elif env_args.config:
+            self.load_config_from_file(env_args.config)
         else:
             default_config = get_default_config_file()
             if default_config is not None:
                 self.load_config_from_file(default_config)
 
         # Load up environment configuration
-        env_vars = self.cfg.get_cmd_args_from_env()
-        if env_vars:
-            env_args = parser.parse_args(env_vars)
-            for k, v in vars(env_args).items():
-                if v is None:
-                    continue
-                if k == "args":
-                    continue
-                self.cfg.set(k.lower(), v)
+        for k, v in vars(env_args).items():
+            if v is None:
+                continue
+            if k == "args":
+                continue
+            self.cfg.set(k.lower(), v)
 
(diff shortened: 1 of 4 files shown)


gunicorn/app/base.py
```diff
@@ -147,26 +147,27 @@ def load_config(self):
             for k, v in cfg.items():
                 self.cfg.set(k.lower(), v)
 
-        if args.config:
-            self.load_config_from_file(args.config)
+        env_args = parser.parse_args(self.cfg.get_cmd_args_from_env())
+
+        if args.config or env_args.config:
```
tilgovi: Should one override the other rather than allowing both of these branches?

Code0x58:
I can kind of understand why overriding would make sense, but inheriting feels more natural to me.

After looking at this more I've realised the hierarchy feels a bit weird (as it was before the patch) as options from the config file on the command line are overridden by arguments from the environment:
  app config ← ((env config file ← arg config file) | default config file) ← env options ← arg options

This also gives the issue that applications don't get augments from the environment, so for applications to get configured properly `Application.init` has to be after all of the config loading.

_Blerg._

berkerpeksag:
Some comments:

* Please leave changes to THANKS file for another PR. You can add your name in this PR.
* There is no need to rename tests/config/test_cfg.py to tests/config/test_cfg1.py.

gunicorn/app/base.py
berkerpeksag: I agree that `env_args.config` should override `args.config`. We should probably document this at http://docs.gunicorn.org/en/latest/settings.html#settings

tilgovi:
I think this looks good to merge! Thank you so much, @Code0x58!

tilgovi:
And thanks for your patience. I had my personal laptop melt down last weekend, so I've been a little bit offline more than usually.

Code0x58:
Don't worry, I was delayed too as I was on a mobile for 5 days.

Is it worth me putting in a separate PR to either sort the THANKS, or remove the sorting part from CONTRIBUTING?

tilgovi:
I'm fine to keep it sorted.

Issue #1564, opened 2017-08-06
Separate instrumentation and logging


tilgovi:
Issues like #1405 demonstrate that having statsd instrumentation built on top of the logging handlers means instrumentation and logging are coupled. Some have found this coupling surprising.

ykusakabe:
Hello @tilgovi 

I made decision for the PR above after considering:
- making instrumentation a log handler, which I didn't feel natural as handler.emit() gets LogRecord created as argument (not necessary for gauge/counter-type and not sufficient for access-kind histogram)
- making instrumentation a mixin class in a logger class, which still means the coupling between logger and instrumentation, and also makes sub-classing it necessary for custom logger. (ambiguous config structure)

This fix makes instrumentation directly from config setting no matter what logger is being loaded. I believe this config structure makes it easy to add more instrumentation in the future as well.

Do you think it's something workable? I'm open to any other idea as well :) Thanks!

tilgovi:
I think it's a very reasonable approach. Thank you for laying out your thoughts so well, for taking care of backward compatibility, and for adding configuration tests. I have a few comments on the PR, but I'll wait and make them as a review on the PR issue if we agree this is the way to go.

@benoitc @berkerpeksag how do you feel about this?

benoitc:
i would lik to create a project (in github sense) for it and collect the ticket there.

About the implementation, I would like to simply implement some hooks that can be used by any implementation to add their own logic. And then support by default opentelemetry. Thoughts?

calebwoofenden:
I would love to see some solution to this implemented. The statsD metrics that are exported right now are useful, but there is a lot of other information that I would like to be able to see. It would be good if I could easily write some code/config to be able to do that.

ykusakabe:
As of now, the PR of my proposal technically can handle any number of custom instrumentation classes mixed into logger class. Though statsD is only available at moment (I didn't wanna change the existing config flag of `statsd_host`), it sort of works as "hooks" to add any instrument class with methods matching with logger.

However, though it can support multiple custom instrumentation classes, this approach still means coupling between logger and instrumentation since instrumentation methods getting mixed into logger methods.
So maybe a decision can be made on whether we completely separate the way to call instrumentation from logger, and for that, maybe we can get started from what log data and metrics data would eventually be provided by gunicorn? As of now for statsD instrumentation looks sufficient to be coupled with logger to me.

paullaffitte:
[...]

PR #1583, opened 2017-09-06
Fix Separate Instrumentation

ykskb:
Proposed fix for #1564 on separation of instrumentation from logging by dynamically wrapping instrumentation methods and logging methods.
This allows multiple instrumentation possible if desired, loading from newly added config property, `instrumentation_classes`.


gunicorn/config.py
```diff
@@ -177,6 +171,18 @@ def logger_class(self):
...
 
     @property
+    def instrumentation_classes(self):
+        inst_class_uris = []
+        inst_classes = []
+        if 'statsd_host' in self.settings and self.settings['statsd_host'].value is not None:
+            inst_class_uris.append("gunicorn.instrument.statsd.Statsd")
+
+        for uri in inst_class_uris:
+            inst_classes.append(util.load_class(uri))
```
wking: Probably not a big deal, and you'll have to update your test to use:

    c.instrumentation_classes.next()

instead of your current:

    c.instrumentation_classes[0]

but this feels like it should be an iterator to me, instead of collecting arrays.  Something like:

```python
if 'statsd_host' in self.settings and self.settings['statsd_host'].value is not None:
    yield util.load_class(uri)
```

would cover your whole method.

gunicorn/arbiter.py
```diff
@@ -92,6 +92,8 @@ def setup(self, app):
         self.cfg = app.cfg
 
         if self.log is None:
+            if self.cfg.instrumentation_classes:
+                glogging.add_instrumentation(self.cfg)
```
wking: I'd just call `glogging.add_instrumentation(self.cfg)` instead of conditioning it on `self.cfg.instrumentation_classes` here.  If `cfg.instrumentation_classes` returns no classes, the `add_instrumentation` call will just be a no-op.

gunicorn/glogging.py
```diff
@@ -328,7 +329,7 @@ def access(self, resp, req, environ, request_time):
             request_time))
 
         try:
-            self.access_log.info(self.cfg.access_log_format, safe_atoms)
+            self.access_log.info(self.cfg.access_log_format, dict(safe_atoms))
```
wking: Why are you adding a `dict(…)` wrapper here?

gunicorn/glogging.py
```diff
@@ -458,3 +459,33 @@ def _get_user(self, environ):
...
         return user
+
+
+def add_instrumentation(cfg):
+    insts = []
+    for inst_class in cfg.instrumentation_classes:
+        insts.append(inst_class(cfg))
+
+    for inst in insts:
+        add_inst_methods(cfg.logger_class, inst)
```
wking: No need to double up on the loops here.  You can just use:

```python
for inst_class in cfg.instrumentation_classes:
    add_inst_methods(cfg.logger_class, inst)
```

gunicorn/glogging.py
```diff
@@ -458,3 +459,33 @@ def _get_user(self, environ):
...
+
+def _wrap_log_method(log_method, inst_method):
+    def wrap_func(*args, **kwargs):
+        try:
+            log_method(*args, **kwargs)
+            inst_method(*args, **kwargs)
+        except TypeError:
+            if sys.stderr:
+                print("Error: Argument mismatch between logger and instrumentation", file=sys.stderr)
+                sys.stderr.flush()
```
wking: This makes me nervous.  What I'd really like is an error-level log sent to wherever error logs are going.  Are we really sure [we can't make instrumentation work as a `Handler`][1]?

[1]: https://github.com/benoitc/gunicorn/issues/1564#issuecomment-327563265

gunicorn/glogging.py
berkerpeksag: @ykusakabe thank you for working on this! We are going to drop Python 2 support pretty soon. See #1764 for details.

berkerpeksag: Yes, you can safely remove that and the other Python 2-related changes (`__future__` imports etc.) in this PR.

lelabo-marc:
Any news about this fix, #1564 seems to said that this fix was promising ?

ykusakabe:
I had updated this fix according to the feedbacks above, but haven't heard much since then.
I think we need maintainers' & contributors' feedback on this in terms of 
- architecture: if this is really the way to fix or not
- releasing: this fix includes new config value

I'm happy to update the codes to be able to merge once there's an agreement :)

Issue #1584, opened 2017-09-06
Can't boot eventlet workers with eventlet 0.21.0


jwg4:
Eventlet workers fail to boot using the most recent release of eventlet:

To reproduce this I am using the minimal app from http://docs.gunicorn.org/en/latest/run.html

Error message:
```
(venv) jack@neptune-vm:~/z/test_gunicorn$ gunicorn -b 0.0.0.0:5015 -w 4 myapp:app -k eventlet
[2017-09-06 17:40:13 +0000] [8760] [INFO] Starting gunicorn 19.7.1
[2017-09-06 17:40:13 +0000] [8760] [INFO] Listening at: http://0.0.0.0:5015 (8760)
[2017-09-06 17:40:13 +0000] [8760] [INFO] Using worker: eventlet
[2017-09-06 17:40:13 +0000] [8765] [INFO] Booting worker with pid: 8765
[2017-09-06 17:40:13 +0000] [8765] [ERROR] Exception in worker process
Traceback (most recent call last):
  File "/home/jack/z/test_gunicorn/venv/local/lib/python2.7/site-packages/gunicorn/arbiter.py", line 578, in spawn_worker
    worker.init_process()
  File "/home/jack/z/test_gunicorn/venv/local/lib/python2.7/site-packages/gunicorn/workers/geventlet.py", line 102, in init_process
    self.patch()
  File "/home/jack/z/test_gunicorn/venv/local/lib/python2.7/site-packages/gunicorn/workers/geventlet.py", line 91, in patch
    hubs.use_hub()
  File "/home/jack/z/test_gunicorn/venv/local/lib/python2.7/site-packages/eventlet/hubs/__init__.py", line 70, in use_hub
    mod = get_default_hub()
  File "/home/jack/z/test_gunicorn/venv/local/lib/python2.7/site-packages/eventlet/hubs/__init__.py", line 38, in get_default_hub
    import eventlet.hubs.epolls
  File "/home/jack/z/test_gunicorn/venv/local/lib/python2.7/site-packages/eventlet/hubs/epolls.py", line 27, in <module>
    from eventlet.hubs.hub import BaseHub
  File "/home/jack/z/test_gunicorn/venv/local/lib/python2.7/site-packages/eventlet/hubs/hub.py", line 23, in <module>
    from eventlet.support import greenlets as greenlet, clear_sys_exc_info, monotonic, six
[...]

PR #1585, opened 2017-09-06
Check all the worker pids instead of every child.

jwg4:
This solves a problem where this wait is triggered by something done on a child process launched by a third-party library which we use when setting up workers.

This seems to be a reasonable fix, we only need to wait on the pids of the workers, and not any other arbitrary child pid.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -508,7 +508,7 @@ class Arbiter(object):
         """\
         Reap workers to avoid zombie processes
         """
-        for pid in tuple(self.WORKERS):
+        for pid in self.WORKERS.keys():
             try:
                 wpid, status = os.waitpid(pid, os.WNOHANG)
                 if self.reexec_pid == wpid:
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -511,8 +511,6 @@ class Arbiter(object):
         for pid in tuple(self.WORKERS):
             try:
                 wpid, status = os.waitpid(pid, os.WNOHANG)
-                if not wpid:
-                    continue
                 if self.reexec_pid == wpid:
                     self.reexec_pid = 0
                 else:
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -508,11 +508,11 @@ class Arbiter(object):
         """\
         Reap workers to avoid zombie processes
         """
-        try:
-            while True:
-                wpid, status = os.waitpid(-1, os.WNOHANG)
+        for pid, _ in self.WORKERS.items():
+            try:
+                wpid, status = os.waitpid(pid, os.WNOHANG)
                 if not wpid:
-                    break
+                    continue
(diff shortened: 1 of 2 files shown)


gunicorn/arbiter.py
```diff
@@ -508,11 +508,11 @@ def reap_workers(self):
         """\
         Reap workers to avoid zombie processes
         """
-        try:
-            while True:
-                wpid, status = os.waitpid(-1, os.WNOHANG)
+        for pid, _ in self.WORKERS.items():
```
temoto: No need in `.items()` since you don't use dictionary value.

`for pid in tuple(self.WORKERS):` might be a better line. tuple is to force copy of pids, because `WORKERS` dict might change in the meantime.

jwg4: Used, thanks.

RonRothman: The idiomatic way to iterate over the keys of a possibly-changing dict would be:

```
for pid in self.WORKERS.keys():
```

Why not use that? It's much clearer what you're trying to do.

RonRothman:
Just checking: Is there a possible race condition where the arbiter would have tried to spin up a new worker process before `WORKERS` was updated?

Also, just curious: what problem is this PR intended to solve?

Thanks!

benoitc:
About the patch looking only for the known pids at the time is not really reliable on all systems, it's quite better to wait for all child processes that can exist and filter them which what does the current code.  also the result of waitpid might be (0,0) on am system so this value need to checked forts.

@RonRothman i guess it's related to https://github.com/benoitc/gunicorn/issues/1584

@jwg4 see my comment on the ticket

temoto:
There is race condition, one way to fix it is to make `WORKERS` a `threading.Queue`.

jwg4:
@temoto I am not sure if this would fix the race condition. When a new worker forks couldn't it conceivably die before adding itself to `WORKERS`. This could happen whether or not `WORKERS` itself was thread-safe.

temoto:
@jwg4 worker could not add itself because it's a separate process, it can't modify master process memory. Consider code:

```
# in master process
pid = os.fork()  # or subprocess.Popen()
# if pid > 0
workers_queue.put(pid)
```

jwg4:
Yes, I miss the check for the return value. However, doesn't the same objection apply? Couldn't the child thread die before the main thread has had a chance to modify the list of workers?

temoto:
Yes, you are right. One way to fix that is to check whether new pid is still running after `workers_queue.put()`. Gut tells it's getting too complicated for a reliable production system.

benoitc:
@jwg4 this can't happen though, signals are queued so it will be handled at some point. Anyway I think the fix I proposed to the related issue is enough. It makes sure that signals handlers are resetted after the worker is spawned and before eventlet is setup.

jwg4:
I agree @benoitc your fix is more straightforward and less uncertain.

benoitc:
@jwg4 OK, i will make release on thursday and send a pr later tonight.

PR #1589, opened 2017-09-15
Tweaks

Code0x58:
Some trivial tweaks to things I saw while going through code

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -31,12 +31,6 @@ KNOWN_SETTINGS = []
 PLATFORM = sys.platform
 
 
-def wrap_method(func):
-    def _wrapped(instance, *args, **kwargs):
-        return func(*args, **kwargs)
-    return _wrapped
-
-
 def make_settings(ignore=None):
     settings = {}
     ignore = ignore or ()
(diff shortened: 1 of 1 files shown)


tilgovi:
Looks good to me.

tilgovi:
Thanks!

gunicorn/config.py
```diff
@@ -1510,9 +1496,10 @@ class OnStarting(Setting):
     validator = validate_callable(1)
     type = six.callable
 
-    def on_starting(server):
+    @staticmethod
```
berkerpeksag: This is actually wrong. It was written `default = staticmethod(some_callable)` because `some_callable` was used to document the name and the signature of the hook.

For example, take a look at the `on_starting` documentation.

Before this change:

```py
def on_starting(server):
    pass
```

See http://docs.gunicorn.org/en/stable/settings.html#on-starting for details.

After this change:

```py
@staticmethod
def default(server):
    pass
```

See http://docs.gunicorn.org/en/latest/settings.html#on-starting for details.

I think this PR needs to be reverted.

tilgovi: My apologies. I suspected there might be a reason but I couldn't see it. I should have asked for another pair of eyes. Feel free to revert.

Issue #1590, opened 2017-09-16
Consistently use Python's warnings module and remove `utils.warn`


Code0x58:
There is a `util.warn` function which prints to stderr with some fanfare which is only used to say that two pasterapp functions are deprecated in favour on `gunicorn --paste $config`. These should use a [`DeprecationWarning`](https://docs.python.org/2/library/warnings.html#warning-categories).
```
$ ag "warn\(" gunicorn
gunicorn/argparse_compat.py
1571:            warnings.warn(
2306:        warnings.warn(
2332:        warnings.warn(

gunicorn/util.py
206:        warnings.warn('tests may fail, delete still pending for ' + pathname,
524:def warn(msg):

gunicorn/app/pasterapp.py
178:    util.warn("""This command is deprecated.
202:    util.warn("""This command is deprecated
```

berkerpeksag:
`DeprecationWarning`s are silent by default and since most users don't use Gunicorn for local development and don't set `-Wd` on production, I prefer our more explicit `util.warn` helper.

tilgovi:
Even better might be to get the warning into the Gunicorn log, wherever that is.

We could change the two places where we do this in the pasterapp code to set `app = Paster...()` and then call `app.logger.warn()` before `app.run()`.

Code0x58:
I was thinking that warnings on stderr would usually just go nowhere, or if done well then into a log file somewhere if the service's (error) output is written anywhere.

I think this would be an improvement for people who use [`logging.captureWarnings(True)`](https://docs.python.org/2/library/logging.html#logging.captureWarnings) along with `logging.simplefilter("default")` and specifying a _py.warnings_ handler.

I'm not a huge fan of it going to `gunicorn.error` either, although it is simple enough to set the parent of `py.warnings` to be `gunicorn.error` so that it propagates to it.

I think it would be better practice to configure `py.warnings` for the user so that it goes into it's own handler/log file (maybe requiring a flag to configure it). A lot of libraries/frameworks use it (including code gunicorn uses), and I think the world would be a better place if people used it or were even aware of it more. Something like `--warning-log=$log_file` sounds like a good idea to me - I do this kind kind of thing with self rolled logging config, which probably explains why I didn't put the flag in to begin with.

tilgovi:
All very interesting. I do not know yet what the best approach is. This is the first time I've learned about the warnings module, myself.

I like using a standard mechanism for warnings. I dislike adding more CLI options.

[...]

PR #1591, opened 2017-09-16
#1590 Use warnings module to state deprecations

Code0x58:


Code0x58:
*cleaning up* I'm not inclined to add a test for this, even a shill to increase coverage. Can this be reviewed and the coverage bypassed if passed?

benoitc:
Do we really need to tell the version ? For me it suffices to say it will be deprecated. So the plan can be changed...

Code0x58:
Not necessarily, I would have been imitating Django's codebase which supports their [policy](https://docs.djangoproject.com/en/dev/internals/release-process/#deprecation-policy).

berkerpeksag:
https://github.com/benoitc/gunicorn/issues/1590 has already been closed. Thank you for the PR!

PR #1593, opened 2017-09-16
Update Debian installation instructions

tnir:
Update Debian installation instructions as two years passed after #1112 and 'stretch' Debian 9 was released.


berkerpeksag:
Thanks for updating the installation documentation!

PR #1597, opened 2017-09-17
Fix reST format in docs

tnir:
- Improve `docs/gunicorn_ext.py` for empty string as default value
- Regenerate `docs/source/settings.rst` from `gunicorn/config.py` via `make -C docs html`
- Fix reST format in docs to fix broken docs

  - http://docs.gunicorn.org/en/latest/2015-news.html
  - http://docs.gunicorn.org/en/latest/2016-news.html
  - http://docs.gunicorn.org/en/latest/settings.html#disable-redirect-access-to-syslog


docs/source/settings.rst
```diff
@@ -826,7 +826,7 @@ statsd_prefix
 ~~~~~~~~~~~~~
 
 * ``--statsd-prefix STATSD_PREFIX``
-* ````
+* ``None``
```
berkerpeksag: `docs/source/settings.rst` is automatically generated by `docs/gunicorn_ext.py` and it extracts all information from `gunicorn/config.py`. In this case, the default value of `statsd_prefix` is an empty string:

https://github.com/benoitc/gunicorn/blob/4c3ee6f4ec304722acb40d113fc4566f33bf3a6e/gunicorn/config.py#L1422

I think the best way to fix this would be to make `docs/gunicorn_ext.py` return `(empty string)` instead of \`\`\`\`.

docs/source/settings.rst
```diff
@@ -619,7 +619,7 @@ The Access log file to write to.
 .. _disable-redirect-access-to-syslog:
 
 disable-redirect-access-to-syslog
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
berkerpeksag: This one is interesting. Do you have time to debug `fmt_setting` function? The relevant code is here:

https://github.com/benoitc/gunicorn/blob/4c3ee6f4ec304722acb40d113fc4566f33bf3a6e/docs/gunicorn_ext.py#L66

tnir: @berkerpeksag Thanks. I don't think `fmt_setting` is broken. `make -C doc` was not used to make Commit 919871d (part of #1528) as confirmed below:

```
$ git show --oneline
4c3ee6f Merge pull request #1589 from Code0x58/tweaks
$ make -C docs html
[...]
```
and then `git diff docs/source/settings.rst` returns:
```diff
diff --git a/docs/source/settings.rst b/docs/source/settings.rst
index 3aab716..3fe7f54 100644
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
[...]
@@ -618,8 +623,8 @@ The Access log file to write to.

 .. _disable-redirect-access-to-syslog:

-disable-redirect-access-to-syslog
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+disable_redirect_access_to_syslog
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

 * ``--disable-redirect-access-to-syslog``
 * ``False``
@@ -903,7 +908,8 @@ on_starting
[...]
```

docs/source/settings.rst
tnir: OK

tnir:
@berkerpeksag Thanks for your review. I just updated the PR with passed tests.

berkerpeksag:
Thanks!

PR #1600, opened 2017-09-17
Enable pycodestyle in prospector CI

### LINKED ISSUE #1599: TEST: enable pycodestyle in CI
While enabled in #1453 (#947) through introduction of lint and then disabled in #1543, now we should enable pycodestyle (formally known and integrated to prospector as pep8) in CI.

tnir:
This PR enables pycodestyle in lint tox through prospector. The version of pycodestyle is exactly 2.0.0 due to the constraints from prospector while the latest is [2.3.1](https://pypi.python.org/pypi/pycodestyle/2.3.1).

Starts with the following lints:

- E127
- E128
- E262
- E265
- E266
- E301
- E302
- E303
- W291
- W293

In this PR the following lints are disabled:

- E129

Closes #1599

### Tests

- Failed before fixing errors/warnings: https://travis-ci.org/tnir/gunicorn/jobs/276449817
- Passed after fixing errors/warnings: https://travis-ci.org/tnir/gunicorn/jobs/276451030

### DIFF
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -27,8 +27,9 @@ from gunicorn.six import MAXSIZE
 
 class Worker(object):
 
-    SIGNALS = [getattr(signal, "SIG%s" % x)
-            for x in "ABRT HUP QUIT INT TERM USR1 USR2 WINCH CHLD".split()]
+    SIGNALS = [getattr(signal, "SIG%s" % x) for x in (
+                   "ABRT HUP QUIT INT TERM USR1 USR2 WINCH CHLD".split()
+               )]
 
     PIPE = []
 
@@ -197,11 +198,13 @@ class Worker(object):
     def handle_error(self, req, client, addr, exc):
         request_start = datetime.now()
         addr = addr or ('', -1)  # unix socket case
-        if isinstance(exc, (InvalidRequestLine, InvalidRequestMethod,
+        if isinstance(exc, (
+                InvalidRequestLine, InvalidRequestMethod,
                 InvalidHTTPVersion, InvalidHeader, InvalidHeaderName,
                 LimitRequestLine, LimitRequestHeaders,
                 InvalidProxyLine, ForbiddenProxyRequest,
-                SSLError)):
+                SSLError,
+            )):
 
             status_int = 400
             reason = "Bad Request"
--- a/.prospector.yaml
+++ b/.prospector.yaml
@@ -17,7 +17,19 @@ ignore-paths:
   - tests/test_gaiohttp.py  # TODO: We are going to remove this worker.
 
 pep8:
-  run: false
+  enable:
+    - E127
+    - E128
+    - E262
+    - E265
+    - E266
+    - E301
+    - E302
+    - E303
+    - W291
+    - W293
+  disable:
+    - E129
 
 pyflakes:
   run: false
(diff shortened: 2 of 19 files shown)


tilgovi:
I love it! I've been wanting to do this for a long time.

tilgovi:
I have no strong feeling either way.

I rarely have to blame and when I do I can iteratively blame once or twice until I find the interesting commit, but I frequently have to selectively commit when my format-on-save makes changes like this.

I also understand not wanting unnecessary churn and to keep the blame simple, so I haven't committed these changes myself in the past.

Either way is fine for me, @berkerpeksag.

tnir:
@berkerpeksag Yes, I feel similar as you, but this code churn is just **one time**, not frequent. While I love to `git blame`, repeating blame one or two times doesn't bother me as @tilgovi does.

tilgovi:
Yeah, I agree with that, too. This would be one time, and with pycodestyle enabled in CI once we have a clean baseline we never have this problem again.

tilgovi:
I'm actually surprised—and happy!—that there are not more such changes.

tnir:
Just pushed commits rebased against master to resolve conflicts.

tilgovi:
@berkerpeksag I think we're both happy either way and just waiting to hear from you. Let us know if we've persuaded you or you still want to remove the non-newline changes.

berkerpeksag:
I still think it's not worth the code churn. Most of these files (e.g. `gunicorn/http/*`) are quite stable and this one time commit will stay as a last change to those files for a while. Plus, I deliberately skip them last time I worked on our linter configuration to avoid code churn. But it's ultimately up to @benoitc, I don't claim I have an authority here :)

If we decide to keep them, we need to add some missing trailing commas to make future diffs more readable. Here are some examples I think we need to add trailing commas:

```diff
+from gunicorn.http.errors import (
+    NoMoreData, ChunkMissingTerminator, InvalidChunkSize
+)
```

and

```diff
+        _sendfile.argtypes = [
+            ctypes.c_int, ctypes.c_int,
+            ctypes.POINTER(ctypes.c_uint64), ctypes.c_size_t
+        ]
```

and

```diff
-                SSLError)):
+                SSLError
+            )):
```

and so on.

tnir:
@berkerpeksag Thanks for commenting for improving this PR. I just added trailing commas to the commit.

benoitc:
i’m ok with such change. But i think we should dedicate some lines about the syntax style  somewhere as i don’t want to simply follow blindly  pep8 or an « air du temps » thing. can be a separate commit/ticket though. Thoughts?

gunicorn/workers/base.py
```diff
@@ -197,11 +198,13 @@ def handle_abort(self, sig, frame):
...
         request_start = datetime.now()
         addr = addr or ('', -1)  # unix socket case
-        if isinstance(exc, (InvalidRequestLine, InvalidRequestMethod,
+        if isinstance(exc, (
+                InvalidRequestLine, InvalidRequestMethod,
                 InvalidHTTPVersion, InvalidHeader, InvalidHeaderName,
                 LimitRequestLine, LimitRequestHeaders,
                 InvalidProxyLine, ForbiddenProxyRequest,
-                SSLError)):
+                SSLError,
```
benoitc: why the extra comma there?

tnir:
FYI: #1678 was merged.

tnir:
Closed in favor of #1784, as this is difficult to fix conflicts due to missing prospector.

PR #1602, opened 2017-09-18
Support Dictionary Config setting.

tilgovi:
Taking over #1110 as per the OP request.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1316,8 +1316,8 @@ class LogConfigDict(Setting):
     default = {}
     desc = """\
     The log config dictionary to use, using the standard Python logging
-    module's dictConfig format.
-    If specified, this takes precedence over logconfig, which uses the older
+    module's dictConfig format added in python 2.7.
+    If available, this takes precedence over logconfig, which uses the older
     fileConfig format.
     """
 
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1309,6 +1309,19 @@ class LogConfig(Setting):
     """
 
 
+class LogConfigDict(Setting):
+    name = "logconfig_dict"
+    section = "Logging"
+    validator = validate_dict
+    default = {}
+    desc = """\
+    The log config dictionary to use, using the standard Python logging
+    module's dictConfig format.
+    If specified, this takes precedence over logconfig, which uses the older
+    fileConfig format.
+    """
+
+
 class SyslogTo(Setting):
     name = "syslog_addr"
     section = "Logging"
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1312,6 +1312,7 @@ class LogConfig(Setting):
 class LogConfigDict(Setting):
     name = "logconfig_dict"
     section = "Logging"
+    cli = ["--log-config-dict"]
     validator = validate_dict
     default = {}
     desc = """\
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1316,10 +1316,14 @@ class LogConfigDict(Setting):
     validator = validate_dict
     default = {}
     desc = """\
-    The log config dictionary to use, using the standard Python logging
-    module's dictConfig format added in python 2.7.
-    If available, this takes precedence over logconfig, which uses the older
-    fileConfig format.
+    The log config dictionary to use, using the standard Python
+    logging module's dictionary configuration format. This option
+    takes precedence over the :ref:`logconfig` option, which uses the
+    older file configuration format.
+
+    Format: https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig
+
+    .. versionadded:: 19.8
     """
 
 
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -8,7 +8,12 @@ import binascii
 import time
 import logging
 logging.Logger.manager.emittedNoHandlerWarning = 1
-from logging.config import fileConfig, dictConfig
+from logging.config import fileConfig
+try:
+    from logging.config import dictConfig
+except ImportError:
+    # python 2.6
+    dictConfig = None
 import os
 import socket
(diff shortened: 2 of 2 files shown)


gunicorn/glogging.py
```diff
@@ -226,7 +231,11 @@ def setup(self, cfg):
                     self.access_log, cfg, self.syslog_fmt, "access"
                 )
 
-        if cfg.logconfig:
+        if dictConfig and cfg.logconfig_dict:
+            config = CONFIG_DEFAULTS.copy()
+            config.update(cfg.logconfig_dict)
+            dictConfig(config)
```
berkerpeksag: We probably need to catch the following exceptions and raise `RuntimeError(message_from_catched_exception)` like we already did for `logconfig` in line 247:

> If an error is encountered during configuration, this function will raise a ValueError, TypeError, AttributeError or ImportError with a suitably descriptive message.

gunicorn/config.py
```diff
@@ -1309,6 +1309,19 @@ class LogConfig(Setting):
     """
 
 
+class LogConfigDict(Setting):
+    name = "logconfig_dict"
+    section = "Logging"
+    validator = validate_dict
+    default = {}
+    desc = """\
+    The log config dictionary to use, using the standard Python logging
```
berkerpeksag: We need to add a `versionadded` label.

gunicorn/config.py
```diff
@@ -1309,6 +1309,19 @@ class LogConfig(Setting):
...
+    module's dictConfig format added in python 2.7.
```
berkerpeksag: dictConfig -> \`\`dictConfig\`\`

berkerpeksag: We also need add a link to https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig

gunicorn/config.py
```diff
@@ -1309,6 +1309,19 @@ class LogConfig(Setting):
...
+    If available, this takes precedence over logconfig, which uses the older
```
berkerpeksag: logconfig -> :ref:\`--log-config <logconfig>\`

gunicorn/config.py
```diff
@@ -1309,6 +1309,19 @@ class LogConfig(Setting):
...
+    fileConfig format.
```
berkerpeksag: fileConfig -> \`\`fileConfig\`\`

gunicorn/config.py
```diff
@@ -1309,6 +1309,19 @@ class LogConfig(Setting):
     """
 
 
+class LogConfigDict(Setting):
+    name = "logconfig_dict"
```
berkerpeksag: Do we need to add `cli = ["--log-config-dict"]`?

gunicorn/config.py
berkerpeksag: It might be better to remove "added in python 2.7" and print a warning if `dictConfig` is `None` in `gunicorn/glogging.py`.

tilgovi:
@berkerpeksag I think I addressed all your comments. Thanks for the review.

gunicorn/config.py
```diff
@@ -1309,6 +1309,24 @@ class LogConfig(Setting):
...
+    section = "Logging"
+    cli = ["--log-config-dict"]
+    validator = validate_dict
+    default = {}
```
berkerpeksag: This might be a little bit misleading since we use `glogging.CONFIG_DEFAULTS` as default. Should we document this or keep it as implementation detail?

gunicorn/config.py
```diff
@@ -1309,6 +1309,24 @@ class LogConfig(Setting):
...
+    desc = """\
+    The log config dictionary to use, using the standard Python
+    logging module's dictionary configuration format. This option
+    takes precedence over the :ref:`logconfig` option, which uses the
```
berkerpeksag: This is a minor issue, but I'd say let's deprecate `--log-config` documentation-only and tell users to switch to `--log-config-dict`.

tilgovi: If you can explain a bit more, I will do a PR. Or I will review a PR. Thanks!

gunicorn/config.py
tilgovi: Maybe we should set the default as `None`? Would that be more "unset" / "undefined" / "default" to people?

Maybe it doesn't matter much either way.

gunicorn/config.py
```diff
@@ -1309,6 +1309,24 @@ class LogConfig(Setting):
     """
 
 
+class LogConfigDict(Setting):
+    name = "logconfig_dict"
+    section = "Logging"
+    cli = ["--log-config-dict"]
```
rthille: This doesn't seem to work, as there doesn't seem to be any conversion from the string supplied by the shell in some `ARGV[x]` to an actual dict:
```
bash-3.2$ env/bin/gunicorn --version
env/bin/gunicorn --version
gunicorn (version 19.9.0)
bash-3.2$ env/bin/gunicorn --log-config-dict {} 'flask.main:app'
env/bin/gunicorn --log-config-dict {} 'flask.main:app'

Error: Value is not a dictionary: {} 
```

javabrett: Right, maybe `cli` should have been left off this option until someone JSON-parses it.  Can you use config file?

Also https://stackoverflow.com/a/18609361/452210 re `argparse` capabilities in this area.

rthille: I think we can extend our config we're passing via `--config=python:cariumlib.djangolib.gunicorn_config`
It's funny, I tried hacking `type=json.loads` in, but couldn't get it to work.

javabrett: Maybe you could convert your PR comment into a new issue along the lines of advertised CLI option `--log-config-dict` cannot work.

rthille: @javabrett Done: https://github.com/benoitc/gunicorn/issues/1909

Issue #1603, opened 2017-09-18
make the use  SO_REUSE_PORT optional


benoitc:
This ticket supersedes #1553 . I also think this should be optional, there are some case where w don't want to reuse a port and  it should probably be used implicitly.

The change looks easy: 
https://github.com/benoitc/gunicorn/blob/master/gunicorn/sock.py#L42-L47

Thoughts?

benoitc:
just linking an old discussion about it https://github.com/benoitc/gunicorn/pull/1131#issuecomment-245559091 for the record.

tilgovi:
:+1: already has access to the config object, too. Seems straightforward.

btubbs:
I've had a number of times where a zombie Gunicorn process was running at the same time as the one I was developing, but I didn't know it.  If disabling SO_REUSEPORT by default would make such situations fail fast with a "port already in use" error instead of silently pretending that everything's OK, that would be a plus for me.

tilgovi:
Let's do it. Adding the Help Wanted label.

martinbroadhurst:
I have submitted a PR for this: #1669

PR #1606, opened 2017-09-25
Let Errno 92 pass (support windows Linux Subsystem)

alex-gon:
When running gunicorn in a windows Linux Subsystem I get this error. 

    16:21:30 web.1   |     sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
    16:21:30 web.1   | OSError: [Errno 92] Protocol not available

This fix in this pull request mirrors the way they fixed the bug in eventlet. https://github.com/eventlet/eventlet/issues/418

### DIFF
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -43,7 +43,7 @@ class BaseSocket(object):
             try:
                 sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
             except socket.error as err:
-                if err.errno not in (errno.ENOPROTOOPT, errno.EINVAL):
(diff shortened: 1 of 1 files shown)


tilgovi:
Thank you for the pull request!

It turns out that SO_REUSEPORT can be a surprising behavior so in #1603 there is the proposal to make it optional and maybe off by default. If we turn it off by default, then I think this patch is not necessary. Do you agree?

berkerpeksag:
#1603 has been fixed and `SO_REUSEPORT` is now disabled by default. @tilgovi @alex-gon can we close this one now?

tilgovi:
I think so, @berkerpeksag

tilgovi:
@alex-gon if you get a chance to test master, please let us know if it works for you.

Issue #880, opened 2014-09-10
Timeout exception after greenlet.switch with gevent worker on gunicorn newer than 0.15


w0rse:
Hello, gunicorn!

We are seeing a lot of the following errors after upgrading gunicorn to the latest version:
2014-09-10 03:19:12 [3367] [ERROR] Socket error processing request.
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/gunicorn/workers/async.py", line 42, in handle
    req = six.next(parser)
  File "/usr/lib/python2.6/site-packages/gunicorn/http/parser.py", line 39, in **next**
    self.mesg = self.mesg_class(self.cfg, self.unreader, self.req_count)
  File "/usr/lib/python2.6/site-packages/gunicorn/http/message.py", line 151, in **init**
    super(Request, self).**init**(cfg, unreader)
  File "/usr/lib/python2.6/site-packages/gunicorn/http/message.py", line 48, in **init**
    unused = self.parse(self.unreader)
  File "/usr/lib/python2.6/site-packages/gunicorn/http/message.py", line 164, in parse
    self.get_data(unreader, buf, stop=True)
  File "/usr/lib/python2.6/site-packages/gunicorn/http/message.py", line 155, in get_data
    data = unreader.read()
  File "/usr/lib/python2.6/site-packages/gunicorn/http/unreader.py", line 37, in read
    d =  self.chunk()
  File "/usr/lib/python2.6/site-packages/gunicorn/http/unreader.py", line 63, in chunk
    return self.sock.recv(self.mxchunk)
  File "/usr/lib64/python2.6/site-packages/gevent/socket.py", line 392, in recv
    self._wait(self._read_event)
  File "/usr/lib64/python2.6/site-packages/gevent/socket.py", line 298, in _wait
    self.hub.wait(watcher)
  File "/usr/lib64/python2.6/site-packages/gevent/hub.py", line 341, in wait
    result = waiter.get()
  File "/usr/lib64/python2.6/site-packages/gevent/hub.py", line 568, in get
    return self.hub.switch()
  File "/usr/lib64/python2.6/site-packages/gevent/hub.py", line 331, in switch
    return greenlet.switch(self)
timeout: timed out

Further tests revealed that the same code produces that errors on any version of gunicorn newer than 0.15. Is there anything we can do to fight this behavior? We use gevent workers version 1.0.1 and gunicorn is behind a proxy with keepalive enabled.

ochronus:
[...]

PR #1616, opened 2017-10-06
Force gevent connected socket to blocking mode

jkemp101:
The gevent worker assumes sockets are running in blocking mode.  The current code forces the listening socket into blocking mode [here](https://github.com/benoitc/gunicorn/blob/19.7.1/gunicorn/workers/ggevent.py#L98).  But the connected socket is not forced to blocking mode.  Therefore if other code has set the default socket timeout with a call to `setdefaulttimeout` it will be set as the connected socket's timeout in the gevent initialization code [here](https://github.com/gevent/gevent/blob/v1.2a2/src/gevent/_socket3.py#L90).

The gevent worker relies on the keepalive timeout to close idle connections.   Setting the default timeout below the keepalive timeout causes the worker to run in a undefined mode.  I believe this may fix the issues in https://github.com/benoitc/gunicorn/issues/880 but it was not clear if everyone having that issue were also calling `setdefaulttimeout`.  In our particular case we are calling `setdefaulttimeout` which caused sockets to close before the keepalive expired.  This causes a bunch of socket timeout error messages and potentially 504 errors with AWS ELBs.

### DIFF
--- a/gunicorn/workers/ggevent.py
+++ b/gunicorn/workers/ggevent.py
@@ -148,6 +148,12 @@ class GeventWorker(AsyncWorker):
         except:
             pass
 
+    def handle(self, listener, client, addr):
+        # Connected socket timeout defaults to socket.getdefaulttimeout().
+        # This forces to blocking mode.
+        client.setblocking(1)
+        super(GeventWorker, self).handle(listener, client, addr)
+
     def handle_request(self, listener_name, req, sock, addr):
         try:
             super(GeventWorker, self).handle_request(listener_name, req, sock,


tilgovi:
Thank you for providing such detailed information and a patch.

The patch is simple and it sounds like it would work, but do you think that the assumption Gunicorn makes about the socket being blocking is a bad one? Should we discuss any alternative changes?

jkemp101:
I was also concerned about that but came to the conclusion its probably ok.  This gets a bit confusing since we are talking about blocking sockets that in the gevent world aren't necessarily really blocking anything.

My thoughts:
1. The default value for the timeout puts the connected socket in blocking mode.  Therefore the majority of the gevent worker users are already running with blocking sockets.  I would guess most people calling `setdefaulttimeout` don't realize the implications to the gevent worker.
1. The `gevent.Timeout` object in `timeout_ctx` seems to handle "idle" sockets that are waiting to receive data.  The keepalive timeout seems appropriate here to close connections.  This was important to me because AWS CLBs do a lot of pre-opened connections that may never send a request.  I wanted to make sure if the CLB didn't close these connections after a reasonable amount of time they wouldn't stack up in gunicorn.  The keepalive timer kicked in and closed these connections even when no request had been set.
1. I believe setting the default timeout also affects sending data to clients.  With blocking sockets the send would theoretically block the greenlet until the underlying TCP timers detected the socket was broken and closed it.  I don't think we would want to use the keepalive timeout for this anyway so a new send timeout would probably be needed with an associated refactoring of the send response code.  This seems like a pretty unlikely edge case.

tilgovi:
Thanks for all that analysis. I agree with you. It does sound like it would fix #880, too. At least one comment there suggests that changing the socket timeout fixed the problem.

PR #1617, opened 2017-10-07
#825 Possibility to disable server header

mprzytulski:
Original code by @mathiasverhoeven just resurrecting a PR

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -1277,3 +1277,18 @@ The variables are passed to the the PasteDeploy entrypoint. Example::
 
 .. versionadded:: 19.7
 
+server_name
+~~~~~~~~~~~~~~~~~~~
+
+* ``--server-name STRING``
+* ``gunicorn/X.Y.Z``
+
+Server name to use in the ``Server`` HTTP Header on responses.
+
+server_name
+~~~~~~~~~~~~~~~~~~~
+
+* ``--no-server-name``
+* ``None``
+
+If set, the ``Server`` HTTP Header will not be returned on responses.
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1870,6 +1870,38 @@ class DoHandshakeOnConnect(Setting):
     """
 
 
+class ServerName(Setting):
+    name = "server_name"
+    section = "Server Mechanics"
+    cli = ["--server-name"]
+    validator = validate_string
+    default = SERVER_SOFTWARE
+
+    desc = """\
+        Server name to use in the ``Server`` HTTP Header on responses.
+
+        If not set, the default Gunicorn server name will be used:
+        ``gunicorn/X.Y.Z``.
+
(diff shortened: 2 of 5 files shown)


benoitc:
why this PR?

benoitc:
I think I would prefer a PR that optimally remove the server name instead of allowing someone to change it, do you mind to change that pr to do that?

tilgovi:
This PR does allow you to remove it also. I don't think I have any objection. I try to be hesitant to add more options, but I don't think I see a problem with these or a lot of maintenance burden.

We would need to update the `versionadded` directives if we want to merge this, though.

benoitc:
@tilgovi but I don't want an option to change the server name actually, I don't really the see the point. 

Maybe better than an option we could allow an application to bypass it by providing its own header, if it's not already possible. So anyone that provide a custom `Server` would header in the application via a middle ware or else would replace it. 

Thoughts?

tilgovi:
Would we add the option to disable it, though?

RonRothman:
@benoitc I think the reasoning behind this request is that it's considered more secure to not disclose which server is running (because it would help narrow down which exploits a hacker would want to try).

E.g. https://www.acunetix.com/blog/articles/configure-web-server-disclose-identity/

In our production server, we do this:
```
gunicorn.SERVER_SOFTWARE = <some generic server token - not gunicorn>
```

mprzytulski:
@benoitc I'm totally fine with all the approaches - what I really need is to be able to disable server header, if you like I can remove an option to overwrite it. I can also take a look how difficult it would be to bypass custom `Server` header when provided. Just please let me know if that's the prefered solution.

whisller:
@benoitc @tilgovi did you get in to the conclusion what approach you prefer? :)

benoitc:
Sorry for the delay to answer. In the vein of nginx , if people wants such kind of feature  I think I would prefer to have one option: `server_tokens =true|false|string` . With default to true

Enables or disables emitting gunicorn version on error pages and in the “Server” response header field  (true or false). Additionally the Server header value could be replaced by a string. An empty string would completely remove it. 

cf http://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens


@mprzytulski cna you revisit your PR in that way? If not I may abble to do it on the next release.

berkerpeksag:
+1 for adding a `server_tokens` setting.

tilgovi:
Also +1 from me to support `server_tokens` default to `True` but allow `False`. It's nice to have some similarity to existing configuration (NGINX) people may be familiar with. If it's easy enough, with our configuration system, to support a string then that can specify the name. Otherwise, just a way to enable/disable is enough, and much better than two settings.

whisller:
WOW it's almost year when this PR got created :D LOL

benoitc:
@whisller so what?

lihan:
+1

benoitc:
superseded by #2233

PR #1623, opened 2017-10-14
Rewrite parse_address utils

hramezani:
better handling for unix socket address in `parse_address`

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -21,6 +21,7 @@ import inspect
 import errno
 import warnings
 import logging
+import re
 
 from gunicorn import _compat
 from gunicorn.errors import AppImportError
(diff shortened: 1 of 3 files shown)


tilgovi:
Thank you for the PR. This is definitely cleaner. It is only used by the arbiter. I had to double check that because if this were used for every connection I would worry about performance.

I think this is good, though! 👍

gunicorn/util.py
```diff
@@ -232,11 +233,8 @@ def is_ipv6(addr):
 
 
 def parse_address(netloc, default_port=8000):
-    if netloc.startswith("unix://"):
-        return netloc.split("unix://")[1]
-
-    if netloc.startswith("unix:"):
-        return netloc.split("unix:")[1]
+    if re.match('unix:(//)?', netloc):
```
berkerpeksag: I'd use `r''` strings for regexes.

hramezani:
@berkerpeksag , thanks. tests added.

berkerpeksag:
Thanks!

PR #1629, opened 2017-10-20
Improve parse_address utils tests

hramezani:


tests/test_util.py
```diff
@@ -3,10 +3,22 @@
 # This file is part of gunicorn released under the MIT license.
 # See the NOTICE for more information.
 
+import pytest
+
 from gunicorn import util
 
 
-def test_parse_address():
+def test_parse_address(capsys):
```
tilgovi: Seems unused. Is it needed?

tests/test_util.py
```diff
@@ -3,10 +3,22 @@
...
+
+    assert util.parse_address('') == ('0.0.0.0', 8000)
+    assert util.parse_address('[::1]:8000') == ('::1', 8000)
+    assert util.parse_address('localhost:8000') == ('localhost', 8000)
+    assert util.parse_address('127.0.0.1:8000') == ('127.0.0.1', 8000)
+    assert util.parse_address('localhost') == ('localhost', 8000)
+
+    with pytest.raises(RuntimeError) as err:
+        assert util.parse_address('127.0.0.1:test')
+    assert err.value.message == "'test' is not a valid port number."
```
berkerpeksag: This will raise an `AttributeError` under Python 3. I'd use `str(exc)` instead.

tests/test_util.py
```diff
@@ -3,10 +3,22 @@
...
 from gunicorn import util
 
 
-def test_parse_address():
+def test_parse_address(capsys):
     # Test unix socket addresses (PR #1623)
     assert util.parse_address('unix://var/run/test.sock') == 'var/run/test.sock'
     assert util.parse_address('unix:/var/run/test.sock') == '/var/run/test.sock'
+
+    assert util.parse_address('') == ('0.0.0.0', 8000)
```
berkerpeksag: We could turn this test to use `pytest.mark.parametrize`: https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize

hramezani:
@berkerpeksag, thanks. it is ok for use pytest.mark.parametrize , but how do I test RuntimeError with pytest.mark.parametrize?

berkerpeksag:
> but how do I test RuntimeError with pytest.mark.parametrize?

I'd move it to a separate test function (e.g. `test_parse_address_error`or `test_parse_address_invalid`)

PR #1632, opened 2017-10-24
Fix umask option bug(issues#1622)

hramezani:
Fix umask option bug. [issue_1622](https://github.com/benoitc/gunicorn/issues/1622l)

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -43,8 +43,8 @@ def make_settings(ignore=None):
 
 
 def auto_int(_, x):
-    if x.startswith('0') and not x.lower().startswith('0x'):
-        # for compatible with octal numbers in python3
+    # for compatible with octal numbers in python3
+    if re.match(r'0(\d)', x, re.IGNORECASE):
         x = x.replace('0', '0o', 1)
     return int(x, 0)
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -374,7 +374,9 @@ def test_reload(options, expected):
 
 
 @pytest.mark.parametrize("options, expected", [
-    (["--umask 0", "myapp:app"], 0),
+    (["--umask", "0", "myapp:app"], 0),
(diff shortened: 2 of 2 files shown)


gunicorn/config.py
```diff
@@ -43,8 +43,8 @@ def make_settings(ignore=None):
 
 
 def auto_int(_, x):
-    if x.startswith('0') and not x.lower().startswith('0x'):
-        # for compatible with octal numbers in python3
+    # for compatible with octal numbers in python3
+    if re.match('0(\d)', x, re.IGNORECASE):
```
berkerpeksag: I'd use `r''` to avoid subtle bugs.

hramezani:
@berkerpeksag thanks, I add those examples to `test_umask_config`.

berkerpeksag:
I think the problem here is that our test code doesn't call `auto_int` so even if we already added a test for `--umask 0` we missed this bug.

berkerpeksag:
Nevermind, the existing test was wrong.

tests/test_config.py
```diff
@@ -375,6 +375,8 @@ def test_reload(options, expected):
 
 @pytest.mark.parametrize("options, expected", [
     (["--umask 0", "myapp:app"], 0),
```
berkerpeksag: This should read:

```py
(["--umask", "0", "myapp:app"], 0),
```

tests/test_config.py
```diff
@@ -375,6 +375,8 @@ def test_reload(options, expected):
...
+    (["--umask 0o0", "myapp:app"], 0),
```
berkerpeksag: Same as above:

```py
(["--umask", "0o0", "myapp:app"], 0),
(["--umask", "0x0", "myapp:app"], 0),
```

hramezani:
@berkerpeksag, thanks, you are right. I fixed the tests

berkerpeksag:
Thanks!

Issue #1234, opened 2016-03-24
timestamps wrong in logs


tisdall:
This may be related to the change in #850 .

I recently upgraded to gunicorn 19.4.5 and noticed this today:

```
[2016-03-24 07:52:52 +0000] [24183] [INFO] Booting worker with pid: 24183
[2016-03-24 07:52:52 +0000] [24184] [INFO] Booting worker with pid: 24184
[2016-03-24 07:52:52 +0000] [24185] [INFO] Booting worker with pid: 24185

[deploy@prod-app01 ~]$ date
Thu Mar 24 07:54:39 EDT 2016
```

The logs are showing the current time, but showing "+0000" when the actual timezone is "-0400".

I'm running Django 1.4 with timezone support on and the current timezone set correctly.

tilgovi:
Looks like maybe this is a bug with Python 2 logging: http://stackoverflow.com/questions/27858539/python-logging-module-emits-wrong-timezone-information

benoitc:
@tilgovi what about having http://stackoverflow.com/a/27865750 ?

tilgovi:
@benoitc might be a good idea

berkerpeksag:
If we want to log the current *local* time, can't we use something like

```sh
export TZ="Europe/Istanbul"
```

and

```py
time.tzset()
time.localtime()
```

in `logging.Formatter.formatTime`?

berkerpeksag:
So I think is a Python 2-only bug:

```
$ gunicorn test:app --access-logfile=-
[2017-11-04 09:46:39 +0300] [30874] [INFO] Starting gunicorn 19.7.1
[2017-11-04 09:46:39 +0300] [30874] [INFO] Listening at: http://127.0.0.1:8000 (30874)
[2017-11-04 09:46:39 +0300] [30874] [INFO] Using worker: sync
[2017-11-04 09:46:39 +0300] [30877] [INFO] Booting worker with pid: 30877
127.0.0.1 - - [04/Nov/2017:09:46:43 +0300] "GET / HTTP/1.1" 200 14 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/49.0.2623.112 Safari/537.36"
[...]

PR #1636, opened 2017-10-29
Fix wrong log timestamp base on http://stackoverflow.com/a/27865750

hramezani:
fix [issue_1234](https://github.com/benoitc/gunicorn/issues/1234) based on @benoitc [response](https://github.com/benoitc/gunicorn/issues/1234#issuecomment-201217687)

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -14,9 +14,40 @@ import socket
 import sys
 import threading
 import traceback
+from datetime import datetime, timedelta, tzinfo
 
 from gunicorn import util
 from gunicorn.six import PY3, string_types
+
+
+# based on http://stackoverflow.com/a/27865750
+class LocalTimeZone(tzinfo):
+    _offset = timedelta(seconds=-time.timezone)
+    _dst = timedelta(0)
+    _name = "local time zone"
+    def utcoffset(self, dt):
+        return self.__class__._offset
+    def dst(self, dt):
+        return self.__class__._dst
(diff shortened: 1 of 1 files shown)


gunicorn/glogging.py
```diff
@@ -81,7 +102,7 @@
             "generic": {
                 "format": "%(asctime)s [%(process)d] [%(levelname)s] %(message)s",
                 "datefmt": "[%Y-%m-%d %H:%M:%S %z]",
-                "class": "logging.Formatter"
+                "class": "Formatter"
```
wking: Doesn't this have to be the full `gunicorn.glogging.Formatter`?

hramezani: @wking thanks, I changed it to `gunicorn.glogging.Formatter`

setup.py
```diff
@@ -87,6 +87,9 @@ def run_tests(self):
     zip_safe=False,
     packages=find_packages(exclude=['examples', 'tests']),
     include_package_data=True,
+    install_requires=[
+        'tzlocal',
```
berkerpeksag: I think our policy is to not have dependencies and I'm not sure it's worth to add `tzlocal`.

hramezani: @berkerpeksag thanks, I remove `tzlocal`, and convert datetime to local time by pyhton standard library.

berkerpeksag: I'd say wait for the result of the discussion in #1234.

PR #1231, opened 2016-03-22
prevent crash when reporting an error

benoitc:
Try to not crash when reporting an error. This changes makes sure we can log the uri and also handle SSL errors.

fix #1177

PR #1637, opened 2017-11-03
workers.base: Log 500 error message

wking:
This makes it a bit easier to debug than just logging the fact that there was an exception.  Also restore logging in the URI-less case, since that was broken in 79011d0c (#1071) but not restored in 1ccebab7 (#1231).

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -11,7 +11,6 @@ import select
 import signal
 import sys
 import time
-import traceback
 
 from gunicorn.errors import HaltServer, AppImportError
 from gunicorn.pidfile import Pidfile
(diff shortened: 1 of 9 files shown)


wking:
I've expanded the scope a bit after discovering the wonders of `exc_info`, and pushed a new commit that uses `exc_info` where appropriate, and also:

* Drops a direct `traceback` call which snuck in in #1254 after #1201 had removed the previous occurrences.
* Removes the “Error processing SSL request” debug messages, which we've had since 1198cc27, but which aren't needed because `handle_error` will be [logging those anyway][1].
* Restores URI-less 500 logging, as discussed in the initial PR comment.

[1]: https://github.com/benoitc/gunicorn/blob/19.7.1/gunicorn/workers/base.py#L232

wking:
Tests for the lines I've touched may be more trouble than they're worth.  I've worked up one for a broken worker_exit hook to show what they look like, but left the others untested for now.  Let me now if the “coverage has worsened” errors are acceptable in this case, or if I need to push through and add enough similar tests to cover all of the log instances.

dpetukhov:
Hi. Your pull request can help fix my issue, check it out: https://github.com/benoitc/gunicorn/issues/1660
So I'm voting to work this out and merge if everything is okay.

Will you PR fix issues related to SSLError exceptions?

benoitc:
this is need q more carefull review, exposing traceback in logs can expose some private data. I will comment the changes asap.

tilgovi:
Also, I don't think we need to add `exc_info` where the exception is the current exception on the stack and `log.exception()` is called.

benoitc:
stalled

PR #1638, opened 2017-11-03
faq: Fix --log-file → --error-logfile

### LINKED ISSUE #1294: Possible mistake in FAQ "Why I don’t see any logs in the console?"
This FAQ entry:
http://docs.gunicorn.org/en/latest/faq.html#why-i-don-t-see-any-logs-in-the-console

Says:

> In version R19, Gunicorn doesn’t log by default in the console. To watch the logs in the console you need to use the option --log-file=-. In version R20, Gunicorn logs to the console by default again.

I think "R20" should instead be "R19.2", given:
http://docs.gunicorn.org/en/latest/settings.html#errorlog

Also `--log-file=-` has been replaced by `--error-logfile=-` in v0.13.0.

wking:
The option rename is from 66f7271c.

Also fix the version for console restoration from 7be15d63 (#892).  7be15d63 didn't know which release it would end up in, but it turned out to be 19.2:

```
$ git describe --contains 7be15d633
19.2~74^2
```

Fixes #1294.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -107,8 +107,8 @@ Why I don't see any logs in the console?
 ----------------------------------------
 
 In version R19, Gunicorn doesn't log by default in the console.
-To watch the logs in the console you need to use the option ``--log-file=-``.
-In version R20, Gunicorn logs to the console by default again.
+To watch the logs in the console you need to use the option ``--error-logfile=-``.
+In version R19.2, Gunicorn logs to the console by default again.
 
 Kernel Parameters
 =================


docs/source/faq.rst
```diff
@@ -107,8 +107,8 @@ Why I don't see any logs in the console?
 ----------------------------------------
 
 In version R19, Gunicorn doesn't log by default in the console.
```
berkerpeksag: I'd say drop this sentence, move the last sentence to the top, and change it to start with "Starting with Gunicorn 19.2, [...]".

docs/source/faq.rst
```diff
@@ -107,8 +107,8 @@ Why I don't see any logs in the console?
...
-To watch the logs in the console you need to use the option ``--log-file=-``.
-In version R20, Gunicorn logs to the console by default again.
+To watch the logs in the console you need to use the option ``--error-logfile=-``.
```
berkerpeksag: Also, it would be nice to mention that Gunicorn doesn't log access logs by default. They need to pass `--access-logfile=-` to make Gunicorn logs everything to the console.

benoitc:
@berkerpeksag what should be done to close that PR?

benoitc:
fixed in 1aa9cf0529fe87e4084dea9dadc97848934cceb4 . Thanks for the patch!

PR #1644, opened 2017-11-17
TCP RST is sent if application does not consume all posted data

msmania:
If we receive POST but WSGI application does not consume all posted data from
the stream, we just close the socket, resulting in an abortive session close
with TCP RST without graceful finish.

rst

This was observed when HTTP POST was sent via XHR over HTTPS.  If this happens,
Chrome displays ERR_CONNECTION_RESET and we cannot see response payload.

chrome

The proposed fix is to discard data in the body before closing the socket as
Eventlet does in `HttpProtocol.handle_one_request`.

### DIFF
--- a/gunicorn/http/body.py
+++ b/gunicorn/http/body.py
@@ -257,3 +257,12 @@ class Body(object):
                 line, data = data[:pos + 1], data[pos + 1:]
                 ret.append(line)
         return ret
+
+    def discard(self):
+        discarded_len = 0
+        bytes_read = six.MAXSIZE
+        while bytes_read > 0:
+            data = self.reader.read(1024)
+            bytes_read = len(data)
+            discarded_len += bytes_read
+        return discarded_len
--- a/gunicorn/workers/sync.py
+++ b/gunicorn/workers/sync.py
@@ -186,6 +186,17 @@ class SyncWorker(base.Worker):
             finally:
                 if hasattr(respiter, "close"):
                     respiter.close()
+            # Discard data in the body before closing the socket to prevent
+            # unexpected TCP RST from being sent.
+            try:
+                discarded_len = req.body.discard()
+                if discarded_len > 0:
+                    self.log.debug("Discarding %d bytes in the body.",
+                                   discarded_len)
+            except Exception:
+                # Response has been sent.  Ignore exceptions while discarding
+                # data in order not to invoke handle_error.
+                pass
         except EnvironmentError:
             # pass to next try-except level
             six.reraise(*sys.exc_info())
--- a/THANKS
+++ b/THANKS
@@ -159,6 +159,7 @@ Sébastien Fievet <zyegfryed@gmail.com>
 TedWantsMore <TedWantsMore@gmx.com>
 Thomas Grainger <tagrain@gmail.com>
 Thomas Steinacher <tom@eggdrop.ch>
+Toshihito Kikuchi <k.toshihito@yahoo.de>
 Travis Cline <travis.cline@gmail.com>
 Travis Swicegood <development@domain51.com>
 Trey Long <trey@ktrl.com>


jamadden:
gevent also consumes outstanding data when a response is finished. There are some considerations to [avoid DoS-type scenarios](https://github.com/gevent/gevent/blob/e913b3f64b353dcf1c2bbff1d58762ae81a4ed4b/src/gevent/pywsgi.py#L125) and [prevent spurious exception reporting](https://github.com/gevent/gevent/blob/e913b3f64b353dcf1c2bbff1d58762ae81a4ed4b/src/gevent/pywsgi.py#L964).

msmania:
@jamadden Good to know!  [eventlet](https://github.com/eventlet/eventlet/blob/c669f3040f250d04b8bec266409c9ad442f79876/eventlet/wsgi.py#L592) has the same exception handling.  Will add it to my commit.

I can't find a flag like gevent's `_chunked_input_error` in Gunicorn's reader.  Adding exception handling is enough..?

benoitc:
@tilgovi should we close it?

tilgovi:
The patch looks good, but I would love @jamadden to sign off. It would also be good if we can add a test somehow.

gunicorn/workers/sync.py
```diff
@@ -186,6 +186,17 @@ def handle_request(self, listener, req, client, addr):
             finally:
                 if hasattr(respiter, "close"):
                     respiter.close()
+            # Discard data in the body before closing the socket to prevent
+            # unexpected TCP RST from being sent.
+            try:
+                discarded_len = req.body.discard()
```
benoitc: if we already have the length maybe we could also try to ready with a larger payload length?

gunicorn/http/body.py
```diff
@@ -257,3 +257,12 @@ def readlines(self, size=None):
                 line, data = data[:pos + 1], data[pos + 1:]
                 ret.append(line)
         return ret
+
+    def discard(self):
+        discarded_len = 0
+        bytes_read = six.MAXSIZE
+        while bytes_read > 0:
```
jamadden: `self.reader` here is already terminated by the HTTP framing (content-length or chunks) right? If so this seems fine (otherwise it's a new DOS opportunity).

gunicorn/workers/sync.py
jamadden: It feels like this should be an a `finally` block...but I guess gunicorn doesn't really do request pipelining so it probably doesn't make much difference.

gunicorn/workers/sync.py
```diff
@@ -186,6 +186,17 @@ def handle_request(self, listener, req, client, addr):
             finally:
                 if hasattr(respiter, "close"):
                     respiter.close()
+            # Discard data in the body before closing the socket to prevent
```
jamadden: This appearing in `sync.py` makes be think that the same code block needs to appear in `base_async.py` for the async workers. (It seems that there are essentially two copies of `handle_request` at this point, one in each of those files, that are almost the same.)

benoitc:
rereading that change. I am not sure we want  to do it. There are many reason we want to close the connection (USR2, HUP...), closing it after having read all the data can take times and this is why are not doing it. Thoughts?

jamadden:
gevent and waitress both ensure that the entire request body is consumed. This matters to the server if multiple requests are going to be supported, and it matters to the client, because, according to the original poster:

> If we receive POST but WSGI application does not consume all posted data from the stream, we just close the socket....If this happens, Chrome displays ERR_CONNECTION_RESET and we cannot see response payload.

Is this super critical? Probably not, because a piece of WSGI middleware can always be inserted that ensures the body is consumed. The stdlib's wsgiref *doesn't* consume the body, as far as I can see, but it also doesn't terminate the input or handle chunked encoding. werkzeug.serving also doesn't appear to handle it. So an application that wants to be truly portable will have to include such middleware.

benoitc:
@jamadden OK I see, thanks for the enlightenment :) And I agree the same part should be ported to the async workers.

gunicorn/http/body.py
benoitc: I would just iterate the reader there, since it's terminated. Like the `req.body.readlines()` without appending the data.

tilgovi:
This kind of logic might be neatly implemented inside the body readers so that they do the discard on close.

PR #1654, opened 2017-12-01
Fix access logs not emitted when using logconfig_dict

### LINKED ISSUE #1652: Access log not emitted when using `logconfig_dict`
Using the unreleased version from `master`, HTTP requests do not create log records in the logger `gunicorn.access` when using the new `logconfig_dict`.

See relevant snippet from `glogging.py`:

```python
    def access(self, resp, req, environ, request_time):
        """ See http://httpd.apache.org/docs/2.0/logs.html#combined
        for format details
        """

        if not (self.cfg.accesslog or self.cfg.logconfig or
           (self.cfg.syslog and not self.cfg.disable_access_log_redirection)):
            return
```

NicolasLM:
Fixes #1652

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -339,6 +339,7 @@ class Logger(object):
         """
 
         if not (self.cfg.accesslog or self.cfg.logconfig or
+           self.cfg.logconfig_dict or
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Thank you! :)

PR #1669, opened 2017-12-28
Make SO_REUSEPORT Optional

martinbroadhurst:
I'm open to opinions as to whether ``SO_REUSEPORT`` should be the default or not.
Also, would appreciate some advice about unit testing this change.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -428,6 +428,14 @@ to enable or disable its usage.
 .. versionchanged:: 19.6
    added support for the ``SENDFILE`` environment variable
 
+.. _reuse-port:
+
+reuse_port
+~~~~~~~~~~
+
+* ``--reuse-port``
+* ``False``
+
 .. _chdir:
 
 chdir
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -436,6 +436,10 @@ reuse_port
 * ``--reuse-port``
 * ``False``
 
+Set the ``SO_REUSEPORT`` flag on the listening socket.
+
+.. versionadded:: 19.8
+
 .. _chdir:
 
 chdir
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -205,6 +205,10 @@ class Config(object):
         return True
 
     @property
+    def reuse_port(self):
+        return self.settings['reuse_port'].get()
+
+    @property
     def paste_global_conf(self):
         raw_global_conf = self.settings['raw_paste_global_conf'].get()
         if raw_global_conf is None:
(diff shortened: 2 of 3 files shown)


tilgovi:
I am slightly in favor of a `False` default. This option matters most for users seeking the performance benefits, I think. Those users may know that they want this and they are likely to read through the server options for tweaks. New users, however, can sometimes be killing and restarting their server over and over while testing out code in development and reusing the port can sometimes lead to having two instances of Gunicorn running on a development machine, serving different code. To minimize confusion, I would make it `False` by default. I'm open to arguments for `True`, though.

No strong opinion on how to test this.

martinbroadhurst:
Thanks. I have updated the documentation.

docs/source/settings.rst
```diff
@@ -428,6 +428,14 @@ to enable or disable its usage.
 .. versionchanged:: 19.6
    added support for the ``SENDFILE`` environment variable
 
+.. _reuse-port:
+
+reuse_port
+~~~~~~~~~~
+
+* ``--reuse-port``
```
berkerpeksag: Documentation for the new option will be generated automatically from `gunicorn/config.py` when you run `make html` in `docs/` directory.

gunicorn/config.py
```diff
@@ -953,6 +957,18 @@ class Sendfile(Setting):
...
+class ReusePort(Setting):
+    name = "reuse_port"
+    section = "Server Mechanics"
+    cli = ["--reuse-port"]
+    validator = validate_bool
+    action = "store_true"
+    default = False
+
+    desc = """\
+        Set the ``SO_REUSEPORT`` flag on the listening socket.
```
berkerpeksag: Please add `.. versionadded:: 19.8`.

docs/source/settings.rst
martinbroadhurst: I'm not sure I understand your comment, @berkerpeksag. I generated this change by running `` make -C docs html``. Am I not supposed to commit settings.rst?

gunicorn/config.py
tilgovi: I always forget this and you always catch it. Thanks! :-D

docs/source/settings.rst
berkerpeksag: I thought you manually edited it because I didn't see the "Set the ``SO_REUSEPORT`` flag on the listening socket." part in the generated file. Now I don't know why `make html` didn't get the documentation in the `desc` attribute.

martinbroadhurst: I generated it, but then edited it because running it also brought in ``logconfig_dict`` (which presumably was omitted previously and should be in its own PR), and also changed ``chdir``, ``user`` and ``group``. I must have got the edit wrong. Sorry about that. It's all there now.

berkerpeksag:
The cause of build failure is https://github.com/landscapeio/prospector/issues/200. All tests passed and I don't think we need a test case for this since testing the actual change reliably would be hard.

PR #1671, opened 2018-01-01
Rewrite `parse_address` util and add one test

hramezani:
Maybe better implementation of `parse_address` util.

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -247,33 +247,28 @@ def is_ipv6(addr):
     return True
 
 
-def parse_address(netloc, default_port=8000):
+def parse_address(netloc, default_port='8000'):
     if re.match(r'unix:(//)?', netloc):
         return re.split(r'unix:(//)?', netloc)[-1]
 
     if netloc.startswith("tcp://"):
         netloc = netloc.split("tcp://")[1]
+    host, port = netloc, default_port
 
-    # get host
     if '[' in netloc and ']' in netloc:
-        host = netloc.split(']')[0][1:].lower()
+        host = netloc.split(']')[0][1:]
+        port = (netloc.split(']:') + [default_port])[1]
     elif ':' in netloc:
-        host = netloc.split(':')[0].lower()
+        host, port = (netloc.split(':') + [default_port])[:2]
     elif netloc == "":
-        host = "0.0.0.0"
-    else:
-        host = netloc.lower()
+        host, port = "0.0.0.0", default_port
 
-    #get port
-    netloc = netloc.split(']')[-1]
-    if ":" in netloc:
-        port = netloc.split(':', 1)[1]
-        if not port.isdigit():
-            raise RuntimeError("%r is not a valid port number." % port)
+    try:
         port = int(port)
-    else:
-        port = default_port
-    return (host, port)
+    except ValueError:
+        raise RuntimeError("%r is not a valid port number." % port)
+
+    return host.lower(), port
 
 
 def close_on_exec(fd):
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -13,8 +13,12 @@ from urllib.parse import SplitResult
 @pytest.mark.parametrize('test_input, expected', [
     ('unix://var/run/test.sock', 'var/run/test.sock'),
     ('unix:/var/run/test.sock', '/var/run/test.sock'),
+    ('tcp://localhost', ('localhost', 8000)),
+    ('tcp://localhost:5000', ('localhost', 5000)),
     ('', ('0.0.0.0', 8000)),
     ('[::1]:8000', ('::1', 8000)),
+    ('[::1]:5000', ('::1', 5000)),
+    ('[::1]', ('::1', 8000)),
     ('localhost:8000', ('localhost', 8000)),
     ('127.0.0.1:8000', ('127.0.0.1', 8000)),
     ('localhost', ('localhost', 8000))


hramezani:
@berkerpeksag  thanks,
I think it is a cleaner and easy to underestanding implementation.
I think there is no improvement in speed and performance

hramezani:
@berkerpeksag, I changed the port conversation to int.

gunicorn/util.py
```diff
@@ -232,33 +232,29 @@ def is_ipv6(addr):
     return True
 
 
-def parse_address(netloc, default_port=8000):
+def parse_address(netloc, default_port='8000'):
+    host, port = [netloc, default_port]
```
tilgovi: I would move this down to an `else` statement. That would also make sure the `tcp://` stripping take effect. Right now, `tcp://` with an address but no port might not work.

tilgovi: I would move this down to an `else` statement or at least after these first two blocks. Probably an `else` because I like branching where all assignments are at the same level/depth. ;) That would also make sure the `tcp://` stripping take effect. Right now, `tcp://` with an address but no port might not work.

gunicorn/util.py
```diff
@@ -232,33 +232,29 @@ def is_ipv6(addr):
...
         return re.split(r'unix:(//)?', netloc)[-1]
 
     if netloc.startswith("tcp://"):
         netloc = netloc.split("tcp://")[1]
 
-    # get host
     if '[' in netloc and ']' in netloc:
-        host = netloc.split(']')[0][1:].lower()
+        host = netloc.split(']')[0][1:]
+        port = (netloc.split(']:') + [default_port])[-1]
```
tilgovi: I feel like this should be `[1]` instead of `[-1]`? We want _either_ the port that was specified in the netloc or the default port, right?

If we're rewriting this code I think we can make it less tricky by being a bit clearer about this stuff.

hramezani:
@tilgovi, thanks for your comments, I changed my code as your request and add some test for covering `tpc://` address

hramezani:
@tilgovi @benoitc any news?

gunicorn/util.py
```diff
@@ -232,33 +232,30 @@ def is_ipv6(addr):
     return True
 
 
-def parse_address(netloc, default_port=8000):
+def parse_address(netloc, default_port='8000'):
+
```
berkerpeksag: Style nit: No need to add a newline.

gunicorn/util.py
```diff
@@ -232,33 +232,30 @@ def is_ipv6(addr):
...
     if re.match(r'unix:(//)?', netloc):
         return re.split(r'unix:(//)?', netloc)[-1]
 
     if netloc.startswith("tcp://"):
         netloc = netloc.split("tcp://")[1]
 
-    # get host
+    host, port = [netloc, default_port]
```
berkerpeksag: No need to create a new list here. `host, port = netloc, default_port` should work, right?

gunicorn/util.py
```diff
@@ -232,33 +232,30 @@ def is_ipv6(addr):
...
-    else:
-        host = netloc.lower()
-
-    #get port
-    netloc = netloc.split(']')[-1]
-    if ":" in netloc:
-        port = netloc.split(':', 1)[1]
-        if not port.isdigit():
-            raise RuntimeError("%r is not a valid port number." % port)
+        host, port = ["0.0.0.0", default_port]
```
berkerpeksag:
```py
host, port = "0.0.0.0", default_port
```

gunicorn/util.py
```diff
@@ -232,33 +232,30 @@ def is_ipv6(addr):
...
+
+    try:
         port = int(port)
-    else:
-        port = default_port
-    return (host, port)
+    except ValueError:
+        raise RuntimeError("%r is not a valid port number." % port)
```
berkerpeksag: Since master now targets Gunicorn 20, perhaps we can raise `ValueError` as it's more correct exception to raise. @benoitc @tilgovi what do you think?

```py
except ValueError:
    raise ValueError("%r is not a valid port number." % port) from None
```

gunicorn/util.py
```diff
@@ -232,33 +232,30 @@ def is_ipv6(addr):
...
+
+    return (host.lower(), port)
```
berkerpeksag: Style nit:

```py
return host.lower(), port
```

hramezani:
@berkerpeksag thanks, I fixed all of your comments. Also rebase the PR with master

tests/test_util.py
```diff
@@ -24,7 +27,7 @@ def test_parse_address(test_input, expected):
 
 
 def test_parse_address_invalid():
-    with pytest.raises(RuntimeError) as err:
+    with pytest.raises(ValueError) as err:
```
benoitc: RuntimeError should be kept there as it used by the CLI.

gunicorn/util.py
```diff
@@ -247,33 +247,28 @@ def is_ipv6(addr):
...
     elif ':' in netloc:
-        host = netloc.split(':')[0].lower()
+        host, port = (netloc.split(':') + [default_port])[:2]
     elif netloc == "":
-        host = "0.0.0.0"
-    else:
-        host = netloc.lower()
-
-    #get port
-    netloc = netloc.split(']')[-1]
```
benoitc: but then what if a port different from the default  is given via the netloc ?

tests/test_util.py
```diff
@@ -13,8 +13,11 @@
 @pytest.mark.parametrize('test_input, expected', [
```
benoitc: tests are missing the case where the port in the netloc string is different from the default.

hramezani:
@benoitc thanks for the comment. I changed the `ValueError` to `RuntimeError`.
Also, I added a test for the case which port in `netloc`  is different from the default.

gunicorn/util.py
tilgovi: Sounds good to me!

tilgovi:
I think we should merge this as-is. If we would like to change the `RuntimeError` on master, we can do that separately. The patch is good the way it is, though, and doesn't change the exception behavior.

tilgovi:
Quiet for a while now and everything looks good. Merging! Thanks for your help and patience, @hramezani.

PR #1684, opened 2018-01-15
Fix 1683 handle malformed basic auth header

### LINKED ISSUE #1683: glogging._get_user() fails to catch malformed authorization header
I have a client that abuses the HTTP authorization header, passing "Basic (token)" where token is a random alphanumeric string. This causes the gunicorn logger to choke because, although it is a valid base64 string of data, it does not represent base64 encoded utf-8 string.

```
[2018-01-15 18:23:06 +0000] [15] [ERROR] Error handling request
  File "/app/.heroku/python/lib/python3.6/site-packages/gunicorn/glogging.py", line 269, in 
    'u': self._get_user(environ) or '-',
  File "/app/.heroku/python/lib/python3.6/site-packages/gunicorn/glogging.py", line 446, in 
    auth = auth.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 0: invalid start byte
```

You can reproduce it with the following line:
```base64.b64decode("ixsTtkKzIpVTncfQjbBcnoRNoDfbnaXG".encode('utf-8')).decode('utf-8')```

I don't expect _get_user() to work properly in this case but it would be good for it to catch the error. Currently the code is in a try/catch but it is only handling TypeError, and binascii.Error, it should also catch and handle UnicodeDecodeError. I'm working on a PR now, should be ready shortly.

jeiting:
Fixes #1683 

Added a unit test and tests pass on my branch.

Also tested against my server and confirmed the fix works.

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -471,10 +471,7 @@ class Logger(object):
                     if PY3:  # b64decode returns a byte string in Python 3
                         auth = auth.decode('utf-8')
                     auth = auth.split(":", 1)
-                except TypeError as exc:
-                    self.debug("Couldn't get username: %s", exc)
-                    return user
-                except binascii.Error as exc:
+                except (TypeError, binascii.Error, UnicodeDecodeError) as exc:
                     self.debug("Couldn't get username: %s", exc)
                     return user
                 if len(auth) == 2:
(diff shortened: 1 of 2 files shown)


berkerpeksag:
Thanks!

Issue #1662, opened 2017-12-10
couverture.io & coverage


benoitc:
I ws looking at the different PR and I see that most have couverture.io failing. I would like to make sure this is an issue on our own or not ? 

Are all the reports expected ? IMO if we activate such tool we should at least make sure to fix the issues found or we should disable it. What do you think? 

btw @aconrad when will the service go out of beta?

tilgovi:
No strong opinion. Lots of ways to do it:

- Coverage can never decrease
- Coverage can never fall below some threshold
- Coverage can never decrease by % in a commit

Generally, we have a question about testing policy.

aconrad:
Hi can you tell me which PRs are falling? I looked at a few and it seems to work for the few I picked.

aconrad:
I think what you were saying is that coverage reports are failing for many PRs. This is due to the fact that the gunicorn tests have a code path that runs at random and covers one or the other code paths in a non-deterministic way.

If you look at this coverage report:

https://app.couverture.io/coverage/github/benoitc/gunicorn/bfb3999646e92b2cd535dc2fccbf944455df1b98/py27

You’ll see that the code of the PR is covered but the file “gunicorn/http/message.py” runs random code paths, sometimes it hits line 215 and sometimes it hits line 218.

There was a conversation about it here:

https://github.com/benoitc/gunicorn/pull/1483

Once this is fixed, I think coverage reports will feel more “solid” from your perspective and will only fail for valid reasons. 👍

benoitc:
I don't understand that. If the code is covered, whatever the path it's executed on it should be covered.  

Anyway, so you're suggesting to only test a fixed set of data instead of random?

berkerpeksag:
I agree that removing randomness from `tests/treq.py` would be nice. However, the implementation of the `request` class is not trivial and the whole test suite probably needs some refactoring (we should provide a test framework before asking contributors write tests -- currently, it's not so easy to write tests for some parts of codebase) All current maintainers are volunteers and we all have lots of items in our TODO lists. I don't think we will have enough time to fix these issues in near future.

I also think that the messages like the following doesn't help the project:

![screenshot from 2018-01-11 08 10 06](https://user-images.githubusercontent.com/26338/34809572-e7e4ccc6-f6a6-11e7-8f67-7b2e3e27e5ff.png)

So I propose removing coverage tools until we solve the issues I mentioned above.

benoitc:
[...]

PR #1685, opened 2018-01-16
Temporarily disable code coverage reporting

aconrad:
Per the discussion in https://github.com/benoitc/gunicorn/issues/1662.

I also changed the travis directive `after_script` to `after_success` because we don't really need to submit coverage if the tests failed in the first place.


berkerpeksag:
Can't we just delete it? It's easy to `git revert` it and change `after_script` to `after_success` when we enable it back.

berkerpeksag:
Thanks!

Issue #1324, opened 2016-08-05
ValueError: bad marshal data (unknown type code)


nashuiliang:
run: `gunicorn --config __pycache__/a.gunicorn.conf.cpython-35.pyc a:app`
version: `python3.5.2`
err: 

```
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/gunicorn/app/base.py", line 95, in get_config_from_filename
    execfile_(filename, cfg, cfg)
  File "/usr/lib/python3.5/site-packages/gunicorn/_compat.py", line 69, in execfile_
    code = _get_codeobj(fname)
  File "/usr/lib/python3.5/site-packages/gunicorn/_compat.py", line 53, in _get_codeobj
    code_obj = marshal.loads(data[8:])
ValueError: bad marshal data (unknown type code)
```

when I run in python2, it's right.
From the article (PEP 3147) https://www.python.org/dev/peps/pep-3147/#background, **Byte code files contain two 32-bit big-endian numbers followed by the marshaled code object.**

---

in python3.5, I change the file _/usr/lib/python3.5/site-packages/gunicorn/_compat.py_ at   **Line 32**  to `code_obj = marshal.loads(data[12:])`. It's ok.

benoitc:
@berkerpeksag any idea? did something changed ?

@nashuiliang how did you install the package? On which machine?

berkerpeksag:
The OP's fix looks good to me.

nashuiliang:
`# uname -a`
> Linux arch 4.6.4-1-ARCH #1 SMP PREEMPT Mon Jul 11 19:12:32 CEST 2016 x86_64 GNU/Linux

`# pip3 install gunicorn`

tilgovi:
If we can fix it so that it works on on Python2 and Python3 then we should do it!

benoitc:
@berkerpeksag what's the correct way to check the python version there ? will it be for anything > 3.5 ?

berkerpeksag:
I think the easiest way to check it would be

```
header_size = 12 if PY3 else 8
```

I haven't had a chance to test it, but I think the same line would fail any Python 3 version newer than 3.2.

I'll try to address this before the next release.

dmiyakawa:
It seems pyc after 3.3 has 12bytes header instead of 8. See the following for more detail.

* http://qiita.com/amedama/items/698a7c4dbdd34b03b427
[...]

PR #1696, opened 2018-02-03
Issue #1324: Fix passing pyc files to --config in Python 3.3+

hramezani:
fix this issue base on @berkerpeksag [comment](https://github.com/benoitc/gunicorn/issues/1324#issuecomment-254093136), and add some test to check it.

### DIFF
--- a/gunicorn/_compat.py
+++ b/gunicorn/_compat.py
@@ -50,7 +50,8 @@ def _get_codeobj(pyfile):
         # So to get code object, just read the 8th byte onwards till EOF, and
         # UN-marshal it.
         import marshal
-        code_obj = marshal.loads(data[8:])
+        header_size = 12 if PY33 else 8 # TODO Remove in 20.0
+        code_obj = marshal.loads(data[header_size:])
 
     elif result is PY_SOURCE:
         # This is a .py file.
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,6 +5,7 @@
 
 import os
 import sys
+import py_compile
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


tests/test_config.py
```diff
@@ -22,6 +22,16 @@ def alt_cfg_module():
     return 'config.test_cfg_alt'
 def cfg_file():
     return os.path.join(dirname, "config", "test_cfg.py")
+def pyc_cfg_file():
+    PY33 = (sys.version_info >= (3, 3))
```
tilgovi: Can we use the `py_compile.compile()` function instead, so that this is totally portable and we don't have to check in bytecode?

hramezani:
@tilgovi thanks for your comment, I fixed the test as you mentioned in your comment

gunicorn/_compat.py
```diff
@@ -50,7 +50,8 @@ def _get_codeobj(pyfile):
         # So to get code object, just read the 8th byte onwards till EOF, and
         # UN-marshal it.
         import marshal
-        code_obj = marshal.loads(data[8:])
+        header_size = 12 if PY33 else 8
```
berkerpeksag: Could you add a `# TODO` comment so we don't forget to remove this check when we start to develop 20.0?

tests/test_config.py
```diff
@@ -14,6 +14,7 @@
 from gunicorn.workers.sync import SyncWorker
 from gunicorn import glogging
 from gunicorn.instrument import statsd
+import py_compile
```
berkerpeksag: Style nit: Please group stdlib modules (i.e. move this to line line)

hramezani:
@berkerpeksag thanks, changes done

tests/test_config.py
```diff
@@ -22,6 +23,11 @@ def alt_cfg_module():
     return 'config.test_cfg_alt'
 def cfg_file():
     return os.path.join(dirname, "config", "test_cfg.py")
+def pyc_cfg_file():
+    file_path = os.path.join(dirname, "config", "test_cfg.py")
+    pyc_file_path = file_path + 'c'
+    py_compile.compile(file_path, pyc_file_path)
```
berkerpeksag: I think we also need to delete `pyc_file_path` after the test ran. We could use https://docs.pytest.org/en/latest/xunit_setup.html to delete it.

hramezani:
@berkerpeksag, thanks. please check my new changes.

benoitc:
do we really need to support compiled config file though? That sounds rather complicated for me and break the contract of graceful reloads... I would rather not support it. What are the reasons to have it?

berkerpeksag:
It looks like at least one user uses it (https://github.com/benoitc/gunicorn/issues/1324#issue-169532465) but I'm OK to deprecate it in the next release.

benoitc:
I think we can close that issue now. Next master will only python 3 and sup version. Also I would be in favour to deprecate such option. Thoughts ? cc @berkerpeksag @tilgovi

PR #1699, opened 2018-02-06
Avoid 'Connection refused' when using gthread with keep-alive

### LINKED ISSUE #1698: "Connection refused" when using keep-alive with gthread
Gunicorn version: 19.7.1 (also tried master branch)

This bug could be reproduced with attached files:
- test_http_gunicorn_raw_keep-alive-requests.py (test runner)
- myapp.py (gunicorn app that should be in the same directory)

Test case starts command:
```
gunicorn --worker-class gthread --workers 1 --threads 20 --keep-alive 9999 \
           --log-level DEBUG --bind localhost:7777 myapp:app
```
with logs redirected to `/tmp/_test_gunicorn.out`
and then create separate threads to open http session and send 1000 requests within it.

In my case gunicorn server received request `/request/1096` and then reset connection (see `tcp.stream eq 10` from attached `gunicorn_reset_keep_alived_connection.pcapng`

NOTE: this is race condition so it may happen that all requests finish with success (see myapp.py -> without `time.sleep` it almost never fails) but with that `time.sleep` it usually fails and then script should output line similar to:
`[1] Failed request id=1096 with ('Connection aborted.', error(104, 'Connection reset by peer'))`
it mean that thread sent request `GET /request/1096` and received `Connection reset by peer` (so the last one request from that session that succeed is `GET /request/1095`).

ddzialak:
Fixes #1698

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -74,11 +74,6 @@ class TConn(object):
     def close(self):
         util.close(self.sock)
 
-    def __lt__(self, other):
-        return self.timeout < other.timeout
-
-    __cmp__ = __lt__
-
 
 class ThreadWorker(base.Worker):
 


benoitc:
Thanks for the patch! Looking :)

ddzialak:
any update?

ddzialak:
I've added two commits to make added attribute corresponding to it's name, these commits don't affect solution for `connection refused` problem. Now `is_active` should be True if it's processed and should not be present on self._keep and `False` if processing is done and should be present on self._keep. Please squash these commits if everything is ok.

ddzialak:
@tilgovi That was my first try and after that received connection refused again but most probably I've done something wrong with testing..

ddzialak:
Now it's clean why moving whole `murder_keepalived` under lock doesn't solved that issue. Other important condition is that bug may happen only with Python2 because `__cmp__` in python2 is used to override `==` operator while function `__lt__` in Python3 doesn't.

ddzialak:
any update?

benoitc:
Sorry for the delay. looks good, merging it now :) Thanks !

ddzialak:
Thanks ☺️

ddzialak:
@benoitc What is the estimated time when this change is pushed to PyPI?

PR #1705, opened 2018-02-16
Better docs for statsd prefix

confiq:
This tiny text made me look at code only to figure out it is possible...

Might help somebody else!


tilgovi:
I saw you added a new commit. I don't think more text was bad. I like it either way. There's extra whitespace now, though.

confiq:
>  I don't think more text was bad.

Honestly, I'm not native English speaker so I didn't want to make it incomprehensible; therefore I've removed the text. 

I do think that link and the example how to use `statsd-prefix` can help someone else like me next time.

tilgovi:
Looks good. Thank you for making the Gunicorn docs better!

PR #1718, opened 2018-03-07
Add extras_require for various worker types

### LINKED ISSUE #1717: Use extras_require in setup for optional dependencies
Gunicorn [may require extra dependencies](http://docs.gunicorn.org/en/stable/install.html#async-workers) to be installed, but they seem to be optional. It looks like it should be possible to install these using [setuptools's extra](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies).

It looks like Gevent or Eventlet are 2 options, but the [worker option](http://docs.gunicorn.org/en/stable/run.html#commonly-used-arguments) seems to indicate there are more.

Would it make sense to add this? It has the advantage of documenting indirect dependencies.

browniebroke:
Fixes #1717

- [x] Add [`extras_require`](https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) to `setup.py`
- [x] Update documentation

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -33,12 +33,12 @@ want to consider one of the alternate worker types.
 
 ::
 
-    $ pip install greenlet  # Required for both
-    $ pip install eventlet  # For eventlet workers
-    $ pip install gevent    # For gevent workers
+    $ pip install gunicorn[eventlet]  # For eventlet workers
+    $ pip install gunicorn[gevent]    # For gevent workers
 
 .. note::
-    If installing ``greenlet`` fails you probably need to install
+    Both require ``greenlet``, which should get installed automatically,
+    If its installation fails, you probably need to install
     the Python headers. These headers are available in most package
     managers. On Ubuntu the package name for ``apt-get`` is
     ``python-dev``.
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -33,8 +33,11 @@ want to consider one of the alternate worker types.
 
 ::
 
-    $ pip install gunicorn[eventlet]  # For eventlet workers
-    $ pip install gunicorn[gevent]    # For gevent workers
+    $ pip install greenlet            # Required for both
+    $ pip install eventlet            # For eventlet workers
+    $ pip install gunicorn[eventlet]  # Or, using extra
+    $ pip install gevent              # For gevent workers
+    $ pip install gunicorn[gevent]    # Or, using extra
 
 .. note::
     Both require ``greenlet``, which should get installed automatically,
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -119,10 +119,12 @@ you might want to choose one of the other worker classes.
 A string referring to one of the following bundled classes:
 
 * ``sync``
-* ``eventlet`` - Requires eventlet >= 0.9.7
-* ``gevent``   - Requires gevent >= 0.13
-* ``tornado``  - Requires tornado >= 0.2
-* ``gthread``  - Python 2 requires the futures package to be installed
+* ``eventlet`` - Requires eventlet >= 0.9.7 (or install using
+  gunicorn[eventlet])
+* ``gevent``   - Requires gevent >= 0.13 (or install using gunicorn[gevent])
+* ``tornado``  - Requires tornado >= 0.2 (or install using gunicorn[tornado])
+* ``gthread``  - Python 2 requires the futures package to be installed (or
+  install using gunicorn[gthread])
 * ``gaiohttp`` - Deprecated.
 
 Optionally, you can provide your own worker by giving Gunicorn a
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -115,16 +115,19 @@ The type of workers to use.
 The default class (``sync``) should handle most "normal" types of
 workloads. You'll want to read :doc:`design` for information on when
 you might want to choose one of the other worker classes. Required
-libraries may be installed using setuptools' extra_require feature.
+libraries may be installed using setuptools' ``extra_require`` feature.
 
 A string referring to one of the following bundled classes:
 
 * ``sync``
-* ``eventlet`` - Requires eventlet >= 0.9.7 (extra gunicorn[eventlet])
-* ``gevent``   - Requires gevent >= 0.13 (extra gunicorn[gevent])
-* ``tornado``  - Requires tornado >= 0.2 (extra gunicorn[tornado])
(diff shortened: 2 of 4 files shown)


docs/source/settings.rst
```diff
@@ -119,10 +119,12 @@ you might want to choose one of the other worker classes.
 A string referring to one of the following bundled classes:
 
 * ``sync``
-* ``eventlet`` - Requires eventlet >= 0.9.7
-* ``gevent``   - Requires gevent >= 0.13
-* ``tornado``  - Requires tornado >= 0.2
-* ``gthread``  - Python 2 requires the futures package to be installed
+* ``eventlet`` - Requires eventlet >= 0.9.7 (or install using
```
berkerpeksag: Could you please make these changes in `gunicorn/config.py` and then run `make html` in `docs` directory? (and please  don't include unrelated changes to `docs/source/settings.rst` after you ran `make html`)

setup.py
```diff
@@ -72,6 +72,16 @@ def run_tests(self):
         errno = pytest.main(self.test_args)
         sys.exit(errno)
 
+
+extra_require = {
+    'gevent':  ["gevent>=0.13"],
+    'eventlet': ["eventlet>=0.9.7"],
+    'tornado': ["tornado>=0.2"],
+    'gthread': []
```
berkerpeksag: Style nit: Please add trailing comma.

setup.py
```diff
@@ -72,6 +72,16 @@ def run_tests(self):
         errno = pytest.main(self.test_args)
         sys.exit(errno)
 
+
+extra_require = {
+    'gevent':  ["gevent>=0.13"],
```
berkerpeksag: Style nit: We usually try to stick with the most used quoting style in a file. For `setup.py`, it looks like we went with single quotes. It would be nice use it in new code.

setup.py
```diff
@@ -98,5 +108,6 @@ def run_tests(self):
 
     [paste.server_runner]
     main=gunicorn.app.pasterapp:paste_server
-    """
+    """,
+    extras_require=extra_require
```
berkerpeksag: Style nit: Please add trailing comma.

docs/source/install.rst
```diff
@@ -33,12 +33,12 @@ want to consider one of the alternate worker types.
 
 ::
 
-    $ pip install greenlet  # Required for both
-    $ pip install eventlet  # For eventlet workers
-    $ pip install gevent    # For gevent workers
+    $ pip install gunicorn[eventlet]  # For eventlet workers
+    $ pip install gunicorn[gevent]    # For gevent workers
```
benoitc: i’m not sure about this changes. some people are installing gunicorn with both workers support and not always at the same time. I would suggest to keep it like this and maybe add a new line.

docs/source/settings.rst
benoitc: indeed that should be done in settings.py in any case

browniebroke: > please don't include unrelated changes to `docs/source/settings.rst` after you ran `make html`

Looking at the generated `settings.rst`, I'm not sure what you mean. Do you mean I should not commit it at all?

berkerpeksag: After you ran `make html`, if you do `git diff` on `settings.rst` you will see changes like:

```diff
diff --git a/docs/source/settings.rst b/docs/source/settings.rst
index 4a8d027..738e18a 100644
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -450,7 +450,7 @@ chdir
 ~~~~~
 
 * ``--chdir``
-* ``/Users/benoitc/work/gunicorn/py27/gunicorn/docs``
+* ``/home/berker/projects/gunicorn/docs``
 
 Chdir to specified directory before apps loading.
 
@@ -521,7 +521,7 @@ user
 ~~~~
 
 * ``-u USER, --user USER``
-* ``501``
+* ``1000``
 
 Switch worker processes to run as this user.
 
@@ -535,7 +535,7 @@ group
 ~~~~~
 
 * ``-g GROUP, --group GROUP``
-* ``20``
+* ``1000``
 
 Switch worker process to run as this group.
 
@@ -594,7 +594,7 @@ temporary directory.
 secure_scheme_headers
 ~~~~~~~~~~~~~~~~~~~~~
 
-* ``{'X-FORWARDED-PROTOCOL': 'ssl', 'X-FORWARDED-PROTO': 'https', 'X-FORWARDED-SSL': 'on'}``
+* ``{'X-FORWARDED-PROTO': 'https', 'X-FORWARDED-SSL': 'on', 'X-FORWARDED-PROTOCOL': 'ssl'}``
 
 A dictionary containing headers and values that the front-end proxy
 uses to indicate HTTPS requests. These tell Gunicorn to set
@@ -763,13 +763,30 @@ The log config file to use.
 Gunicorn uses the standard Python logging module's Configuration
 file format.

```

I was talking about these unrelated changes in my previous comment. They are created by `docs/gunicorn_ext.py` but we usually discard them when we generate `settings.rst` for documentation changes.

browniebroke: Ok, thanks for the clarification. Should be better now.

gunicorn/config.py
```diff
@@ -605,15 +605,17 @@ class WorkerClass(Setting):
 
         The default class (``sync``) should handle most "normal" types of
         workloads. You'll want to read :doc:`design` for information on when
-        you might want to choose one of the other worker classes.
+        you might want to choose one of the other worker classes. Required
+        libraries may be installed using setuptools' extra_require feature.
```
berkerpeksag: extra_require -> \`\`extra_require\`\`

gunicorn/config.py
```diff
@@ -605,15 +605,17 @@ class WorkerClass(Setting):
...
 
         A string referring to one of the following bundled classes:
 
         * ``sync``
-        * ``eventlet`` - Requires eventlet >= 0.9.7
-        * ``gevent``   - Requires gevent >= 0.13
-        * ``tornado``  - Requires tornado >= 0.2
+        * ``eventlet`` - Requires eventlet >= 0.9.7 (extra gunicorn[eventlet])
```
berkerpeksag: Maybe something like "(or install it via \`\`gunicorn[...]\`\`)" or "(or install it via \`\`pip install gunicorn[...]\`\`)" would make it clearer.

browniebroke: Thanks for the suggestion, I wasn't happy with my wording myself.

berkerpeksag:
Thank you and congratulations for your first contribution to Gunicorn!

Issue #1719, opened 2018-03-09
resolving hostnames for bind


wanneut:
gunicorn takes a list of hostnames, checks for everyone if it is a IPv6 address and if it is not, it binds to the IPv4 address of the hostname.
The relevant parts are:
https://github.com/benoitc/gunicorn/blob/d1f5268b150aec65067e4766a61eab2bd25526ea/gunicorn/config.py#L125
https://github.com/benoitc/gunicorn/blob/d1f5268b150aec65067e4766a61eab2bd25526ea/gunicorn/sock.py#L132
https://github.com/benoitc/gunicorn/blob/d1f5268b150aec65067e4766a61eab2bd25526ea/gunicorn/util.py#L227

This makes two wrong assumptions: 
- Every  hostname has exact one IP-address.
- A hostname can not resolve to an IPv6 address.

Even for the most common example localhost this assumptions are not true:
localhost resolves on the most systems to ::1 and 127.0.0.1. But gunicorn binds only to 127.0.0.1 (the first IPv4 address)
If a standards-compliant software tries to connect to gunicorn on localhost it will at first try to connect to ::1 and only after that times out to 127.0.0.1.
So most of my clients disable IPv6 to make gunicorn react faster…

A much better approach would be to loop over all results of getaddrinfo and just bind to whatever family it returns:
https://docs.python.org/2.5/lib/socket-example.html

tilgovi:
Thank you for reporting this issue! If you would like to make a PR, I'd be happy to review it with you.

tilgovi:
It seems like we might have to change that first comprehension to complex comprehension and have `parse_address` yield one or multiple addresses. I am open to other approaches, but that seems the least invasive.

wanneut:
Tried to make a nicer approach to handle different family types there https://github.com/wanneut/gunicorn
But there is still a little work to do. (Some former tests fail, because they are still using is_ipv6... And I still don't understand how you close sockets.)
I am not that fit in python programming. Can you have a look?

tilgovi:
@wanneut I took a look and left some style comments, but I like your approach! I won't be able to look at tests until later today, but this is great. Thank you. You should be able to close sockets with `.close()`.

tilgovi:
We also have a helper in `gunicorn.util` to close a socket and catch any error when you don't care about errors.

wanneut:
I start adopting the tests.

gozdal:
[...]

PR #1723, opened 2018-03-14
Issue #1719: resolving hostnames for bind

wanneut:
Changes proposed there: https://github.com/benoitc/gunicorn/issues/1719 .

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -125,7 +125,7 @@ class Config(object):
         
         addresses = []
         for bind in s:
-            addresses=addresses+util.parse_address(_compat.bytes_to_str(bind))
+            addresses.extend(util.parse_address(_compat.bytes_to_str(bind)))
         return addresses
 
     @property
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -122,7 +122,11 @@ class Config(object):
     @property
     def address(self):
         s = self.settings['bind'].get()
-        return [util.parse_address(_compat.bytes_to_str(bind)) for bind in s]
+        
+        addresses = []
+        for bind in s:
+            addresses=addresses+util.parse_address(_compat.bytes_to_str(bind))
+        return addresses
 
     @property
     def uid(self):
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -159,7 +159,8 @@ def create_sockets(conf, log, fds=None):
         for fd in fds:
             sock = socket.fromfd(fd, socket.AF_UNIX, socket.SOCK_STREAM)
             sock_name = sock.getsockname()
-            sock_type = _sock_type(sock_name)
+            sockinfo = (sock.family, sock.type, sock.proto, '', socket.getsockname())
+            sock_type = _sock_type(sockinfo)
             listener = sock_type(sock_name, conf, log, fd=fd)
             listeners.append(listener)
 
@@ -197,6 +198,7 @@ def create_sockets(conf, log, fds=None):
 def close_sockets(listeners, unlink=True):
     for sock in listeners:
         sock_name = sock.getsockname()
+        socket_family = sock.family
         sock.close()
-        if unlink and _sock_type(sock_name) is UnixSocket:
+        if unlink and socket_family == socket.AF_UNIX:
             os.unlink(sock_name)
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -15,17 +15,17 @@ from gunicorn.six import string_types
 
 
 class BaseSocket(object):
-
+    
     def __init__(self, address, conf, log, fd=None):
         self.log = log
         self.conf = conf
 
-        self.cfg_addr = address
+        self.cfg_addr = address[4]
         if fd is None:
-            sock = socket.socket(self.FAMILY, socket.SOCK_STREAM)
+            sock = socket.socket(address[0], socket.SOCK_STREAM)
             bound = False
         else:
-            sock = socket.fromfd(fd, self.FAMILY, socket.SOCK_STREAM)
+            sock = socket.fromfd(fd, address[0], socket.SOCK_STREAM)
             os.close(fd)
             bound = True
 
@@ -74,8 +74,6 @@ class BaseSocket(object):
 
 class TCPSocket(BaseSocket):
 
-    FAMILY = socket.AF_INET
-
     def __str__(self):
         if self.conf.is_ssl:
             scheme = "https"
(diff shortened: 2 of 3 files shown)


wanneut:
I don't understand what this should do:
https://github.com/benoitc/gunicorn/blob/d1f5268b150aec65067e4766a61eab2bd25526ea/gunicorn/util.py#L247
netloc has should always have a value which is set Bind:
https://github.com/benoitc/gunicorn/blob/d1f5268b150aec65067e4766a61eab2bd25526ea/gunicorn/config.py#L536

gunicorn/config.py
```diff
@@ -538,9 +538,9 @@ class Bind(Setting):
     validator = validate_list_string
 
     if 'PORT' in os.environ:
-        default = ['0.0.0.0:{0}'.format(os.environ.get('PORT'))]
+        default = ['[::]:{0}'.format(os.environ.get('PORT'))]
```
tilgovi: At least on my machine, this returns only IPv6 addresses from `socket.getaddrinfo()`. Since this is a list, maybe we want to give it two entries so Gunicorn binds to IPv4 and IPv6 by default?

tilgovi:
> I don't understand what this should do

I think it is just being safe and not relying on the config code to set it.

However, on my machine, an empty host does what I wanted from the default config where I commented:

``` python
>>> socket.getaddrinfo('', 80, proto=socket.IPPROTO_TCP)
[(<AddressFamily.AF_INET6: 30>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('::1', 80, 0, 0)), (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('127.0.0.1', 80))]
```

So maybe the right default is the empty string and we don't need to handle it specially in `parse_address`?

tilgovi:
Oh, I'm sorry, that's a good default when we don't automatically default to binding to every interface (`PORT` is undefined).

gunicorn/sock.py
```diff
@@ -166,7 +159,8 @@ def create_sockets(conf, log, fds=None):
         for fd in fds:
             sock = socket.fromfd(fd, socket.AF_UNIX, socket.SOCK_STREAM)
             sock_name = sock.getsockname()
-            sock_type = _sock_type(sock_name)
+            sockinfo = (sock.family, sock.type, sock.proto, '', socket.getsockname())
+            sock_type = _sock_type(sockinfo)
```
berkerpeksag: Why did you pass a tuple here? `_sock_type()` only uses the first element of the tuple `sockinfo`. Can't we just pass `sock.family`?

gunicorn/config.py
wanneut: If you bind to :: you will bind to any address  (IPv4 and IPv6) as long the IPV6_V6ONLY flag is not set.

wanneut:
I don't think that 0.0.0.0 is a good default. localhost is much better. I don't like programs witch listen to the world by default while it is very common to configure them insecure.
Think there are many gunicorn instances without security measures.

gunicorn/sock.py
wanneut: It was done so before. But I can change that.

wanneut:
So now all test pass. (And I added a view.) I think all cases are handled properly now.

wanneut:
OK, not on Ubuntu Trusty Tahr (from 2014). On debian jessie (from 2015) and stretch all checks pass so it is on CentOS 7 and Ubuntu Bionic Beaver.

benoitc:
_One question I have twith this PR is how it will work with OSes that doesn't support or have not been started with IPV6 support. I do think we still need to make sure ipv6 is supported on the OS somehow.

wanneut:
I think it works perfect. It just binds to IPv4. Only some tests fail. Tried a view examples on a VM with disabled IPv6 in /proc/sys/net/.

benoitc:
mm if ipv6 is not enabled at all on the machine at the kernel level it shouldn’t work normally. i will test anyway

gunicorn/config.py
```diff
@@ -534,9 +538,9 @@ class Bind(Setting):
...
```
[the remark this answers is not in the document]
wanneut: The older version dosn't work on IPv6 installations at all.
This version, works everywhere (IPv4 like IPv6) exempt of some obscure Linux-configurations (I think there are not many other OSes where something can be done.), where someone explicitly removed IPv6 support. I think such users should know how to specify another bind address.
If someone shoots himself into his foot he should not wonder if it hurts.
Gunicorn requires Python 2 (which is newer than IPv6.) multithreading and several other things which are much more uncommon than IPv6.
At the end you have to decide what your defaults are.
I mean what about installations with disabled IPv4? It is deprecated for 20 years now. Time to turn it of?

gunicorn/config.py
```diff
@@ -122,7 +122,11 @@ def worker_class(self):
     @property
     def address(self):
         s = self.settings['bind'].get()
-        return [util.parse_address(_compat.bytes_to_str(bind)) for bind in s]
+
```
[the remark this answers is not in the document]
wanneut: It was described at the beginning. It can handle only one address.

wanneut:
It works not if you set the PORT environment variable but no bind-address. The defaults work. If you set a IPv4 bind-address it woks also.

wanneut:
I think the TOXENV=lint check is broken.

tilgovi:
@wanneut why are some of the test cases commented out?

wanneut:
On almost all OSes (even older ones) localhost resolves to ::1 and 127.0.0.1. So I used this to test an address that is known to be dual stack. But this seems not to be true for Ubuntu Trusty Tahr. Since this is the environment where the tests are executed I commented them out.
Yesterday there was a release of an new Ubuntu LTS version. So may the testing environment is updated and the Tests can be used again.
Another option would be to replace localhost by a "real" domain but this would require a working Internet connection for testing. I don't know if that is wanted.
So I just commented them out and hoped that someone else who reads it decides what to do.
Btw.: I thought about testing the protocol specific localhost addresses but this is even worse for testing on different OSes. On some they are named ip4-loopback and ip6-loopback on some localhost4 on some localhost-v4 on some there are only the v6 specific and the v4 specific missing...

tilgovi:
If these tests passed before, they should pass now, unless behavior has changed. Where the behavior has changed deliberately, can we update the tests to show what the new expectation is?

I would love to get this merged.

wanneut:
Is anybody still following this?

tilgovi:
@wanneut thanks for the nudge. I'll try to pick this up.

tilgovi:
I'll pick this up again for the next cycle after R20. Sorry for the delay and thank you for your patience and efforts.

rivimey:
bump - is this still relevant? I'm interested in getting gunicorn listening on ipv6 for a dual-family host.

benoitc:
Closing due to inactivity. This PR has been open for nearly 6 years without resolution on the IPv6 design questions and test issues. If this feature is still needed, please feel free to open a fresh PR against current master. Thank you for your contribution!

Issue #1634, opened 2017-10-26
release 19.8.0


benoitc:
As I would like to release a 19.8 version this week. I don't see anything preventing it right now. Tests pass here but let me know if I missed something.

benoitc:
cc @tilgovi @berkerpeksag

berkerpeksag:
I would like to get the following PRs merged for 19.8:

* #1499 (I can address my own comments if we decide to merge this one)
* #1569 (will merge this tonight)
* #1632 (reviewed)

benoitc:
@berkerpeksag i commented/approved then. Let me know if you need some help as I may have some cycles tomorrow :)

tilgovi:
I have open PRs. Do we want to merge any of them?

- #1481 (allow Python config files to set chdir)
- #1602 (I can address comments from @berkerpeksag)

berkerpeksag:
Both PRs good to have if we have time to get them in.

tilgovi:
Mine are in. Can I help with reviewing anything?

berkerpeksag:
#1499 is the only remaining one. I've addressed my review comments.

benoitc:
@berkerpeksag i've commented also, what do you think about it?

benoitc:
@berkerpeksag i didn't get an answer about #1499 . I'm dubious about putting back a eature that have been removed a long itme ago. We should see the pro/cons about it. Or at least makng sure it has no impact. 

I think however that before releasing i will disable by defaut the SO_REUSEPORT option of the socket. Thoughts?

RonRothman:
@berkerpeksag I'm not sure I understand - why would we want to change the default for `SO_REUSEPORT` based on that PR?

(We do rely on `SO_RESUSEPORT` to help us efficiently serve tens of thousands of requests/sec, so I'm careful around changes to it.)

tilgovi:
Having SO_REUSEPORT as a default seems to confused some people, especially those new to Gunicorn or using it in a development environment. It makes it possible to start a new Gunicorn without killing the old one, resulting in multiple versions of an application running. This is not a very friendly experience for a first time user who wonders why their application serves the right response only half of the time.

We will definitely not remove the functionality even if we change the default.

berkerpeksag:
> @berkerpeksag i didn't get an answer about #1499 . I'm dubious about putting back a eature that have been removed a long itme ago. We should see the pro/cons about it. Or at least makng sure it has no impact.

[...]

PR #1727, opened 2018-03-19
Add PyPI badges to README

hugovk:
As suggested in https://github.com/benoitc/gunicorn/issues/1634:

> Pypi seems to be still at 19.7.1 https://pypi.python.org/pypi/gunicorn
As a side thought could the README.md be made to have stable version number?
going trough pypi to check what is the current release feels bit funky, or perhaps I'm missing some obvious version number displayed somewhere 😕

Preview:

image

Also upgrade the Travis badge to SVG which looks better, and whilst we're here, update HTTP links to HTTPS.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,7 +58,7 @@ License
 Gunicorn is released under the MIT License. See the LICENSE_ file for more
 details.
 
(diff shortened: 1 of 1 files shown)


berkerpeksag:
Thanks.

PR #1741, opened 2018-04-02
Documentation. Fixed test app.

djudman:


tilgovi:
It is easier to just use a bytes literal (`b'Hello, world!\n`). Gunicorn only supports Python 2.6+. You can avoid the version check that way.

djudman:
@tilgovi you are right, thanks. I missed that Gunicorn only supports Python 2.6+. Pushed.

PR #1754, opened 2018-04-23
added on_reload_spawn hook

tarabanton:
We have a pretty "fat" app which takes a lot of time to load.
Simple reload via SIGHUP is not enough, because gunicorn kills old workers before new ones loads up completely.
We have added workaround by adding hook call after spawn_worker() and before manage_workers() in reload().

Simple example:
```
def post_reload_spawn(server):
    server.log.info("Forked workers. Waiting 30 seconds...")
    time.sleep(30);
    pass
```
If we are inventing something that already exists, please direct us in the right way.
Thx.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -480,6 +480,9 @@ class Arbiter(object):
         for _ in range(self.cfg.workers):
             self.spawn_worker()
 
+        # call post_reload_spawn hook
+        self.cfg.post_reload_spawn(self)
+
         # manage workers
         self.manage_workers()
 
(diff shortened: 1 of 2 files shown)


tilgovi:
The location of this hook (after spawning, before killing) feels very specific to this use case. It suggests a need to address the problem in a different way.

You can use SIGUSR2 to start a new master with new workers without shutting down the old. Then, you can shut down the old master once the new workers have had time to warm up.

tilgovi:
It feels like a common enough use case that I've seen several requests for something similar. Also, we have some requests in the past for boot timeouts. I think we should continue the discussion in #1757 and close this PR. I would rather address the need directly, if it's easy enough to do, than create another hook.

tilgovi:
@tarabanton can you or any of the six people who 👍 your comment please comment on #1757? Do you need a hook, or do you specifically need a pause?

benoitc:

PR #1762, opened 2018-04-28
Ignore .pytest_cache

hugovk:
Pytest now uses a .pytest_cache directory. Ignore it.

> The default cache directory has been renamed from `.cache` to `.pytest_cache` after community feedback that the name `.cache` did not make it clear that it was used by pytest. ([#3138](https://github.com/pytest-dev/pytest/issues/3138))

https://docs.pytest.org/en/latest/changelog.html#pytest-3-4-0-2018-01-30


berkerpeksag:
Thanks for the PR, but we usually prefer keeping things like this in a user's own [global `.gitignore` file](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore). We have cleaned up our  `.gitignore` file to encourage people to use a global one in commit 4185b2bd2789b33c229ad9be87f192d88bf9cd46.

tilgovi:
In this case, pytest is in use by the project and it's not user specific. I'd be fine with this.

berkerpeksag:
@tilgovi fine by me, I don't have a strong objection. I already added it to my own global .gitignore file since I use pytest in multiple projects.

tilgovi:
Merged it :). Thanks.

Issue #1755, opened 2018-04-24
Gunicorn w/ Django on nginx + unix socket: REMOTE Adress is "b''" (a string containing the three characters b'')


anx-ckreuzberger:
I just had a huge "wtf moment" when I was setting up a Django Middleware to parse the ``HTTP_X_FORWARDED_FOR`` header. For security reasons (more a precaution) my middleware would check the content of the ``REMOTE_ADDR`` header, to verify that the request is actually coming from a trusted proxy (e.g., 127.0.0.1).

Now, as already discussed in issue #797 gunicorn does not provide the ``REMOTE_ADDR`` anymore when used with an unix socket, and instead an empty string is being sent (see comment by @benoitc on Oct 9, 2014: "To summarise the context, the remote address is right now empty by default when using the unix socket. ").

So I thought I'll just make this little thing in a Django Middleware:
```python
if request.META['REMOTE_ADDR'] == "": # empty string
    request.META['REMOTE_ADDR'] = request.META['HTTP_X_FORWARDED_FOR'].split(",")[0].strip()
```

Oh well...  It seems that I was wrong.

Would you like to know what ``request.META['REMOTE_ADDR']`` contains? It's actually the three characters ``b''`` (no, not the byte representation b'', but actually the three characters!).

I believe this is not the intended behaviour, though I'm not involved enough with the whole gunicorn/nginx unix socket setup to point out where the error might be.

Here are some Details of the software that is being used:

* Debian
* Python 3.5.3
* gunicorn 19.7.1
* nginx 1.10.3
* Django 1.11.12

Gunicorn is started via systemd with the following command:
``gunicorn --workers 3 --bind unix:/run/gunicorn/socket application.wsgi``

Nginx is configured as follows:
```
upstream web {
        ip_hash;
        server unix:/run/gunicorn/socket fail_timeout=0;
}

        location @proxy {
                # proxy stuff
                proxy_read_timeout 600;
                proxy_connect_timeout 600;
                proxy_pass_request_headers on;
                proxy_set_header        X-Real-IP $remote_addr;
                proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_set_header        Host $http_host;
                proxy_pass              http://web;
                proxy_redirect          off;
                # tell the upstream that we are using https (or whatever scheme)
                proxy_set_header        X-Forwarded-Proto $scheme;
        }

```

anx-ckreuzberger:
[...]

PR #1769, opened 2018-05-02
Fix for #1755: Decode remote addr if it is a byte

anx-ckreuzberger:
After some debugging in issue #1755, I found that when using 

* Python 3.5 (<= 3.5.3)
* Python 3.6 (<= 3.6.2)

pythons built-in socket.getsockname returns ``b''`` as the socket name for unix sockets (see [Python Issue 30205](https://bugs.python.org/issue30205)).

This is problematic with gunicorn, as the ``REMOTE_ADDR`` is set to ``"b''``, which is caused by the following lines of code:
https://github.com/benoitc/gunicorn/blob/73cc860ead2e926e4d11bcf0cde0037dec8805ef/gunicorn/http/wsgi.py#L154-L164

Specifically this line: ``environ['REMOTE_ADDR'] = str(client)``

To overcome this issue, I propose using ``client.decode()`` instead of ``str(client)``:

```python
b''.decode()
# returns an empty string
str(b'')
# returns a string containing the three characters b''
```

This fixes issue #1755 for me.

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -158,7 +158,7 @@ def create(req, sock, client, server, cfg):
     if isinstance(client, string_types):
         environ['REMOTE_ADDR'] = client
     elif isinstance(client, binary_type):
-        environ['REMOTE_ADDR'] = str(client)
+        environ['REMOTE_ADDR'] = client.decode()
     else:
         environ['REMOTE_ADDR'] = client[0]
         environ['REMOTE_PORT'] = str(client[1])


gunicorn/http/wsgi.py
```diff
@@ -158,7 +158,7 @@ def create(req, sock, client, server, cfg):
     if isinstance(client, string_types):
         environ['REMOTE_ADDR'] = client
     elif isinstance(client, binary_type):
-        environ['REMOTE_ADDR'] = str(client)
+        environ['REMOTE_ADDR'] = client.decode()
```
berkerpeksag: I usually prefer specifying the encoding explicitly in a codebase that should support both Python 2 and Python 3 since the default encoding is different between the two versions, but in this case I don't think `client` can contain non-ASCII values in Python 2.

PR #1780, opened 2018-05-14
Fix access logging

yunstanford:
Today I observed that my service access log doesn't look correct, None of those identifiers/atoms has been replaced with proper value.

Looks like this commit (https://github.com/benoitc/gunicorn/commit/610596c9d93b3c9086becd6212ab4ba81d476dc4) in new version of gunicorn has broken the behavior.
And switching it back to `%` works fine as usual.

Any legacy code in python logging module cause this ?


yunstanford:
nvm, figured out.

PR #1782, opened 2018-05-16
Labeled the name of the file in the sample

W7PEA:
The sample as-is was not clear about what file the configuration was to go in. 
This sample is to go into nginx.conf, but nginx as of 20180516 has nginx.conf + conf.f/default.conf
This sample overrides all that into the single file so being specific is better.

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -17,7 +17,7 @@ An `example configuration`_ file for fast clients with Nginx_:
 
 .. literalinclude:: ../../examples/nginx.conf
    :language: nginx
-   :caption: nginx.conf
+   :caption: **nginx.conf**
 
 If you want to be able to handle streaming request/responses or other fancy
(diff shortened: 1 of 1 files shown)


docs/source/deploy.rst
```diff
@@ -15,6 +15,9 @@ You can use Hey_ to check if your proxy is behaving properly.
 
 An `example configuration`_ file for fast clients with Nginx_:
 
+*nginx.conf*
```
berkerpeksag: It would be better to use the `caption` option of `literalinclude`:

```rst
 .. literalinclude:: ../../examples/nginx.conf
     :language: nginx
     :caption: nginx.conf
```

W7PEA:
Was not aware that was an option... sounds good.

berkerpeksag:
Thanks! I've opened #1788 to cleanup `docs/source/deploy.rst` if you're looking for another issue to contribute to Gunicorn.

PR #1783, opened 2018-05-16
Pid and Logging changes

W7PEA:
PID didn't work at all on Ubuntu 16.04 with the provided sample, my commented out change does, also added logging change to conform to what is out of the box.

### DIFF
--- a/examples/nginx.conf
+++ b/examples/nginx.conf
@@ -2,13 +2,8 @@ worker_processes 1;
 
 user nobody nogroup;
 # 'user nobody nobody;' for systems with 'nobody' as a group instead
-error_log  /var/log/nginx/error.log warn;    # Ubuntu 16.04 + Nginx 1.14
-pid        /var/run/nginx.pid;               # Ubuntu 16.04 + Nginx 1.14
-
-# On some systems, these settings are preferred
-# pid /tmp/nginx.pid;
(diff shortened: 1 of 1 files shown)


tilgovi:
Thanks!

It's possible `/tmp` is not writable by the nginx owner process. That is my guess.

The `/var/run` and `/var/log` directories are part of the [Filesystem Hierarchy Standard](https://en.m.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). I think those are sensible defaults. Can you change your PR to replace the existing example with your settings, please?

tilgovi:
I would take the comments out. You don't need to specify Ubuntu and you don't need to say some systems prefer /tmp. The Filesystem Hierarchy Standard is very widespread. It's a reasonable assumption. Anyone who needs to do something else can figure that out. There is value in the example being simple and having it just work. I'd rather someone just copy and paste something that works and has reasonable defaults than wonder whether it will work if they don't use Ubuntu. Chances are it will :).

tilgovi:
My preference would be for:

```
error_log  /var/log/nginx/error.log;
pid        /var/run/nginx.pid;
```

tilgovi:
Closed in f0c54993caa3e96cc8d0c9ad40c12ba19244992d with the style change. Thank you!

PR #1787, opened 2018-05-18
fixing zero bytes case in atoms access log

yunstanford:
`getattr(resp, 'sent', None)` also means false if the value is 0.


.gitignore
```diff
@@ -5,6 +5,7 @@
 .coverage
 .pytest_cache
 .tox
+.cache
```
tilgovi: This is from an older version of pytest. The newer version is `.pytest_cache` and is already in this file. Pleas remove this when you can, thanks!

yunstanford:
@tilgovi ok, updated `.gitignore`

tilgovi:
Thanks!

PR #1791, opened 2018-05-21
Fix #1790: check for new directories in InotifyReloader

### LINKED ISSUE #1790: InotifyReloader misses new modules
When using `InotifyReloader`, modules that are loaded after the workers are initialized are not included in watched directories.

This doesn't play nicely with Django, which will dynamically load things as needed.

grantisu:
The general idea in this PR is to call `get_dirs()` every time `event_gen()` signals a timeout (~1 sec), so that we keep up with new modules that have been loaded dynamically.

The `ReloaderBase` stuff is a minor refactor that can be dropped if nobody likes it.


berkerpeksag:
I like that refactoring commit, it simplifies the code a bit, and doesn't break any existing `Reloader*` subclasses.

benoitc:
how are given these extra files though?

benoitc:
ah it's in #1846 . Then why not the only question i have is if it can't be a configuration flag only instead of another command arg. Thoughts?

tilgovi:
@grantisu if you can rebase today, we would love to merge this. If not, I will try to find time to do so.

gunicorn/reloader.py
```diff
@@ -68,48 +70,46 @@ def run(self):
...
 
-            for extra_file in extra_files:
-                self.add_extra_file(extra_file)
-
         def add_extra_file(self, filename):
             dirname = os.path.dirname(filename)
 
-            if dirname in self._dirs:
-                return
+            with self._extra_files_lock:
```
tilgovi: This lock was removed in #2134 because the reloader starts after the application has been loaded now.

gunicorn/reloader.py
```diff
@@ -68,48 +70,46 @@ def run(self):
...
-                os.path.dirname(COMPILED_EXT_RE.sub('py', module.__file__))
-                for module in tuple(sys.modules.values())
-                if getattr(module, '__file__', None)
-            ]
+            dirnames = [os.path.dirname(fname) for fname in self.get_files()]
+            return set(dirnames)
 
-            return set(fnames)
+        def refresh_dirs(self):
+            with self._extra_files_lock:
```
tilgovi: Here also we do not need the log anymore. Gunicorn will call `add_extra_file` during initialization and then start the reloader. Once starter, only the reloader thread is expected to add more files / directories, so there is no lock.

tilgovi:
I'm confused by the diff, because master no longer has a lock and it's not re-introduced here, but I still see in the GitHub UI. 😕

tilgovi:
Is this still needed after #2134?

grantisu:
> I'm confused by the diff, because master no longer has a lock and it's not re-introduced here, but I still see in the GitHub UI.

I rebased this to include 51a4afd7ef2c0941047f40d7ec79f6344fa47528, and dropped the locking.

> Is this still needed after #2134?

 #2134 improves things, but doesn't fix my use case: Django will dynamically load modules as requests come in, so the set of files to watch can grow significantly after startup.

tilgovi:
I'd like to take another look at this.

When does Django lazily load modules? Is there a setting that can be enabled to toggle this on/off? If so, we could just document that reloader works best with that switch set to eagerly load.

If this is still useful, should we resolve the conflicts? I think it's safe, even without a lock, since nothing should be modifying the reloader from outside once it's been started.

grantisu:
> When does Django lazily load modules? Is there a setting that can be enabled to toggle this on/off? If so, we could just document that reloader works best with that switch set to eagerly load.

Django lazily loads modules that define/handle URL resolution on a per-request basis, which is not easy to turn off: https://docs.djangoproject.com/en/3.0/topics/http/urls/#how-django-processes-a-request

The specific code base that I ran into this problem on also loads a bunch of other things lazily, but I'm not sure how much of that I can blame on Django.

gunicorn/reloader.py
```diff
@@ -66,49 +68,45 @@ def run(self):
...
-            ]
+            dirnames = [os.path.dirname(os.path.abspath(fname)) for fname in self.get_files()]
+            return set(dirnames)
 
-            return set(fnames)
+        def refresh_dirs(self):
+            new_dirs = self.get_dirs().difference(self._dirs)
+            self._dirs.update(new_dirs)
+            for new_dir in new_dirs:
+                if os.path.isdir(new_dir):
```
tilgovi, line 101: `get_dirs()` is already calling `dirname()` so I don't think we need this check.

gunicorn/reloader.py
```diff
@@ -66,49 +68,45 @@ def run(self):
...
-            self._dirs = self.get_dirs()
-
-            for dirname in self._dirs:
-                if os.path.isdir(dirname):
-                    self._watcher.add_watch(dirname, mask=self.event_mask)
+            self.refresh_dirs()
 
             for event in self._watcher.event_gen():
                 if event is None:
+                    self.refresh_dirs()
```
tilgovi, line 109: Why is this the right place to call this?

gunicorn/reloader.py
```diff
@@ -118,7 +116,7 @@ def run(self):
 else:
 
     class InotifyReloader(object):
-        def __init__(self, callback=None):
+        def __init__(self, extra_files=None, callback=None):
```
tilgovi, line 119: Good catch!

gunicorn/reloader.py
```diff
@@ -66,49 +68,45 @@ def run(self):
...
             self._dirs = set()
             self._watcher = Inotify()
 
-            for extra_file in extra_files:
-                self.add_extra_file(extra_file)
-
         def add_extra_file(self, filename):
             dirname = os.path.dirname(filename)
 
+            super(InotifyReloader, self).add_extra_file(filename)
```
tilgovi, line 85: Can we lift this to the top of the function, please? I always like calling `super` as the first thing unless there's a specific need to do things in a different order.

gunicorn/reloader.py
tilgovi: Maybe this is the key I'm missing. I don't think I understand how this fixes the issue with Django because I don't know when the watcher would generate an empty event and cause this refresh. When will the reloader know to refresh and what happens if the file that changes is not in the initial set of directories that Django imported from?

gunicorn/reloader.py
grantisu: `dirname()` doesn't check to see if the path actually exists; this is the same logic from f39b065d4b880de331884e46191c1d031f98025e

gunicorn/reloader.py
grantisu: By default, `Inotify()` will watch for one second at a time, so this line will end up refreshing the list of watched directories every second.

grantisu:
Is there anything else left to do for this to get merged?

benoitc:
Applied this fix in commit 35614fe3. InotifyReloader now refreshes watched directories on each event loop timeout to catch dynamically loaded modules. Thank you for the contribution!

PR #1796, opened 2018-05-26
1795 rename async module

### LINKED ISSUE #1795: async and await are now reserved words in python 37
Hi guys

I've been playing with Python 3.7 to create a guide to migrate my company apps and I found that currently we can't run gunicorn with python 3.7 since `async` is now a reserved word and there is a module named `gunicorn.workers.async` that raises a syntax error when imported.

I'm wondering if there is anyone already working on this issue? if not, should I just rename the module and open a PR? Or this kind of change needs to be discussed to avoid BC breaks in the projects?

Thanks

diegoholiveira:
Closes #1795


benoitc:
renaming it  `_async` would mean that module is private where some can import its objects to create other async workers.  I would rather rename it as `async_base` or `base_async` whatever.

diegoholiveira:
changes requested done ;)

benoitc:
@tilgovi @berkerpeksag thoughts? I am thinking we should create a pre-20 branch that include that change and the removal of python 2. Thoughts?

requirements_test.txt
```diff
@@ -1,3 +1,3 @@
 coverage>=4.0,<4.4  # TODO: https://github.com/benoitc/gunicorn/issues/1548
-pytest==3.0.5
-pytest-cov==2.4.0
+pytest
```
berkerpeksag: Please leave unrelated changes to another PR.

docs/source/news.rst
```diff
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+19.9.0 / 2018/05/26
+===================
+
+- the internal module `gunicorn.workers.async` was renamed to `gunicorn.workers.base_async`
```
berkerpeksag: the -> The

berkerpeksag: Style nit: Please use double backtics.

docs/source/news.rst
```diff
@@ -2,6 +2,14 @@
 Changelog
 =========
 
+19.9.0 / 2018/05/26
```
berkerpeksag: I would change 2018/05/26 with "not released".

.gitignore
```diff
@@ -15,3 +15,5 @@ examples/frameworks/pylonstest/pylonstest.egg-info/
 MANIFEST
 nohup.out
 setuptools-*
+.cache
```
berkerpeksag: .cache was renamed to .pytest_cache and we already added it to .gitignore.

.gitignore
```diff
@@ -15,3 +15,5 @@ examples/frameworks/pylonstest/pylonstest.egg-info/
...
+.eggs
```
berkerpeksag: This looks like a good candidate for having a global .gitignore file.

diegoholiveira:
@berkerpeksag done

gunicorn/workers/__init__.py
```diff
@@ -17,6 +17,6 @@
 }
 
 
-if sys.version_info >= (3, 3):
+if sys.version_info >= (3, 4):
     # gaiohttp worker can be used with Python 3.3+ only.
```
mozillazg: `3.3+` -> `3.4+`?

diegoholiveira: I did this change to match this: https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gaiohttp.py#L10

mozillazg: I mean the comment `# gaiohttp worker can be used with Python 3.3+ only.` should be updated to `# gaiohttp worker can be used with Python 3.4+ only.`

berkerpeksag:
Can we please use the "squash and merge" option when there are intermediate commits in a PR? Commits like https://github.com/benoitc/gunicorn/commit/07dc7167003d231364991416080367cf103a2c26  and https://github.com/benoitc/gunicorn/commit/66ec02129150e434d82820dcefafbf8229cbf92b shouldn't be merged into the master branch.

We don't have feature branches and develop features the way the Linux kernel team does (long lived branches, no intermediate commits etc.), so I think we can even disable the "create a merge commit" option.

benoitc:
@berkerpeksag well we should probably  have long lived branch. So we can break from now the code and work on next major release and start new iterations from it :) For example, I wouldn't have merged it before bumping our own version. That change will break anyone using a custom worker based on the async base. So that's a good call for such branches :)

PR #1799, opened 2018-06-03
Upgrade pytest and pytest-cov plugin

diegoholiveira:


diegoholiveira:
@berkerpeksag I open another PR to upgrade pytest since it's something not related to #1796

diegoholiveira:
Also, this should be merged only when I drop python 2.6 support since pytest does not support this version.

diegoholiveira:
@berkerpeksag done

PR #1812, opened 2018-06-16
Minor reloader optimizations

JustinTArthur:
I noticed that `--reload` was taking up a decent amount of CPU time using the classic polling reloader. Precompiling the .pyc/.pyo extensions regex like we do in other modules helped. Linux users with inotify are unlikely to notice a difference.

If we wanted to optimize further, the `sub` method could be saved in a module var, but I think readability starts to decline.

### DIFF
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -10,6 +10,8 @@ import sys
 import time
 import threading
 
+COMPILED_EXT_RE = re.compile(r'py[co]$')
+
 
 class Reloader(threading.Thread):
     def __init__(self, extra_files=None, interval=1, callback=None):
@@ -26,8 +28,8 @@ class Reloader(threading.Thread):
 
     def get_files(self):
         fnames = [
-            re.sub('py[co]$', 'py', module.__file__)
-            for module in list(sys.modules.values())
+            COMPILED_EXT_RE.sub('py', module.__file__)
+            for module in tuple(sys.modules.values())
             if getattr(module, '__file__', None)
         ]
 
@@ -92,8 +94,8 @@ if has_inotify:
 
         def get_dirs(self):
             fnames = [
-                os.path.dirname(re.sub('py[co]$', 'py', module.__file__))
-                for module in list(sys.modules.values())
+                os.path.dirname(COMPILED_EXT_RE.sub('py', module.__file__))
+                for module in tuple(sys.modules.values())
                 if hasattr(module, '__file__')
             ]
 


gunicorn/reloader.py
```diff
@@ -26,7 +28,7 @@ def add_extra_file(self, filename):
 
     def get_files(self):
         fnames = [
-            re.sub('py[co]$', 'py', module.__file__)
+            COMPILED_EXT_RE.sub('py', module.__file__)
```
berkerpeksag: The re module has internal caching, so I'm surprised moving the pattern to `re.compile()` had any effect. Out of curiosity, how did you measure the performance gain?

gunicorn/reloader.py
```diff
@@ -26,7 +28,7 @@ def add_extra_file(self, filename):
...
             for module in list(sys.modules.values())
```
berkerpeksag: If we want to optimize this further, we can remove `list()` calls around `sys.modules.values()` as well. `sys.modules.values()` returns a plain `list` in Python 2 and `dict_values` view in Python 3 (even better!) so no need create another list object.

JustinTArthur: I agree. Looked unnecessary regardless of speed. Will add it in.

gunicorn/reloader.py
JustinTArthur: It's true. I looked at cpython's `re` source to be sure. Maybe it's just the extra Python call stack clutter leading up to cache hit. I first noticed the issue after instrumenting workers with [profiling](https://github.com/what-studio/profiling) ([like this](https://github.com/what-studio/profiling/issues/31#issuecomment-397774043)). Isolated results with `timeit` and IPython's `%timeit`.

```python
import re
def old_way():
    return re.sub('py[co]$', 'py', 'stuff_and_things.pyc')

precompiled_pattern = re.compile(r'py[co]$')
def new_way():
    return precompiled_pattern.sub('py', 'stuff_and_things.pyc')
```
```
%timeit a=new_way()
The slowest run took 8.21 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 584 ns per loop

%timeit a=old_way()
The slowest run took 6.37 times longer than the fastest. This could mean that an intermediate result is being cached.
1000000 loops, best of 3: 1.17 µs per loop
```

JustinTArthur:
Rebased with a tuple instead of a list for the iterable copies for now.

berkerpeksag:
Thanks!

Issue #1429, opened 2017-01-12
Improve documentation on running gunicorn from Python with custom applications


duffsterlp:
I'm running gunicorn using the falcon web framework and would like to run gunicorn from the Python interpreter directly, similar to this issue here:
https://github.com/benoitc/gunicorn/issues/91

I now know how to accomplish this from a combination of these links as well as a conversation in the IRC room.
http://docs.gunicorn.org/en/stable/custom.html
https://gist.github.com/quamilek/4fd1f246feb6149dd1dd/

The latter link really shed light that the "Custom Application" section of the documentation was the appropriate place to connect Gunicorn with existing WSGI frameworks. The Gunicorn documentation alone did not provide ample indication of this, in my opinion. 

I'm looking for the following improvements in the Custom Application page:

1. Something that explicitly indicates how to run Gunicorn from Python using existing WSGI frameworks. The example only indicates how to connect a custom application that implements the WSGI specification itself. 
2. Some more documentation on the Gunicorn app base class. The "Custom Application" section provides an example of how to use this base class, but there should be an API that indicates how to use the app base class. Without this API, users are forced to look at the source code to infer how to use it. For example, there is nothing in the documentation that says that the load_config() and load() functions need to be overridden.

benoitc:
maybe @matrixise can help there since he's supposed to maintain that feature?

berkerpeksag:
For 1), perhaps we can provide a factory function that returns an application class?

We can use 2) as an opportunity to refactor the API (if needed) and document it.

benoitc:
@duffsterlp out of curiosity, what's the reaon to not simply use the gunicorn command line?

aldanor:
> @duffsterlp out of curiosity, what's the reaon to not simply use the gunicorn command line?

For example, you have a CLI which you need to pass some parameters to in order to start the server.

JakeTheDev:
Bringing this back from the dead, I've been on a bit of a witch hunt, and I think this would solve my issue. 

Context: We're a microservice type company, and we have virtual machines that act as containers, but container management software can't be adopted in full at $work yet. To cut a huge amount of complexity from $configmanager and $deployer, we're taking the PEX deployable route for python web services, and these apps generally exist as Flask services around here, though Django would be a nice to have.

[...]

PR #1817, opened 2018-06-26
Docs: serving WSGI app modules from Gunicorn

jakethedev:
I had a use case where I needed to be able to standardize how our servers are running, but the application code could vary wildly within the scope of existing as a WSGI app ala flask or django. This had to happen within a PEX context, and I was able to get it working nicely and reliably after digging through the Gunicorn source to find this approach. Now I've got a happy little scaffold for wsgi apps that is standard for our company that works well PEX files by including Gunicorn in the build, but it was a real pain to figure this out, and it would've save me a day or two of researching in circles if this was mentioned somewhere

Hope this matches appropriate formatting, I checked CONTRIBUTING and it didn't mention much about docs and Github isn't rendering the preview quite right

### DIFF
--- a/docs/source/custom.rst
+++ b/docs/source/custom.rst
@@ -14,3 +14,31 @@ a custom Application:
 
 .. literalinclude:: ../../examples/standalone_app.py
     :lines: 11-60
+
+Direct Usage of Existing WSGI Apps
+----------------------------------
+
+If necessary, you can run Gunicorn straight from python, allowing you to 
+specify a WSGI-compatible application at runtime. This might be useful for 
+rolling deploys or in the case of using PEX files to deploy your application,
+as the app and Gunicorn can be bundled in the same PEX file. At the time of
+writing, you can import and instantiate a WSGIApplication like so:
+
+.. code-block:: python
+    from gunicorn.app.wsgiapp import WSGIApplication
+    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
+
+This can be used to run WSGI-compatible app instances such as those produced
+by Flask or Django. Assuming that the code above is in a file called `run.py`,
+your application's WSGI instance is called `app`, and it's exported by a package
+called `exampleapi`, then you should be able to fire up your server with a
+command similar to '``python run.py exampleapi:app``'. 
+
+This command will work with any Gunicorn CLI parameters or a config file - just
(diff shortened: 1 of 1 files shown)


tilgovi:
Thanks so much, @JakeTheDev. I'm fine for this to be in the documentation.

tilgovi:
I'm not sure I see why you're doing this, though. `gunicorn` is that entry point. Why would you need to do `python run.py` vs `gunicorn`?

JakeTheDev:
Sweet!

Initially for flexibility, and the assumption that we'd have system daemons coming out of the same build process as the Gunicorn apps. That pushed us to move that complexity into how PEX was invoked, which led to a pretty big road block. We found that PEX requires either a single pythonic entry point (like app:serve), or a script. It refuses to take two words as a command, even properly quoted off as a single command - it just can't handle something like 'pex -c "gunicorn flaskexample:app" '.

So, building a PEX around a wsgi app requires a runnable script that does all the work, and this PR is the result of climbing that mountain and arriving at "bundle gunicorn with the app and bolt them together with a script" :sweat_smile: It seems that starting Gunicorn like this is a bit like push starting a car. It's rock solid, if a bit unorthodox, so I can dig it if you think this should go undocumented as it relies on internals, but it seemed worth sending your way in case others have the same issues getting PEX going

tilgovi:
Thanks for the explanation! The `gunicorn.app.wsgiapp` module is a runnable script that does this!

JakeTheDev:
Cleaned it up and simplified the docs, let me know if I can clean it up further or if I missed something obvious. Thanks for clearing up the usage of wsgiapp, I'm still trying to figure out how I saw that code and didn't see the __main__ at the bottom haha

JakeTheDev:
PS I think this might resolve #1429 :smile:

jakethedev:
Just finished working with the PEX maintainers to update some docs on PEX usage, and remembered that this was still open - cleaned up a bit of wording and made the PEX usage note better. Lmk if there's anything else I should tweak on this pre-merge :smile:

benoitc:
@tilgovi since you looked it at first, still OK to merge it ? Changes look good for me.

tilgovi:
Looks good to me.

berkerpeksag:
Thanks!

PR #1819, opened 2018-06-27
Collect worker idle time stats and export in Prometheus format

### LINKED ISSUE #1818: Monitoring sync workers idle/busy time
## Background

My current project is migrating to microservice architecture. But unfortunately all code is written in synchronous style. To make refactoring process smoother we decided to migrate code to microservices and then try to rewrite it in async style. But now we should be prepared to correctly scale each service workers, because inter-service communication will generate many requests and many of them are long.

We want to find a solution to understand is current amount of workers for some service is enough. Or there is always requests waiting for a worker to free.

## Solution 1

One solution is to monitor system TCP backlog. But for some reason it's always empty on systems where I tested it. Also there is downside of this method. Precision is limited by frequency of backlog queries. So we can miss backlog length peaks between two queries.

## Solution 2

Otherwise we can add metric collector to gunicorn and export it in some format. For example, Prometheus:

```
# TYPE idle_time_seconds_sum summary
idle_time_seconds_sum{pid="8013"} 51.135032176971436
```

Then we can query Prometheus with `avg (irate(idle_time_seconds_sum[5m]))` to see floating-window average busyness.

This metric will be pretty precise because it is derivative of summarised idle time and nothing will be lost in between metric queries.

nailgun:
Closes #1818 

Ready to fix design issues

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -52,6 +52,7 @@ Denis Bilenko <denis.bilenko@gmail.com>
 Diego Oliveira <contact@diegoholiveira.com>
 Dima Barsky <github@kappa.ac93.org>
 Djoume Salvetti <djoume@freshbooks.com>
+Dmitry Bashkatov <dbashkatov@gmail.com>
 Dmitry Medvinsky <me@dmedvinsky.name>
 Ed Morley <edmorley@users.noreply.github.com>
 Eric Florenzano <floguy@gmail.com>
--- a/gunicorn/app/base.py
+++ b/gunicorn/app/base.py
@@ -68,7 +68,7 @@ class BaseApplication(object):
     def wsgi(self):
         if self.callable is None:
             self.callable = self.load()
-            self.callable = self.add_prometheus_endpoint(self.callable)
+            self.add_prometheus_endpoint()
         return self.callable
 
     def run(self):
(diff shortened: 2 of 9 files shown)


nailgun:
We are now testing this in production and it looks good. This is idle time of three gunicorn instances. Each line is average idle time of all workers of an instance.

image

tilgovi:
This is excellent work, thank you. I don't have time to review it until the weekend, but I'm looking forward to it!

benoitc:
cc @tilgovi

tilgovi:
I am so sorry I did not ever review this and I still very much appreciate the PR.

I agree that the approach is intrusive. Monitoring tools rise and fall in fashion and we have seen different requests over the years for integrations. We should discuss an approach in #1564 that might make this possible through plugins.

benoitc:
I agree.I am closing this PR then. Let's continue the discussion on #1564 as suggested by @tilgovi

nailgun:
Reimplemented this using [middleware](https://github.com/smpio/python-utils/blob/64c36c1a65a65b8da49fde2be98a3428504a688a/utils/wsgi/middeware/idle_counter.py)
and metrics [view](https://github.com/smpio/python-utils/blob/64c36c1a65a65b8da49fde2be98a3428504a688a/utils/django/views.py#L77-L77)

PR #1821, opened 2018-06-28
Display NameError on import

c23009u:
If user code raises a NameError when gunicorn is importing the application, the error will be swallowed.  Gunicorn is catching NameError to handle the case where the entrypoint to the user app is not valid.  However this also swallows other NameErrors in the user code, such as accessing an undefined variable.  This PR makes the correct traceback available.

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -20,7 +20,6 @@ import traceback
 import inspect
 import errno
 import warnings
-import logging
 import re
 
 from gunicorn import _compat
@@ -357,13 +356,12 @@ def import_app(module):
 
     mod = sys.modules[module]
 
-    is_debug = logging.root.level == logging.DEBUG
     try:
         app = eval(obj, vars(mod))
     except NameError:
-        if is_debug:
-            traceback.print_exception(*sys.exc_info())
-        raise AppImportError("Failed to find application object %r in %r" % (obj, module))
+        raise AppImportError("Failed to find application object '{0}' in '{1}':\n{2}".format(
+            obj, module, traceback.format_exc()
+        ))
 
     if app is None:
         raise AppImportError("Failed to find application object: %r" % obj)


tilgovi:
Thanks for the pull request! @berkerpeksag do you think it's okay to log tracebacks for this even when the app is not in debug? It seems alright to me, but would love a second opinion.

c23009u:
I had the same thought, if it's any help my reasoning was that it's not printing anything to stdout.  Previously it would print the traceback to stdout if debug was enabled, and always raise an exception regardless of debug mode.  This simply adds the traceback to the end of the exception message and altogether removes the printing.  It's the same exception class being raised.  I can't imagine having the additional information in the exception message would break anyone's code.

gunicorn/util.py
```diff
@@ -357,13 +356,12 @@ def import_app(module):
...
     mod = sys.modules[module]
 
-    is_debug = logging.root.level == logging.DEBUG
     try:
         app = eval(obj, vars(mod))
     except NameError:
-        if is_debug:
-            traceback.print_exception(*sys.exc_info())
-        raise AppImportError("Failed to find application object %r in %r" % (obj, module))
+        raise AppImportError("Failed to find application object '{0}' in '{1}':\n{2}".format(
```
berkerpeksag: You can use `{0!r}` instead of `'{0}'`. But I agree with @tilgovi's comment at https://github.com/benoitc/gunicorn/pull/1821#issuecomment-402309585 that we don't need to change the exception message.

benoitc:
About the log level, can this error really happen in prod ? I would keep tracebacks for debug situations imo.  Why making it available at the error level? Also we should be cautious with tracebacks as they can see unwanted (sensitive) infos.

benoitc:
@berkerpeksag is there any better way to make sure the exception will be raised at debug level? Maybe just losing the exception there would be enough ?

PR #1826, opened 2018-07-03
Detailed error on gevent import failure

jesper:
Display the ImportError message to get more details as to why gevent failed to be imported.

This comes from a real-world scenario where I was debugging a system where gevent *was* installed, but for some reason greenlet was not. The error given by gunicorn was simply ~"gevent not installed" which was quite confusing. By introducing this change the root cause of the error will be brought forward and the issue made obvious.

### DIFF
--- a/gunicorn/workers/ggevent.py
+++ b/gunicorn/workers/ggevent.py
@@ -18,8 +18,8 @@ if sys.platform == "darwin":
 
 try:
     import gevent
-except ImportError:
-    raise RuntimeError("You need gevent installed to use this worker.")
+except ImportError, e:
+    raise RuntimeError("You need gevent installed to use this worker: %s" % e.message)
 from gevent.pool import Pool
 from gevent.server import StreamServer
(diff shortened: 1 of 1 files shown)


benoitc:
@jesper bump?

sirkonst:
Why debug log? It is a critical error for start application.

benoitc:
well such errors shouldn't be in production logs imo. That should have been checked before deployement.

sirkonst:
Right, but anything can happen :-) That MR just add more details for error message.

But on the other hand this change need to be apply to all workers not only gevent.

@benoitc - What you prefer print to debug log? Just message from `e.message` in this case?

benoitc:
@sirkonst maybe logging the full exception , but indeed it's probably enough.

tilgovi:
I agree with this change and agree we should add it for eventlet and aiohttp, or any workers that have extras requirements.

benoitc:
not applicable today. Thanks for the patch anyway

PR #1827, opened 2018-07-03
Fix information about Python supported versions

diegoholiveira:


tilgovi:
Thank you. I don't think we dropped 3.3 support with this release, though.

tilgovi:
Although, we will shortly.

tilgovi:
Maybe that's okay. 3.3 is EOL, so I'm okay with documenting 3.4+, especially because we are about to drop 3.3.

tilgovi:
Well, let's make it accurate, though. We can change it again when we require 3.4+.

diegoholiveira:
@tilgovi I add 3.3 to the supported versions again.

tilgovi:
I should have said 3.2. I don't think we dropped 3.2, yet, either :)

tilgovi:
Thank you, @diegoholiveira! We will want to merge something like this very soon :)

tilgovi:
I'll add 3.7 to the meta for now.

Issue #1566, opened 2017-08-06
Evaluate support for preload and monkey-patching


tilgovi:
Extending discussion from #927, it would be good if, for R20, we can take a firm stance on whether gevent/eventlet are supported with the preload option. This evaluation will probably depend on our approach to multiprocessing under Windows, as well as the state of gevent/eventlet and how we plan to maintain worker classes going forward.

komuw:
I think the new gc.freeze api would be of use, more details in the cpython pull request and issue ticket;
1. https://github.com/python/cpython/pull/3705
2. https://bugs.python.org/issue31558

hingston:
As of gevent 1.3a1 you get the following `RuntimeWarning`

`/env/lib/python3.6/site-packages/gunicorn/workers/ggevent.py:65: RuntimeWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. Please monkey-patch earlier.`

jamadden:
The warning was added as a result of https://github.com/gevent/gevent/issues/1016. Basically, importing `requests` and then monkey-patching was essentially guaranteed to lead to a `RecursionError` (which looks like #1559).

benoitc:
@tilgovi how do you think we can progress on that topic? Would be good to draft some roadmap for it.

 In the mean time I would postpone it for the next release if you're OK?

tilgovi:
I'm okay to postpone. @jamadden do you think we should try importing gevent earlier? What has been your experience? If I remember correctly, I've seen code you shared from a project where you monkey-patch before loading Gunicorn as a custom application.

benoitc:
postponing to 20.1 let's try to find a solution for it :)

benoitc:
So back on this topic since someone reported on irc he hit the error. The problem I see in monkey patching during preload is that it would make the socket unblocked in the arbiter, which is not really wanted. We should make sure to use the unpatched module/function in the arbiter.

tilgovi:
Does the arbiter perform any blocking calls on the socket? If not, do we care if the socket is patched in the arbiter? Mostly, the arbiter just passes the socket to workers.

benoitc:
The arbiter itself rely on a blocking behaviour for select and signaling. Also it's safer to bind the socket in blocking mode. This is why it's discouraged to patch outside of a worker. Also in term of design i am thinking it's better to have the arbiter isolated from the application logic. 

[...]

Issue #1816, opened 2018-06-25
gevent monkey-patch earlier


buchmoyerm:
Using gunicorn 19.8.1 i get the monkey-patch warning with only a single worker

I am using a pretty basic flask app.py

My command I am running is `gunicorn -b 0.0.0.0:9000 -k gevent -w 1 app:app` and the warning is being printed before my app.py event starts running

```
[2018-06-25 13:58:03 -0400] [149803] [INFO] Starting gunicorn 19.8.1
[2018-06-25 13:58:03 -0400] [149803] [INFO] Listening at: http://0.0.0.0:9000 (149803)
[2018-06-25 13:58:03 -0400] [149803] [INFO] Using worker: gevent
[2018-06-25 13:58:03 -0400] [149815] [INFO] Booting worker with pid: 149815
/(Redacted local path)/venv/lib/python3.6/site-packages/gunicorn/workers/ggevent.py:65: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016
  monkey.patch_all(subprocess=True)
```

I looked and found 

https://github.com/benoitc/gunicorn/blob/bd833e0009b8b07d13117f71534eb28e8dc24c5d/gunicorn/workers/ggevent.py#L304 (file that does the patch_all)

and inside base_async.py

https://github.com/benoitc/gunicorn/blob/bd833e0009b8b07d13117f71534eb28e8dc24c5d/gunicorn/workers/base_async.py#L92 (file that imports ssl)

So I am not sure how this is supposed to work, but it doesn't seem like it can.

jamadden:
I would expect that gevent 1.3.4 would silence the error here.

buchmoyerm:
That seems to have fixed the issue. Looks like version 1.1rc3.3 just crashed the app (which is what I had on prod). And when testing locally I had version 1.3.1 installed which didn't crash the app but printed the message. 1.3.4 did not print the message at all. Thanks.

PR #1828, opened 2018-07-04
gevent monkey patch before any other imports

diegoholiveira:
I'm already using gunicorn with python 3.7 successfully in some of my apps in production. \o/

In those productions servers, since I'm running py3.7, I was forced to use gevent 1.3.4, and after the 1.3.x release, they introduce this warning:

```
/usr/local/lib/python3.7/site-packages/gunicorn/workers/ggevent.py:65: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016.
monkey.patch_all(subprocess=True)
```

Other issues already point this error also, like #1566. I don't know if this fix is naive and can cause other issues, but I hope we can be at least a first draft of a solution.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -1168,11 +1168,11 @@ libraries may be installed using setuptools' ``extra_require`` feature.
 A string referring to one of the following bundled classes:
 
 * ``sync``
-* ``eventlet`` - Requires eventlet >= 0.9.7 (or install it via 
+* ``eventlet`` - Requires eventlet >= 0.9.7 (or install it via
   ``pip install gunicorn[eventlet]``)
-* ``gevent``   - Requires gevent >= 0.13 (or install it via 
+* ``gevent``   - Requires gevent >= 1.0 (or install it via
   ``pip install gunicorn[gevent]``)
-* ``tornado``  - Requires tornado >= 0.2 (or install it via 
+* ``tornado``  - Requires tornado >= 0.2 (or install it via
   ``pip install gunicorn[tornado]``)
 * ``gthread``  - Python 2 requires the futures package to be installed
   (or install it via ``pip install gunicorn[gthread]``)
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -611,11 +611,11 @@ class WorkerClass(Setting):
         A string referring to one of the following bundled classes:
 
         * ``sync``
-        * ``eventlet`` - Requires eventlet >= 0.9.7 (or install it via 
+        * ``eventlet`` - Requires eventlet >= 0.9.7 (or install it via
           ``pip install gunicorn[eventlet]``)
-        * ``gevent``   - Requires gevent >= 0.13 (or install it via 
+        * ``gevent``   - Requires gevent >= 1.0 (or install it via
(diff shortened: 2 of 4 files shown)


jamadden:
Is there more to the error message? gevent 1.3.4 only issues that warning when it detects that there is likely to be a problem with SSL support (such as a RecursionError), and it is supposed to print the causes of that problem. I would expect to see "Modules that had direct imports (NOT patched):" or "Subclasses (NOT patched):"

In https://github.com/benoitc/gunicorn/issues/1816 a different user reported that 1.3.4 silenced the warning, which is what I would generally expect; and if not silenced I would expect more details.

diegoholiveira:
@jamadden here is the full error message:

```
/usr/local/lib/python3.7/site-packages/gunicorn/workers/ggevent.py:65: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['newrelic.packages.requests.packages.urllib3.util (/usr/local/lib/python3.7/site-packages/newrelic/packages/requests/packages/urllib3/util/__init__.py)'].
```

I notice know that the warning is caused by the newrelic command that I use to run my apps:

```
newrelic-admin run-program gunicorn \
  --bind 0.0.0.0:9000 \
  -k gevent \
  --worker-connections 1001 \
  -w 5 \
  app:app
```

When I run the gunicorn directly, those warnings go away. With my patch  I got only one warning in the startup instead of one warning per worker + one warning in the startup.

benoitc:
doing this would also patch the arbiter, wich is not possible since the arbiter needs to block. Is the point about removing the warning?

jamadden:
> doing this would also patch the arbiter, wich is not possible since the arbiter needs to block

I'll note that we successfully run the entire gunicorn stack, master/arbiter and workers, fully monkey-patched by gevent. The arbiter still blocks and still supervises correctly.

jamadden:
> Since which versions of gevent are we talking about?

We've been running like this since September 2012, so 1.0b4?

tilgovi:
I very much support early monkey patching if it's safe to do. I worry a little bit about anyone who might be starting background threads in server hooks.

tilgovi:
> I worry a little bit about anyone who might be starting background threads in server hooks.

This could be addressed through very good documentation with examples of how to create a real thread for CPU-bound work even when threading is already patched.

jamadden:
> I worry a little bit about anyone who might be starting background threads in server hooks.

Aside from CPU or blocking issues, there's another important difference, which is that (real, native) threads do not survive a fork, but greenlets (and thus greenlet-backed monkey-patched threads) *do*.

benoitc:
> I very much support early monkey patching if it's safe to do. 

 I think we should be careful there sinc eit willbreak the isolation between the workers and the arbiter. What advantage would it brings to patch early outside the worker process?

PR #1831, opened 2018-07-17
Clear tornado ioloop before os.fork

yunstanford:
an `IOLoop` cannot survive a fork or be shared across processes in any way. When multiple processes are being used, each process should create its own `IOLoop`. We should probably clear current `IOLoop` if exists before `os.fork`.


tilgovi:
Thank you for the PR.

The documentation for that method says "Intended primarily for use by test frameworks in between tests." I'm a little concerned that this could mask bugs or create confusion. It also doesn't appear to do anything in the built-in Twisted implementation and [Python has an open bug](https://bugs.python.org/issue21998) for handling `fork()` in asyncio.

I'm not sure what to do.

tilgovi:
I think this is probably safe and maybe good. I'll wait to see if anyone else has thoughts before I merge it, though.

yunstanford:
I can add a comment there, but  yeah, i agree, it probably won't fix all cases,

gunicorn/workers/gtornado.py
```diff
@@ -70,6 +70,12 @@ def heartbeat(self):
                     self.ioloop.stop()
 
     def init_process(self):
+        """\
```
berkerpeksag: Style nit: I'd use `# comment`, not a docstring :)

tilgovi:
Looks good to me. Thank you!

Issue #1832, opened 2018-07-17
Fix linting for Pylint 2.0


tilgovi:
It looks like Pylint 2.0 was released and now our lint fails in CI.

PR #1833, opened 2018-07-18
Fix pylint

yunstanford:
https://github.com/benoitc/gunicorn/issues/1832
fix some rules, also ignore some rules as those might not apply well in our case.


yunstanford:
Looks like syntax error for python2.6 build, are we still supporting python 2.6 ? @tilgovi

gunicorn/arbiter.py
```diff
@@ -230,8 +230,6 @@ def run(self):
             self.halt()
         except HaltServer as inst:
             self.halt(reason=inst.reason, exit_status=inst.exit_status)
-        except SystemExit:
```
berkerpeksag: While I think the change looks fine, leaving it would make exception handling clearer for the reader. I'd add try-except-raise to .pylintrc instead.

gunicorn/arbiter.py
```diff
@@ -582,8 +580,6 @@ def spawn_worker(self):
             self.cfg.post_fork(self, worker)
             worker.init_process()
             sys.exit(0)
-        except SystemExit:
```
berkerpeksag: This change is incorrect, because there is a bare `except` clause in line 589. We need to specifically catch and reraise `SystemExit`.

gunicorn/glogging.py
```diff
@@ -318,18 +318,18 @@ def atoms(self, resp, req, environ, request_time):
         if hasattr(req_headers, "items"):
             req_headers = req_headers.items()
 
-        atoms.update(dict([("{%s}i" % k.lower(), v) for k, v in req_headers]))
+        atoms.update({"{%s}i" % k.lower(): v for k, v in req_headers})
```
berkerpeksag: We still need to support Python 2.6, I'd add consider-using-dict-comprehension to .pylintrc with a TODO comment to remind us that we can use dict comprehensions once we drop Python 2.6 support.

gunicorn/workers/base_async.py
```diff
@@ -121,8 +121,6 @@ def handle_request(self, listener_name, req, sock, addr):
                     respiter.close()
             if resp.should_close():
                 raise StopIteration()
-        except StopIteration:
```
berkerpeksag: I'm not sure about this. StopIteration is a subclass of Exception, but I don't know this code well, so I'll just defer to @benoitc and @tilgovi.

gunicorn/glogging.py
yunstanford: sounds good

gunicorn/arbiter.py
yunstanford: sounds good

gunicorn/arbiter.py
yunstanford: got it, should be fine now as we're going to ignore `try-except-raise`.

gunicorn/workers/base_async.py
yunstanford: i guess we could add it back as we're going to ignore `try-except-raise`

berkerpeksag:
Thanks!

PR #1844, opened 2018-08-01
Make option to add new extra files with extension.

firekim2:
Make option to add new extra files with extension.

For example gunicorn ... --reload-extra-extension html
then, it watch and reload all *.html files.


firekim2:
With small change, I can get directory for adding to list or regex type also.

firekim2:
@berkerpeksag agree with that. I'll change it and commit it soon.

firekim2:
I fixed code as you said. It works like these : 
--reload-extra-file path/to/somewhere/* : using * 
--reload-extra-file path/to/somewhere/a.html path/to/another/b.html : spacing

benoitc:
what is the usage for it?

firekim2:
This is for adding reload files simply.
If I try to adding whole templates folder for reload now, i should adding file by file like below:
--reload-extra-file path-to/file1 --reload-extra-file path-to/file2 ....
Buy with this change i can add directory or specific files in direcotry at once(for example: path-to/*.html)

firekim2:
This is for adding reload files simply.
If I try to adding whole templates folder for reload now, i should adding file by file like below:
--reload-extra-file path-to/file1 --reload-extra-file path-to/file2 ....
But with this change i can add directory or specific files in direcotry at once(for example: path-to/*.html)

tilgovi:
Something seems wrong... when I look at the changes now in the GitHub interface I only see changes to the config module (and two .DS_Store files added).

firekim2:
Got it I'll reupload request. Thanks!

benoitc:
no activity since awhile. closing feel free to repost an updated PR as requested if needed.

PR #1869, opened 2018-08-31
Fixed jQuery not using https

Q-back:
Page stopped working on new browsers 
https://developer.mozilla.org/en-US/docs/Web/Security/Mixed_content

output from FF console:
Blocked loading mixed active content “http://ajax.googleapis.com/ajax/libs/jquery/1.7.2/jquery.min.js”
gunicorn.org
Loading failed for the <script> with source “http://ajax.googleapis.com/ajax/libs/jquery/1.7.2/jquery.min.js”. gunicorn.org:182
 ReferenceError: $ is not defined

### DIFF
--- a/docs/site/index.html
+++ b/docs/site/index.html
@@ -179,7 +179,7 @@
     </div>
   </div>
 
-  <script src="http://ajax.googleapis.com/ajax/libs/jquery/1.7.2/jquery.min.js"></script>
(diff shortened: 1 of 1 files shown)


tilgovi:
Thanks!

Q-back:
I'm glad I can help :D

PR #1873, opened 2018-09-05
[closes #1863] Docs: Add /dev/shm as a convenient alternative to making your own tmpfs mount in fchmod FAQ

### LINKED ISSUE #1863: worker_tmp_dir setting should use /dev/shm tmpfs by default
That or http://docs.gunicorn.org/en/stable/faq.html#blocking-os-fchmod should just recommend setting `worker_tmp_dir=/dev/shm`

Perhaps there should be a Cloud/AWS section in docs collecting all of these perf caveats?

ciarancourtney:
FYI I got a very minor perf boost on superficial local testing on ubuntu bionic using `wrk`, but notably I was getting 2 or 3 read errors when not using `/dev/shm`


benoitc:
thanks!

PR #1876, opened 2018-09-10
Remove unnecessary call to dict keys method

BubuYo:
`keys()` is a dynamic view under Python 3 and these dictionaries are not modified concurrently, so we don't need call the dict keys method anymore. And using `len(dict)` is more pythonista to get dict count by the way

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -539,7 +539,7 @@ class Arbiter(object):
         Maintain the number of workers by spawning or killing
         as required.
         """
-        if len(self.WORKERS.keys()) < self.num_workers:
+        if len(self.WORKERS) < self.num_workers:
             self.spawn_workers()
 
         workers = self.WORKERS.items()
@@ -610,7 +610,7 @@ class Arbiter(object):
         of the master process.
         """
 
-        for _ in range(self.num_workers - len(self.WORKERS.keys())):
+        for _ in range(self.num_workers - len(self.WORKERS)):
             self.spawn_worker()
             time.sleep(0.1 * random.random())
 


tilgovi:
You're correct. There's another instance on line 544. Would you also change that one, please?

BubuYo:
Sure, it's done now

benoitc:
But what about atomicity? Counting on keys was ensured to be atomic. Workers can be removed asynchronously from that dict. At least this should be checked.

benoitc:
to be more clear , using `.keys()` is about making sure to not get a runtime error like this `RuntimeError: dictionary changed size during iteration`

tilgovi:
This may have been true once, but I don't think it is today. All signals are handled in the main context by deferring handlers through the `SIG_QUEUE`. I would double check that we don't make calls inside any loops that might change the workers dict, but I think there may be no concurrency.

benoitc:
i think this can still happen there:
https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L568

but will do a review tomorrow

tilgovi:
Oh, also: both of these sites are using `len()`, so I think there's nothing to worry about.

benoitc:
@tilgovi mm mI don't follow, what do you mean by `these sites are using len()`?

BubuYo:
@benoitc Sounds like` len(deepcopy(self.workers))` may fix your concern,but is it a real problem?

benoitc:
mmmright now the python 3 doc state that dict.keys() is using a dynamic
view (that silly)

> keys()
>
> Return a new view of the dictionary’s keys. See the documentation of view
objects.

I'm not sure that len(self.WORKERS) is now atomic, it used to not be the
case. Seems it is as it looks at one value:
https://github.com/python/cpython/blob/e42b705188271da108de42b55d9344642170aa2b/Objects/dictobject.c#L2004-L2008

Anyway since keys() is not using a dynamic view there is indeed no reason
to use it. In any case please make the comment message more explicit about
the changes and reflect the discussion so we can relate on it later.

BubuYo:
@benoitc  keys() is using a dynamic view, right?why you say

> Anyway since keys() is not using a dynamic view there is indeed no reason
to use it.

tilgovi:
Let's merge this :).

@BubuYo could you simplify the commit message to something short like `Remove unnecessary call to dict keys method` and then leave a note in the full commit message explaining that `keys()` is a dynamic view under Python 3 and these dictionaries are not modified concurrently anyway?

Thank you!

BubuYo:
@tilgovi Sure

tilgovi:
Thank you!

Issue #1847, opened 2018-08-03
gunicorn with eventlet doesn't perform eventlet monkey patching since 19.8.0


capnfabs:
Hi there,

I think the fix for https://github.com/benoitc/gunicorn/issues/1584 (in commit 120f503f68929ef98d721a406745a1dbbb9d83fe) inadvertently prevents eventlet monkey patching from running upon worker startup.

My understanding is:
- EventletWorker.init_process now calls `super(EventletWorker, self).init_process()` _before_ `patch()`
- The super method `Worker.init_process()` enters a run loop, and never returns.

This means that eventlet never gets monkey patched by the gunicorn worker init process.

The workaround we've used in the meantime is to just run eventlet monkey patching on startup of our application code, but it's probably worth fixing this if we want to retain the expectation that gunicorn will install eventlet monkey patching.

Happy to send a PR, but wanted to check that I understood what was going on correctly first :)

tilgovi:
Your understanding sounds correct to me. If you can take #1584 into account with a PR that would be great. Thank you!

benoitc:
it actually does patching but after the process is initialized. What is the issue you're experimenting?

tilgovi:
My understanding is that it does not, @benoitc, since `super(EventletWorker, self).init_process()` never returns.

colincadams:
This is still an issue, what is blocking #1884 from being merged?

travelsalesman:
@tilgovi Eventlet has fixed the issue by moving the module which needed the SIGCHILD handler to import time:
https://github.com/eventlet/eventlet/commit/82f1877ff5b950e1ee9debbb8be880ebedb8abcb

Which is before gunicorn starts handling  SIGCHILD, so https://github.com/benoitc/gunicorn/commit/120f503f68929ef98d721a406745a1dbbb9d83fe  can be reverted, i.e. this PR.

Currently eventlet workers run without greening so are pretty much pointless

tilgovi:
Perfect. Let's merge the revert with a PR that bumps the required Eventlet version to 0.24.1.

PR #1884, opened 2018-09-26
fix: eventlet worker never calls monkey_patch()

xgfone:
The method `init_process()` of the parent worker will call the method `run()` in the end, but `run()` never exits until the worker is over. So we must call the method `patch()` before calling `init_process()` of the parent worker. Or `eventlet.monkey_patch()` is never called.

To refer to the gevent worker, calling `patch()` is before `init_process()`.


xgfone:
Repeat

benoitc:
?

xgfone:
@benoitc I just see #1847. I think it has been fixed.

xgfone:
Hasn't it been fixed? If that, Can I reopen this PR? @benoitc

tilgovi:
Re-opening.

tilgovi:
@benoitc could you check this one? If eventlet monkey patching is broken, we should fix it for R20.

tilgovi:
I think the fix we need may be to reset the `SIGCHLD` handler after fork so that the workers are not catching it, by using `signal.signal(signal.SIGCHLD, signal.SIGDFL)`.

I understand from reading #1584 that Gunicorn was catching a `SIGCHLD`. Since the Eventlet worker calls `patch` before its base class calls `init_signals`, my only guess is that the `SIGCHLD` handler installed by the Arbiter is inherited across the `fork()` and maybe it needs to be cleared for Eventlet to initialize properly.

colincadams:
Can this PR be merged as is? Without it, patching never happens.

tilgovi:
@colincadams not until someone can explain an alternative way to address #1584, which is what c2757bd84152e0c2d5f659fa199509fa67e8469e tried to fix.

PR #1887, opened 2018-10-01
Prevent removal unix socket for reuse_port

sirkonst:
If you have two (or more) instances of gunicorn that use `reuse-port`
and bind to single unix socket all work until one of gunicorn will
stopped. Because the first stopped removes unix socket file and other
instances can't longer process requests.


tilgovi:
Makes sense to me.

PR #1897, opened 2018-10-13
add systemd sd_notify support

gdamjan:
roughly based on sd_notify() from systemd and https://github.com/bb4242/sdnotify

only implements `READY=1` and `STATUS=Gunicorn arbiter booted` of the
protocol in the arbiter. in the future, reloads can be notified, and
possibly also other statuses.

see https://www.freedesktop.org/software/systemd/man/sd_notify.html for
more info

sd_notify() is a noop when not run in a systemd service (i.e
NOTIFY_SOCKET environment variable is not set)

### DIFF
--- a/gunicorn/systemd.py
+++ b/gunicorn/systemd.py
@@ -4,6 +4,7 @@
 # See the NOTICE for more information.
 
 import os
+import socket
 
 SD_LISTEN_FDS_START = 3
 
@@ -43,3 +44,37 @@ def listen_fds(unset_environment=True):
         os.environ.pop('LISTEN_FDS', None)
 
     return fds
+
+
+def sd_notify(state, unset_environment=False, debug=False):
+    """Send a notification to systemd. state is a string; see
+    the man page of sd_notify (http://www.freedesktop.org/software/systemd/man/sd_notify.html)
+    for a description of the allowable values.
+
+    If the unset_environment parameter is non-zero, sd_notify() will unset the $NOTIFY_SOCKET environment variable before
+    returning (regardless of whether the function call itself succeeded or not). Further calls to sd_notify() will then fail, but
+    the variable is no longer inherited by child processes.
+
+    Normally this method silently ignores exceptions (for example, if the
+    systemd notification socket is not available) to allow applications to
+    function on non-systemd based systems. However, setting debug=True will
+    cause this method to raise any exceptions generated to the caller, to
+    aid in debugging."""
+
+    addr = os.getenv('NOTIFY_SOCKET')
+    if addr == None:
+        # not run in a service, just a noop
+        return
+    if unset_environment:
+        os.unsetenv('NOTIFY_SOCKET')
+    try:
+        sock = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM | socket.SOCK_CLOEXEC)
+        if addr[0] == '@':
+            addr = '\0' + addr[1:]
+        sock.connect(addr)
+        sock.sendall(state.encode('utf-8'))
+    except:
+        if debug:
+            raise
+    finally:
+        sock.close()
--- a/gunicorn/systemd.py
+++ b/gunicorn/systemd.py
@@ -63,12 +63,12 @@ def sd_notify(state, unset_environment=False, debug=False):
     cause this method to raise any exceptions generated to the caller, to
     aid in debugging."""
 
-    addr = os.getenv('NOTIFY_SOCKET')
+    addr = os.environ.get('NOTIFY_SOCKET')
     if addr is None:
         # not run in a service, just a noop
         return
     if unset_environment:
-        os.unsetenv('NOTIFY_SOCKET')
+        os.environ.pop('NOTIFY_SOCKET')
     try:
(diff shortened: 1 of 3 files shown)


funkyfuture:
this is superb. this solves the issue addressed in #1895 without the need to move the socket out of `/var/app_namespace/` and also stops restarting the service continuously after the timeout period.

gunicorn/systemd.py
```diff
@@ -43,3 +44,39 @@ def listen_fds(unset_environment=True):
         os.environ.pop('LISTEN_FDS', None)
 
     return fds
+
+
+def sd_notify(state, unset_environment=False, debug=False):
```
berkerpeksag: Currently, there is no way to actually override the default values of `unset_environment` and `debug`. I think we either should find a way to customize the behavior (e.g. by passing `debug=self.cfg.reload` to `systemd.sd_notify()` in `Arbiter.start()`) or drop these arguments.

I don't use systemd, but I'm not sure hiding exceptions by default is a good idea.

gunicorn/systemd.py
```diff
@@ -43,3 +44,39 @@ def listen_fds(unset_environment=True):
...
+    sd_notify() will then fail, but the variable is no longer inherited by
+    child processes.
+
+    Normally this method silently ignores exceptions (for example, if the
+    systemd notification socket is not available) to allow applications to
+    function on non-systemd based systems. However, setting debug=True will
+    cause this method to raise any exceptions generated to the caller, to
+    aid in debugging."""
+
+    addr = os.getenv('NOTIFY_SOCKET')
```
berkerpeksag: Usually the `os.environ` API is the preferred way to get an environment variable.

gunicorn/systemd.py
```diff
@@ -43,3 +44,39 @@ def listen_fds(unset_environment=True):
...
+    if addr is None:
+        # not run in a service, just a noop
+        return
+    if unset_environment:
+        os.unsetenv('NOTIFY_SOCKET')
```
berkerpeksag: Same as above. We could use `os.environ.pop()`.

berkerpeksag: Also, any reason not to do this inside the `finally` block?

gunicorn/systemd.py
```diff
@@ -43,3 +44,39 @@ def listen_fds(unset_environment=True):
...
+    try:
+        sock = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM | socket.SOCK_CLOEXEC)
+        if addr[0] == '@':
+            addr = '\0' + addr[1:]
+        sock.connect(addr)
+        sock.sendall(state.encode('utf-8'))
```
berkerpeksag: This is just a question, I'm not sure which way is better, but instead of encoding a str object every time we call this function, can't we just pass a bytes object directly? We could even add a `isinstance(state, bytes)` at the top of the function if we want to be safe.

gunicorn/systemd.py
gdamjan: > but I'm not sure hiding exceptions by default is a good idea.

in `arbiter.py` the call to `systemd.sd_notify` is unconditional, so in the weird case that the environment variable exists but is not quite the socket expected, I've added the additional try/except. other than that, there's not much to fail (I think).

> Currently, there is no way to actually override the default values of unset_environment and debug

I've left those there in case some other uses of sd_notify appears in the future.

gunicorn/systemd.py
gdamjan: I've never heard it's the preferred way. Fine with me either way.

gunicorn/systemd.py
gdamjan: I'd leave it at a string interface, seems more logical to me. also, systemd already specifies it expects utf-8.

 allowing for bytes with a `isinstance` maybe makes sense, but then I'd need to worry about non-utf-8 bytes

gunicorn/systemd.py
gdamjan: I think I had a reason. but let me think about it.

benoitc:
@gdamjan ping :)

gunicorn/systemd.py
benoitc: i think i would remove the default from the signature if it's not needed and maybe add a comment about it. Thoughts?

gunicorn/systemd.py
benoitc: > I think I had a reason. but let me think about it.

did you find the reason?

gunicorn/systemd.py
benoitc: looking at the code if the unset environnement is really needed i would expose it as an option.

for the debug log I would just reuse the logger instance and pass it to the function if possible.

gunicorn/systemd.py
gdamjan: no special reason. I doesn't matter to me any way.

gunicorn/systemd.py
gdamjan: maybe I should use `self.log.exception` ?

gdamjan: how about d71e746c ?

benoitc: looks fine for me. is ˋunset_environment` is needed maybe it can be a setting from the config ?

benoitc:
merged, thanks !

PR #1902, opened 2018-10-24
Travis/Tox version updates

javabrett:
- run lint using latest Python 3.7
- updated py env list in tox.ini
  - added py38-dev
  - removed py36-dev

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -7,7 +7,6 @@ usedevelop = True
 commands = py.test {posargs}
 deps =
   -rrequirements_test.txt
-  py{34,35,36,37,38-dev}: aiohttp
 
 [testenv:lint]
 commands =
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,13 +1,13 @@
...
+envlist = py34, py35, py36, py37, py38-dev, pypy, lint
 skipsdist = True
 
 [testenv]
 usedevelop = True
 commands = py.test {posargs}
 deps =
   -rrequirements_test.txt
-  py{34,35,36,36-dev,37}: aiohttp
+  py{34,35,36,37,38-dev}: aiohttp
```
berkerpeksag: Actually, we can move aiohttp into `requirements_test.txt` now.

javabrett: Right, will do - I previously noticed it was deprecated.

berkerpeksag:
Thanks!

PR #1037, opened 2015-05-21
Document why REMOTE_ADD may not be the user's IP address

Starefossen:
This PR updates the Gunicorn deploy documentation on why `REMOTE_ADDR` may not always be the IP address of the user.

Close: #1035
Related: #633 

Signed-off-by: Hans Kristian Flaatten <hans.kristian.flaatten@turistforeningen.no>

PR #1906, opened 2018-10-27
Provide guidance on X-Forwarded-For access log.

rduplain:
The doc change introduced in #1037 is initially helpful but then internally inconsistent. It correctly points out that X-Forwarded-For is no longer used in setting REMOTE_ADDR (c4873681299212d6082cd9902740eef18c2f14f1), but then confusingly indicates a solution using the X-Forwarded-For header.

The deployment doc provides a full configuration example which includes proxy headers as set by Nginx. What is missing, before this patch, is a clear suggestion on how to make use of the header in gunicorn's access log.

Accordingly, remove the confusing suggestion and replace it with a drop-in replacement for gunicorn's default access log format.

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -67,13 +67,12 @@ Gunicorn 19 introduced a breaking change concerning how ``REMOTE_ADDR`` is
 handled. Previous to Gunicorn 19 this was set to the value of
 ``X-Forwarded-For`` if received from a trusted proxy. However, this was not in
 compliance with :rfc:`3875` which is why the ``REMOTE_ADDR`` is now the IP
-address of **the proxy** and **not the actual user**. You should instead
-configure Nginx to send the user's IP address through the ``X-Forwarded-For``
-header like this::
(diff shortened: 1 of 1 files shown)


docs/source/deploy.rst
```diff
@@ -67,13 +67,12 @@ Gunicorn 19 introduced a breaking change concerning how ``REMOTE_ADDR`` is
...
 compliance with :rfc:`3875` which is why the ``REMOTE_ADDR`` is now the IP
-address of **the proxy** and **not the actual user**. You should instead
-configure Nginx to send the user's IP address through the ``X-Forwarded-For``
-header like this::
+address of **the proxy** and **not the actual user**.
 
-    ...
-    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-    ...
+To have access logs indicate **the actual user** IP, use ``access-logformat``
```
berkerpeksag: \`\`access-logformat\`\` ->:ref:\`access_log_format\`

berkerpeksag:
Thanks!

PR #1928, opened 2018-11-29
Add server_tokens configuration option

### LINKED ISSUE #825: Ability to mask the http response header Server attribute
Right now, all responses have the http header
server: gunicorn/19.0.0

For security reasons, I would like to be able to mask that so that people won't know to search for security vulnerabilities in gunicorn. Is there any way to mask it? via a setting perhaps?

mkgs:
Fixes #825

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -1095,6 +1095,23 @@ The variables are passed to the the PasteDeploy entrypoint. Example::
 
 .. versionadded:: 19.7
 
+.. _server-tokens:
+
+server_tokens
+~~~~~~~~~~~~~
+
+* ``--server-tokens``
+* ``true``
+
+Enable or disable emitting gunicorn version in the ``Server`` http response header.
+
+This setting may also be used to emit a custom ``Server`` header value
+or disable the header entirely.
+
+A value of ``true`` will result in the default behavior. ``false`` will disable
+emittance of gunicorn version. A string value will emit itself as a custom ``Server``
+header value, and an empty string will disable the ``Server`` response header.
+
 Server Socket
 -------------
 
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -17,7 +17,7 @@ import ssl
 import sys
 import textwrap
 
-from gunicorn import __version__, util
+from gunicorn import __version__, util, SERVER_SOFTWARE
 from gunicorn.errors import ConfigError
 from gunicorn.reloader import reloader_engines
 
@@ -220,6 +220,17 @@ class Config(object):
             global_conf[k] = v
 
         return global_conf
+
+    @property
+    def server_tokens(self):
+        s = self.settings['server_tokens'].get()
+        if s == 'true':
+            return SERVER_SOFTWARE
+
+        if s == 'false':
+            return 'gunicorn'
+
+        return s
 
 
 class SettingMeta(type):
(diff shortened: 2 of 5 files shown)


mkgs:
Expanded on earlier PR from @mathiasverhoeven and @mprzytulski, w/ nginx-like --server-tokens true|false|string syntax.

docs/source/settings.rst
```diff
@@ -1095,6 +1095,23 @@ The variables are passed to the the PasteDeploy entrypoint. Example::
 
 .. versionadded:: 19.7
 
+.. _server-tokens:
+
+server_tokens
+~~~~~~~~~~~~~
+
+* ``--server-tokens``
+* ``true``
```
javabrett: Does not match Python `True` or value for other settings.

gunicorn/config.py
```diff
@@ -221,6 +221,17 @@ def paste_global_conf(self):
 
         return global_conf
 
+    @property
+    def server_tokens(self):
+        s = self.settings['server_tokens'].get()
+        if s == 'true':
```
javabrett: Does not match Python `True` or value for other settings.

gunicorn/config.py
```diff
@@ -221,6 +221,17 @@ def paste_global_conf(self):
...
+            return SERVER_SOFTWARE
+
+        if s == 'false':
```
javabrett: Might be nice to have the option to just simply disable the header altogether, not even leaking `gunicorn`.

gunicorn/http/wsgi.py
```diff
@@ -300,10 +299,11 @@ def default_headers(self):
         headers = [
             "HTTP/%s.%s %s\r\n" % (self.req.version[0],
                 self.req.version[1], self.status),
-            "Server: %s\r\n" % self.version,
             "Date: %s\r\n" % util.http_date(),
             "Connection: %s\r\n" % connection
         ]
+        if self.cfg and self.cfg.server_tokens:
```
javabrett: I haven't tested this, but with reference to my previous comments - if you can set an empty string then this is falsey and will be skipped.  Do we still need true/True/false/False?

gunicorn/http/wsgi.py
```diff
@@ -300,10 +299,11 @@ def default_headers(self):
...
+            headers.append("Server: %s\r\n" % self.cfg.server_tokens)
```
javabrett: Noting that the header order will probably change here, which I suppose is not significant.

tests/test_http.py
```diff
@@ -226,3 +226,42 @@ def test_eof_reader_read_invalid_size():
...
+    c.set('server_tokens', 'server software')
+    response = Response(mocked_request, mocked_socket, c)
+    headers = response.default_headers()
+    assert get_server_header(headers) == 'server software'
+
+    # disabled server header
+    c.set('server_tokens', '')
+    response = Response(mocked_request, mocked_socket, c)
+    headers = response.default_headers()
+    assert get_server_header(headers) is None
```
javabrett: Nice.

mkgs:
see https://github.com/benoitc/gunicorn/pull/1617#issuecomment-369581668 for some additional context around the ```true|false|string``` options.

benoitc:
following my suggestion and comments from @tilgovi & @jamadden I'm closing this issue favor of #2233

PR #1957, opened 2019-01-21
Simplify Paste Deployment integration

### LINKED ISSUE #1189: Remove paste server block support
In issues like #1073 the paste `server` block has come up. Right now, gunicorn supports some config options there.

I'm tempted to propose deprecate support for that completely so that all the gunicorn config has to happen in a gunicorn config file.

Feedback requested.

tilgovi:
Remove the `gunicorn_paster` command. With the `--paste` option to the
`gunicorn` command, Gunicorn will no longer read the server section of
the configuration. Instead, server configuration must be done with
Gunicorn configuration files, command line switches, and environment
variables.

The use of Gunicorn as a Paste Deployment server factory is no longer
deprecated. It allows specifying `host` and `port`, as well as `bind`,
but is otherwise more strict with options than in the past. Rather than
ignoring unknown options it will raise an error.

Close #1189

### DIFF
--- a/gunicorn/app/wsgiapp.py
+++ b/gunicorn/app/wsgiapp.py
@@ -13,22 +13,21 @@ from gunicorn import util
 class WSGIApplication(Application):
     def init(self, parser, opts, args):
         if opts.paste:
-            app_name = 'main'
-            path = opts.paste
-            if '#' in path:
-                path, app_name = path.split('#')
-            path = os.path.abspath(os.path.normpath(
-                os.path.join(util.getcwd(), path)))
+            from .pasterapp import has_logging_config
 
-            if not os.path.exists(path):
-                raise ConfigError("%r not found" % path)
+            config_uri = os.path.abspath(opts.paste)
+            config_file = config_uri.split('#')[0]
 
-            # paste application, load the config
-            self.cfgurl = 'config:%s#%s' % (path, app_name)
-            self.relpath = os.path.dirname(path)
+            if not os.path.exists(config_file):
+                raise ConfigError("%r not found" % config_file)
 
-            from .pasterapp import paste_config
-            return paste_config(self.cfg, self.cfgurl, self.relpath)
+            self.cfg.set("default_proc_name", config_file)
+            self.app_uri = config_uri
+
+            if has_logging_config(config_file):
+                self.cfg.set("logconfig", config_file)
+
+            return
 
         if not args:
             parser.error("No application module specified.")
@@ -37,13 +36,11 @@ class WSGIApplication(Application):
         self.app_uri = args[0]
 
     def load_wsgiapp(self):
-        # load the app
         return util.import_app(self.app_uri)
 
     def load_pasteapp(self):
-        # load the paste app
-        from .pasterapp import load_pasteapp
-        return load_pasteapp(self.cfgurl, self.relpath, global_conf=self.cfg.paste_global_conf)
+        from .pasterapp import get_wsgi_app
+        return get_wsgi_app(self.app_uri, defaults=self.cfg.paste_global_conf)
 
     def load(self):
         if self.cfg.paste is not None:
--- a/docs/source/run.rst
+++ b/docs/source/run.rst
@@ -4,8 +4,9 @@ Running Gunicorn
 
 .. highlight:: bash
 
-You can run Gunicorn by using commands or integrate with Django or Paster. For
-deploying Gunicorn in production see :doc:`deploy`.
+You can run Gunicorn by using commands or integrate with popular frameworks
+like Django, Pyramid, or TurboGears. For deploying Gunicorn in production see
+:doc:`deploy`.
 
 Commands
 ========
@@ -78,7 +79,7 @@ See :ref:`configuration` and :ref:`settings` for detailed usage.
 Integration
 ===========
 
-We also provide integration for both Django and Paster applications.
+Gunicorn also provides integration for Django and Paste Deploy applications.
 
 Django
 ------
@@ -104,13 +105,40 @@ option::
 
     $ gunicorn --env DJANGO_SETTINGS_MODULE=myproject.settings myproject.wsgi
 
-Paste
------
+Paste Deployment
+----------------
 
-If you are a user/developer of a paste-compatible framework/app (as
-Pyramid, Pylons and Turbogears) you can use the
-`--paste <http://docs.gunicorn.org/en/latest/settings.html#paste>`_ option
-to run your application.
+Frameworks such as Pyramid and Turbogears are typically configured using Paste
+Deployment configuration files. If you would like to use these files with
+Gunicorn, there are two approaches.
+
+As a server runner, Gunicorn can serve your application using the commands from
+your framework, such as ``pserve`` or ``gearbox``. To use Gunicorn with these
+commands, specify it as a server in your configuration file:
+
+.. code-block:: ini
+
+    [server:main]
+    use = egg:gunicorn#main
+    host = 127.0.0.1
+    port = 8080
+    workers = 3
+
+This approach is the quickest way to get started with Gunicorn, but there are
+some limitations. Gunicorn will have no control over how the application is
+loaded, so settings such as reload_ will have no effect and Gunicorn will be
+unable to hot upgrade a running application. Using the daemon_ option may
+confuse your command line tool. Instead, use the built-in support for these
+features provided by that tool. For example, run ``pserve --reload`` instead of
+specifying ``reload = True`` in the server configuration block. For advanced
+configuration of Gunicorn, such as `Server Hooks`_ specifying a Gunicorn
+configuration file using the ``config`` key is supported.
+
+To use the full power of Gunicorn's reloading and hot code upgrades, use the
+`paste option`_ to run your application instead. When used this way, Gunicorn
+will use the application defined by the PasteDeploy configuration file, but
+Gunicorn will not use any server configuration defined in the file. Instead,
+`configure gunicorn`_.
 
 For example::
 
@@ -120,4 +148,13 @@ Or use a different application::
 
     $ gunicorn --paste development.ini#admin -b :8080 --chdir /path/to/project
 
-It is all here. No configuration files nor additional Python modules to write!
+With both approaches, Gunicorn will use any loggers section found in Paste
+Deployment configuration file, unless instructed otherwise by specifying
+additional `logging settings`_.
+
+.. _reload: http://docs.gunicorn.org/en/latest/settings.html#reload
+.. _daemon: http://docs.gunicorn.org/en/latest/settings.html#daemon
+.. _Server Hooks: http://docs.gunicorn.org/en/latest/settings.html#server-hooks
+.. _paste option: http://docs.gunicorn.org/en/latest/settings.html#paste
+.. _configure gunicorn: http://docs.gunicorn.org/en/latest/configure.html
+.. _logging settings: http://docs.gunicorn.org/en/latest/settings.html#logging
--- a/gunicorn/app/pasterapp.py
+++ b/gunicorn/app/pasterapp.py
@@ -3,206 +3,73 @@
 # This file is part of gunicorn released under the MIT license.
 # See the NOTICE for more information.
 
-# pylint: skip-file
-
+import configparser
 import os
-import pkg_resources
-import sys
 
-try:
-    import configparser as ConfigParser
-except ImportError:
-    import ConfigParser
+from paste.deploy import loadapp
 
-from paste.deploy import loadapp, loadwsgi
-SERVER = loadwsgi.SERVER
-
-from gunicorn.app.base import Application
-from gunicorn.config import Config, get_default_config_file
-from gunicorn import util
+from gunicorn.app.wsgiapp import WSGIApplication
+from gunicorn.config import get_default_config_file
 
 
-def _has_logging_config(paste_file):
-    cfg_parser = ConfigParser.ConfigParser()
-    cfg_parser.read([paste_file])
-    return cfg_parser.has_section('loggers')
+def get_wsgi_app(config_uri, name=None, defaults=None):
+    if ':' not in config_uri:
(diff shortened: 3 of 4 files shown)


tilgovi:
In addition to this, I wonder when is a good time to `s/paster/paste/g`.

tilgovi:
Question: should we error when some options are set that cannot be honored under a paste server runner? These options are often available in the script (`pserve`, `paster serve`, etc).

- `preload` (always preloaded)
- `reload` (and all the related options)
- `daemon`
- `paste` (makes no sense)
- `chdir` (too late)

tilgovi:
An alternative to consider:

- A command line argument is still required but specifies the config file
- The `--paste` option only enables paste support, but takes no argument
- Maybe a `--paste-app` that defaults to the fragment from the config uri (or `main`)
- Maybe a `--paste-server` that defaults to the fragment from the config uri (or `main`)
- Both `--paste-app` and `--paste-server` imply `--paste`

Something like: `gunicorn --paste-server websocket production.ini` would run the `server:websocket` block with the application defined in the `app:main` section of the `production.ini` file.

tilgovi:
I made `--paste` support is as minimal as possible. It will load the application from the configuration file. As a convenience, it will default to loading the logging configuration from the configuration file if it has a `loggers` section. It will not load Gunicorn server settings from the configuration file.

The `gunicorn_paster` script is removed. Gunicorn no longer provides a general Paste Deploy CLI tool. The `--paste` option is not a replacement for that.

Instead, support for using Gunicorn as a Paste Deploy server is no longer deprecated.

 If users of frameworks that rely on Paste Deploy want to continue to use other scripts (`pserve`, `paster`, etc), they can. Gunicorn can be configured as the server for these scripts.

Some features of Gunicorn will not work when used as a Paste Deploy server. For example, Gunicorn receives the application as an argument, meaning it behaves as though `--preload` were in use. In many cases, these features are supplied by Paste Deploy CLI instead. For example, the `paster` command from `pastescript` and Pyramid's `pserve` command each have their own `--reload` option.

tilgovi:
I'm also interested to just remove the `PasterApplication`. It does nothing but treat the first positional CLI argument as the `--paste` argument. I know that `r2` code that is Reddit's old code used it, but I don't see much use reading through search results on GitHub and it can be easily replaced with `--paste`.

benoitc:
> Remove the gunicorn_paster command. With the --paste option to the
gunicorn command, Gunicorn will no longer read the server section of
the configuration. Instead, server configuration must be done with
Gunicorn configuration files, command line switches, and environment
variables.


why is this? Is this common for the users though? The main point here was to allow them to just use their config file but i'm not sure about the new usages with paster these days

benoitc:
> In addition to this, I wonder if there is a good time or reason to s/paster/paste/g. The format is "paste", while "paster" is the CLI script name of pastescript.

I used the same naming as in uwsi and another tool i can't remember :)

benoitc:
> My thinking is that it's unnecessary. Not all options are going to work when run by another paste script runner.

I agree. That's the job of gunicorn to take care of such things

tilgovi:
I removed support for the `[server:main]` section because I think it's confusing two things.

When using a CLI runner for Paste Deploy (`paster serve`, `waitress-serve`, `pserve`), the runner parses the `[server:main]` section and passes the options to the server factory.

When using `gunicorn --paste`, Gunicorn is the server and it is only loading the application defined by the `[app:main]` section.

What happens with other CLI runners that support the server block is that they load the server and call it. For Gunicorn, we were loading the _options_ only from the server block and than _assuming_ that the server block was meant for Gunicorn.

To handle the server block correctly for `--paste` we would want Gunicorn to ask Paste Deploy to load the server and then pass it the app. That's confusing, because it means Gunicorn could run Waitress or `wsgiref` or whatever else is in the `[server:main]` block. It made sense for `gunicorn_paster` to do this, but I think `gunicorn` could be more strict and only support apps. It is not meant to replace these other runners.

Or, we could support it but do the right thing. It might even be simpler than what we had before because we let Paste Deploy do most of the work.

benoitc:
@tilgovi 
> Or, we could support it but do the right thing. It might even be simpler than what we had before because we let Paste Deploy do most of the work.

fine for me. If it's not much work maybe we can do it for the current release? If not i would postpone it for the next release and open a ticket for it. thoughts?

tilgovi:
> If it's not much work maybe we can do it for the current release? 

I am trying it right now but I'm afraid there are too many questions right now and too confusing. I think it's best not to support it. The reason is that so many Gunicorn options don't make sense then. If you run `--worker-class gevent` but then the `[server:main]` block says `use = egg:waitress#main` then actually what you would get is waitress, not a Gunicorn gevent worker, with the app preloaded.

I think this is just not what the `gunicorn` CLI should do.

tilgovi:
What this PR requires from users is that they choose:

1. Use `gunicorn` and set options from the Gunicorn configuration file or command line flags, but they can run a Paste Deploy application using `--paste`.

2. Use another Paste Deploy server runner and specify `use = egg:gunicorn#main`, and they can specify Gunicorn options in the INI but they lose some control (app will be preloaded by the runner).

tilgovi:
I would not remove `--paste` yet because we never deprecated it. Maybe the best thing is to keep the old support for reading the server block and deprecate, then remove it?

We can remove `gunicorn_paster` because that's been deprecated before.

This PR makes `use = egg:gunicorn#main` no longer deprecated.

Does that sound good?

tilgovi:
> Maybe the best thing is to keep the old support for reading the server block and deprecate, then remove it?

Or we can just remove it if we provide enough help with documentation.

benoitc:
> 
> Similarly, I think Gunicorn should work as a server when run from another script that supports running Paste Deploy servers, but it should not replace those scripts.

yes.
> 
> We could remove `--paste`.

This is where I disagree. `--paste` is useful for anyone that want true supervision and no application, graceful reload and other goodies. And since it already work I would like to keep it. 

Following your comments I think it would be a good idea to highlight the usage of Gunicorn inside a paste deploy script and the limitations and advantages  of both methods inside the doc imo. Thoughts?

benoitc:
> 

> I would not remove `--paste` yet because we never deprecated it. Maybe the best thing is to keep the old support for reading the server block and deprecate, then remove it?

for the reason given above I would not deprecate it, just keep it aside and document the usage of both.
> 
> We can remove `gunicorn_paster` because that's been deprecated before.
> 
> This PR makes `use = egg:gunicorn#main` no longer deprecated.
> 
> Does that sound good?

yes :)

benoitc:
@tilgovi test failed on python 3.7.1

```log
$ python --version
Python 3.7.1
$ pip --version
pip 18.1 from /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/pip (python 3.7)
install
0.93s$ pip install tox
Collecting tox
  Using cached https://files.pythonhosted.org/packages/4a/bd/50cd91221508db5bb66e211ac327d0da4a5db6d0db5ad35d6fbf1e4af5dc/tox-3.7.0-py2.py3-none-any.whl
Requirement already satisfied: six<2,>=1.0.0 in /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages (from tox) (1.11.0)
Requirement already satisfied: py<2,>=1.4.17 in /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages (from tox) (1.7.0)
Collecting toml>=0.9.4 (from tox)
  Using cached https://files.pythonhosted.org/packages/a2/12/ced7105d2de62fa7c8fb5fce92cc4ce66b57c95fb875e9318dba7f8c5db0/toml-0.10.0-py2.py3-none-any.whl
Requirement already satisfied: virtualenv>=1.11.2 in /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages (from tox) (16.1.0)
Requirement already satisfied: pluggy<1,>=0.3.0 in /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages (from tox) (0.8.0)
Requirement already satisfied: setuptools>=30.0.0 in /home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages (from tox) (40.6.3)
Collecting filelock<4,>=3.0.0 (from tox)
  Using cached https://files.pythonhosted.org/packages/d7/ca/3c74396a9ed8a4cfab5459800edeef9a1269591cb21f5a49bd71a49c5fa2/filelock-3.0.10-py3-none-any.whl
Installing collected packages: toml, filelock, tox
Successfully installed filelock-3.0.10 toml-0.10.0 tox-3.7.0
30.81s$ tox --recreate
lint recreate: /home/travis/build/benoitc/gunicorn/.tox/lint
lint installdeps: pylint
lint develop-inst: /home/travis/build/benoitc/gunicorn
lint installed: astroid==2.1.0,-e git+https://github.com/benoitc/gunicorn.git@268c7034e0d9c92a93eeaa3fcfa76a66385b72a5#egg=gunicorn,isort==4.3.4,lazy-object-proxy==1.3.1,mccabe==0.6.1,pylint==2.2.2,six==1.12.0,wrapt==1.11.1
lint run-test-pre: PYTHONHASHSEED='2834214185'
lint runtests: commands[0] | pylint -j0 gunicorn tests/test_arbiter.py tests/test_config.py tests/test_gaiohttp.py tests/test_http.py tests/test_invalid_requests.py tests/test_logger.py tests/test_pidfile.py tests/test_sock.py tests/test_ssl.py tests/test_statsd.py tests/test_systemd.py tests/test_util.py tests/test_valid_requests.py
************* Module gunicorn.app.pasterapp
gunicorn/app/pasterapp.py:9:0: W0611: Unused loadserver imported from paste.deploy (unused-import)
------------------------------------
Your code has been rated at 10.00/10
ERROR: InvocationError for command '/home/travis/build/benoitc/gunicorn/.tox/lint/bin/pylint -j0 gunicorn tests/test_arbiter.py tests/test_config.py tests/test_gaiohttp.py tests/test_http.py tests/test_invalid_requests.py tests/test_logger.py tests/test_pidfile.py tests/test_sock.py tests/test_ssl.py tests/test_statsd.py tests/test_systemd.py tests/test_util.py tests/test_valid_requests.py' (exited with code 4)
___________________________________ summary ____________________________________
ERROR:   lint: commands failed
The command "tox --recreate" exited with 1.
cache.2
store build cache
0.00s8.50schange detected (content changed, file is created, or file is deleted):
/home/travis/build/benoitc/gunicorn/.tox/lint/bin/gunicorn_paster
/home/travis/build/benoitc/gunicorn/.tox/lint/lib/python3.7/site-packages/astroid/brain/__pycache__/brain_argparse.cpython-37.pyc
/home/travis/build/benoitc/gunicorn/.tox/lint/lib/python3.7/site-packages/astroid/brain/__pycache__/brain_attrs.cpython-37.pyc
/home/travis/build/benoitc/gunicorn/.tox/lint/lib/python3.7/site-packages/astroid/brain/__pycache__/brain_builtin_inference.cpython-37.pyc
/home/travis/build/benoitc/gunicorn/.tox/lint/lib/python3.7/site-packages/astroid/brain/__pycache__/brain_collections.cpython-37.pyc
/home/travis/build/benoitc/gunicorn/.tox/lint/lib/python3.7/site-packages/astroid/brain/__pycache__/brain_curses.cpython-37.pyc
/home/travis/build/benoitc/gunicorn/.tox/lint/lib/python3.7/site-packages/astroid/brain/__pycache__/brain_dateutil.cpython-37.pyc
/home/travis/build/benoitc/gunicorn/.tox/lint/lib/python3.7/site-packages/astroid/brain/__pycache__/brain_fstrings.cpython-37.pyc
/home/travis/build
...
changes detected, packing new archive
.
uploading simplify-paste-integrations/cache-linux-xenial-886521dd26aa2aaa6e97504d1d7c4758ce511240339b27af9468f9899b2c7fb2--python-3.7.tgz
cache uploaded
Done. Your build exited with 1.
```

https://travis-ci.org/benoitc/gunicorn/jobs/482826354

tilgovi:
> test failed on python 3.7.1

Thanks. I was experimenting with adding back support for the server section, and I left that in by accident.

tilgovi:
I fixed the lint error and updated the documentation. Let me know what you think. I will check back in a few hours.

berkerpeksag:
>  I know that r2 code that is Reddit's old code used it, but I don't see much use reading through search results on GitHub and it can be easily replaced with --paste.

They seem to deprecate/archive the r2 repository a while ago, so we may safely ignore that use case. Still, it would be good if we could mention the use case in the release notes.

I will try to review this PR tonight.

tilgovi:
> They seem to deprecate/archive the r2 repository a while ago, so we may safely ignore that use case. Still, it would be good if we could mention the use case in the release notes.

I agree. I think we should probably remove `PasterApplication` and mention it in the release notes. It was the internal class used by the `gunicorn_paster` command, which is removed in this PR. I'll update to remove it now.

gunicorn/app/wsgiapp.py
```diff
@@ -13,22 +13,21 @@
...
-            if not os.path.exists(path):
-                raise ConfigError("%r not found" % path)
+            config_uri = os.path.abspath(opts.paste)
+            config_file = config_uri.split('#')[0]
 
-            # paste application, load the config
-            self.cfgurl = 'config:%s#%s' % (path, app_name)
-            self.relpath = os.path.dirname(path)
+            if not os.path.exists(config_file):
+                raise ConfigError("%r not found" % config_file)
```
tilgovi: Should this be `parser.error()` or `parser.exit()`?

benoitc: do you want to change this error?

tilgovi: Maybe it's good to show usage here.

How do you feel if we do `parser.error("%r not found" % opts.paste)`?

benoitc: hrm but the issue is that it wouldn't be catch by the `cli`, would it?

tilgovi:
I'll give @berkerpeksag time to review it, but you both can feel free to merge. I will make no more changes.

PR #1958, opened 2019-01-21
Fix root logging

boytm:
root and logger are same level. https://docs.python.org/2/library/logging.config.html#dictionary-schema-details


benoitc:
@berkerpeksag i will not be abble to test it until late so feel free to merge it if everything works :)

tilgovi:
Tested.

PR #1962, opened 2019-01-22
Bump minimum Eventlet and Gevent versions

tilgovi:
Update the Eventlet and Gevent worker classes to check for versions of
Eventlet and Gevent that are stable under Python 3 and remove outdated
compatibility code.


gunicorn/workers/geventlet.py
```diff
@@ -14,9 +14,8 @@
     raise RuntimeError("You need eventlet installed to use this worker.")
 
 # validate the eventlet version
-if eventlet.version_info < (0, 9, 7):
-    raise RuntimeError("You need eventlet >= 0.9.7")
-
+if eventlet.version_info < (0, 19):
```
tilgovi: The [NEWS] file for Eventlet mentions a lot of important Python 3 fixes for 0.18, but the first two patch releases in that series have serious issues. I chose 0.19 to be safe because it includes a number of other fixes that look good and that version is almost three years old now, anyway, so it's not an aggressive upgrade.

gunicorn/workers/ggevent.py
```diff
@@ -10,20 +10,19 @@
...
-if sys.platform == "darwin":
-    os.environ['EVENT_NOKQUEUE'] = "1"
-
 try:
     import gevent
 except ImportError:
     raise RuntimeError("You need gevent installed to use this worker.")
+
+# validate the gevent version
+if gevent.version_info < (1, 2, 0):
```
tilgovi: Version 1.2.0 is when subprocess support really stabilized and compatibility for Python 3.5 and 3.6 was added.[1](http://www.gevent.org/whatsnew_1_2.html).

tilgovi:
@jamadden your 👀 would be appreciated here.

gunicorn/workers/ggevent.py
jamadden: Version 1.2 was also when `gevent.version_info` was deprecated 😄 (It doesn't always get updated because it's not a process automated by zest.releaser.) The suggested alternative is  to use `pkg_resources.parse_version(gevent.__version__)` (or the equivalent `packaging.version.Version(gevent.__version__)`. You could also query the `pkg_resources` working set.

gunicorn/workers/ggevent.py
```diff
@@ -51,22 +50,15 @@ class GeventWorker(AsyncWorker):
...
+        monkey.patch_all()
 
         # monkey patch sendfile to make it none blocking
         patch_sendfile()
 
         # patch sockets
         sockets = []
         for s in self.sockets:
-            sockets.append(socket(s.FAMILY, _socket.SOCK_STREAM,
+            sockets.append(socket(s.FAMILY, socket.SOCK_STREAM,
```
jamadden: Pretty sure this will fail with `AttributeError`, because above there is `from gevent.socket import socket`, so `socket` refers to a class, not a module. Either `from gevent import socket` and then `socket.socket` as needed, or `from gevent.socket import SOCK_STREAM`?

gunicorn/workers/ggevent.py
jamadden: Version 1.4 is the first to fully support Python 3.7.

Personally, I don't bother much with 1.2 anymore. The only thing it has going for it over 1.3 is "official" support for Python 2.7 < 2.7.9. gevent 1.2 didn't support Python 2.6.

If I understand correctly, `master` of gunicorn is now Python 3 only, so I don't really see a reason not to require the latest gevent, gevent 1.4.

Other than the minor comments, this seems fine to me.

tilgovi: Have you encountered any incompatibilities in the wild with third party libraries that would prevent users from upgrading to 1.4?

tilgovi:
Okay. I'm convinced to make Gevent 1.4 the minimum version 😄. Should we be aggressive with Eventlet, too? If anyone knows a project that makes heavy use of Eventlet and Gunicorn, please let them know so they can weigh in here.

gunicorn/workers/ggevent.py
jamadden: Is there a chance you could use one of the other options instead of accessing `gevent.version_info`, since that is deprecated?

jamadden: Thank you!

benoitc:
@tilgovi i think we can be agressive for the 0.20, we already are by removing any python 2 support.

benoitc:
@berkerpeksag indeed!

benoitc:
> Should we update the version requirements at http://docs.gunicorn.org/en/latest/settings.html#worker-class as well? 

that's require a change in config.py. Once done let's merge the change :)

tilgovi:
Suggestions on minimum Eventlet versions? I'm tempted to go all the way to 0.24.0.

benoitc:
@tilgovi i'm fine with using latest stable version of eventlet, so yes :)

PR #1967, opened 2019-01-23
Fix #1965: About gunicorn [CRITICAL] Worker Timeout

### LINKED ISSUE #1965: About gunicorn [CRITICAL] Worker Timeout
Recently, I found that my worker appears timeout at about 9 o 'clock every day.

Finally, one day I found that when the system time was modified (more than the current time), the worker unexpectedly appeared timeout! At this time, I can confirm that the daily worker timeout is caused by time synchronization.

I checked the relevant source code and found that the system time was used to determine whether the worker was timeout.

https://github.com/benoitc/gunicorn/blob/7af6f651c04511e631cbb888f527897d4273703c/gunicorn/arbiter.py#L497

skytoup:
Fixes #1965 

Refer to your suggestions, use `time.monotonic` and `os.utime` to fix system time changes cause the woker timeout


gunicorn/workers/workertmp.py
```diff
@@ -35,14 +36,12 @@ def __init__(self, cfg):
             os.close(fd)
             raise
 
-        self.spinner = 0
-
     def notify(self):
-        self.spinner = (self.spinner + 1) % 2
-        os.fchmod(self._tmp.fileno(), self.spinner)
+        new_time = time.monotonic()
+        os.utime(self._tmp.name, (new_time, new_time))
```
tilgovi: Gunicorn unlinks the temporary file as soon as its created, so that it doesn't exist in the directory listing but is still open for the worker. I think you need to use `fileno()` here, still, which is why I double-checked that `utime()` supports a file descriptor.

tilgovi:
Thanks! One small change and it should be great.

tilgovi:
It would also be great to know if this solves your original issue!

benoitc:
@skytoup does it solve your issue?

gunicorn/workers/workertmp.py
```diff
@@ -35,14 +36,12 @@ def __init__(self, cfg):
...
 
     def last_update(self):
-        return os.fstat(self._tmp.fileno()).st_ctime
+        return os.fstat(self._tmp.fileno()).st_mtime
```
benoitc: why the change? `st_ctime` is platform dependent where `st_mtime` will always tell the most recent content modification expressed in seconds.

benoitc: https://docs.python.org/3/library/os.html?highlight=st_ctime#os.stat_result.st_ctime

skytoup: In fact, my idea is to use the member variable to store the time without modifying the file

benoitc: @skytoup  what do you mean? `os.fstat` is atomic, using the modification time is actually very safe so i'm not sure we would like to change it. Such change can be part of another ticket but let's focus on making the time monotonic imo.

tilgovi: This object is shared between the master process and the child process at fork, both with access to the file descriptor of an open file. After the fork, the memory is not shared. That's why we use the filesystem modification time—to signal between processes.

tilgovi: And we need ctime instead of mtime because we only change the metadata (permissions), and not the content.

benoitc: `ctime` is not portable this is why I was using `mtime`. I would prefer to keep it that way.

tilgovi: I'm confused, @benoitc. This change is from `ctime` to `mtime` and you're asking "why the change" but then you say that you prefer `mtime`?

tilgovi: Even if `ctime` is platform dependent, I think we _have_ to use it because we don't modify the file content in any way. We only modify the permissions.

tilgovi: I'm glad you caught the change. I missed it! @skytoup would you mind changing this back to `ctime`, please?

tilgovi: Oh, wait! I am wrong. I got confused.

We are changing this to use utime instead of fchmod, so it is the modified time that changes. In that case, I think the PR is right.

benoitc:
@skytoup we are doing a new release this week and it would be useful to include such change now, have you any questions about the proposed changes?

skytoup:
I found the reason. Disk error due to system backup and affected gunicorn to modify the file information.

tilgovi:
Do we still want to merge this PR? It makes sense if it works and is portable.

benoitc:
@tilgovi  the idea is still interresting. Provided we replace `os.utime` by  `os.fchmod` since we need to work on the file descriptor.

tilgovi:
@benoitc `os.utime` accepts a file descriptor.

gunicorn/workers/workertmp.py
tilgovi:
```suggestion
        os.utime(self._tmp.fileno(), (new_time, new_time))
```

chaitu-tk:
Hi, 

I am facing the same issue, any plans when/if this will be merged? We are running a service during boot and time sync is causing premature/false timeouts.

chaitu-tk:
I can confirm that the fix works on 18.04, the test case which was failing always passed. It would be good to get this merged.

tilgovi:
@benoitc can you take another look at this please? I think it's correct. The change to using `utime` instead of `fchmod` and checking `st_mtime` instead of `st_ctime` is fine and maybe more portable to Windows filesystems in the future. If the resolution concerns us, we can use the `ns` option of `utime` and check `st_mtime_ns` (though many filesystems won't be that accurate).

kylemanna:
Ran into this issue on an embedded device that doesn't have a RTC so the time is always wrong on a start-up and jumps when it's connected to the Internet.  Applied this patch and the issue seems resolved.

Any updates on getting this merged, would love to see it incorporated?

otavio (approved):

flixr:
Bumping this, would be great to get this merged to finally resolve this issue.

tilgovi:
Superseded by #3095.

tilgovi:
Actually, #3095 does not use the monotonic trick. Feel free to reopen and rebase and we can merge this to fix the other issues.

tilgovi:
I'll just merge, resolving conflicts. This looks goood.

PR #1969, opened 2019-01-24
handle `wsgi.input_terminated` extension

### LINKED ISSUE #1653: Support wsgi.input_terminated flag
Gunicorn supports chunked transfer encoding, but unfortunately the WSGI spec has no support for it. A standard was proposed by @mitsuhiko here: https://gist.github.com/mitsuhiko/5721547. If the WSGI server supports streams without a content length, it should set `environ['wsgi.input_terminated'] = True`. Other servers and clients are starting to support this.

benoitc:
fix #1653

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -73,6 +73,7 @@ def base_environ(cfg):
         "wsgi.multiprocess": (cfg.workers > 1),
         "wsgi.run_once": False,
         "wsgi.file_wrapper": FileWrapper,
+        "wsgi.input_terminated": True,
         "SERVER_SOFTWARE": SERVER_SOFTWARE,
     }
 
@@ -130,6 +131,7 @@ def create(req, sock, client, server, cfg):
             continue
         elif hdr_name == "CONTENT-LENGTH":
             environ['CONTENT_LENGTH'] = hdr_value
+            environ['wsgi.input_terminated'] = False
             continue
 
         key = 'HTTP_' + hdr_name.replace('-', '_')


benoitc:
this works for me but let me know.

tilgovi:
It will take me some time to understand again what this flag is for and if this is correct. I can do it tonight, though, I think!

benoitc:
@tilgovi btw feel free to ping me either on gitter or irc if you have any questions about it :)

gunicorn/http/wsgi.py
```diff
@@ -130,6 +131,7 @@ def create(req, sock, client, server, cfg):
             continue
         elif hdr_name == "CONTENT-LENGTH":
             environ['CONTENT_LENGTH'] = hdr_value
+            environ['wsgi.input_terminated'] = False
```
tilgovi: Is this necessary? The way I read the gist it seems like Gunicorn could safely always set `wsgi.input_terminated`.

Our `LengthReader` will not allow reading past `Content-Length` bytes, but it will also return zero bytes to signal EOF at that point and it will raise `IOError` if the `recv` fails due to client disconnect.

I think we can just set it always.

benoitc: @tilgovi my initial intention was to keep the default behaviour for it waiting for a proper way to test how the implementation is used on different platforms. But maybe we should though we should try to sponsor it as an update of the wsgi protocol since it would be really the default....

tilgovi: It cannot be the default that applications and frameworks assume `wsgi.input_terminated`, because some servers might not limit the number of bytes that can be read to `Content-Length`.

Gunicorn always does, though, so I think it's easier to think about if we always set it.

- The application cannot read past `Content-Length`, if it exists
- The `Content-Length` header, if it exists, is passed through to the application
- The input stream behaves the same always: `IOError` on client disconnect and EOF on chunk termination or `Content-Length` bytes read.

I cannot see a reason to ever set it to `False`, but I may be missing something.

tilgovi: @mitsuhiko if I may ping you here, does this sound right?

benoitc: @tilgovi i guess your assumption is right. I think at the end we have the choice between:

1. beeing conservative: forcing the client to use the whole PEP3333 when a `Content-Length` header is given
2. always set `wsgi.input_terminated` to **true**. which is a breaking change at least for flask based applications
3. consider an incremental path with an environment variable setting the default for the case where we get a `Content-Length` header


Thoughts?

tilgovi: It is breaking because it changes what the environment says, but it doesn't change behavior, does it? Maybe werkzeug would rely on Gunicorn to do what it was already doing itself?

benoitc: @tilgovi well it changes the behaviour for werkzeug and other applications/frameworks that handle `wsgi.input_terminated` hen they process a request. For gunicorn itself, this won't change much the way we are processing the request indeed.

Anyway I think you convinced me, I will make the change to set `wsgi.input_terminated` to true by default.

tilgovi: What I mean is this.

Before: Werkzeug ensures that the stream is terminated, either on `Content-Length` bytes, last chunk, or through `IOError`.
After: Gunicorn ensures that the stream is terminated, either on `Content-Length` bytes, last chunk, or through `IOError`.

So, it's breaking because Werkzeug will work differently under Gunicorn, but it's _not_ breaking because the for the application the behavior is the same, it's just Gunicorn that provides the behavior rather than Werkzeug. And really, Gunicorn was already providing this behavior, just not advertising it with `wsgi.input_terminated`, right?

benoitc:
@tilgovi do you think we need the change still? If yes, I will apply it tomorrow so we can release this week. Thoughts?

tilgovi:
I still think there's no reason for us to set it to `False` since we always ensure our input object is terminated. https://github.com/benoitc/gunicorn/pull/1969#discussion_r251217124

If you have any hesitation about that, maybe leave a comment on the line reflecting my thoughts? Something like:

`# May not be necessary to set this to False. Gunicorn always provides a reader that terminates.`

benoitc:
@tilgovi let's do the change as you suggested, it should be fine.

gunicorn/http/wsgi.py
tilgovi: We can ship this now as is. If we want to change it to always set `wsgi.input_terminated` to `True`, we have one more task that we can do in a separate ticket. We have to raise `IOError` if we detect a disconnect before `CONTENT_LENGTH` bytes.

tilgovi:
I think we can merge this and follow up with better disconnect handling for the `LengthReader`. But first, I might take a look at simplifying our streams and use the `io` base classes better.

berkerpeksag:
> I might take a look at simplifying our streams and use the `io` base classes better.

Do you have a plan/document or WIP branch for this? I think using the APı from io module would be great.

benoitc:
i would like to merge this and make a release. Thoughts?

berkerpeksag:
Release 19.10 or 20.0? I'd like to get https://github.com/benoitc/gunicorn/pull/2012 in and some other PRs for 20.0 first. I will have a four days holiday this week, so I may able to get some PRs merged :)

berkerpeksag:
Also, we may want to release 20.0 as `--pre` release first to gather some feedback.

benoitc:
@berkerpeksag i would go for a 20.0.0 and increment it over the time. +1 for #2012. let me know when it's done :)

tilgovi:
>> I might take a look at simplifying our streams and use the io base classes better.

> Do you have a plan/document or WIP branch for this? I think using the APı from io module would be great.

I had looked at it once a while back. I started looking at it again yesterday and I'll share WIP in the next few days. So far, using `socket.makefile()` to provide buffered I/O on a socket with a file-like interface seems like it will make some of our `Reader` and `Unreader` classes mostly redundant. By letting buffering happen at a lower level and using `readline()` in some of the parsing code (request line, request headers, chunk header, chunk trailers) we should no longer need the `Unreader.unread()` method and we don't have to track and pass around as many `BytesIO` objects.

benoitc:
@tilgovi can you open a ticket for it? in between i will merge this one so we can make a release as soon  as #2012 is merged.

PR #1971, opened 2019-01-24
remove gaiohttp worker

### LINKED ISSUE #1590: Consistently use Python's warnings module and remove `utils.warn`

benoitc:
worker is deprecated and won't work on latest version.

fix #1590


docs/source/design.rst
```diff
@@ -59,32 +59,16 @@ WSGI application, this is not a recommended configuration.
 AsyncIO Workers
 ---------------
 
-These workers are compatible with python3. You have two kind of workers.
+These workers are compatible with python3.
```
tilgovi: nit: Python 3

PR #1992, opened 2019-02-20
Move ThreadPoolExecutor() creation into own method

awiddersheim:
Move ThreadPoolExecutor() creation into it's own method so it is easier
to override when subclassing.


awiddersheim:
Fixes #1992.

tilgovi:
Thanks, @awiddersheim. I'll give others a chance to weigh in, but I think this is nice and harmless and unblocks you to experiment.

awiddersheim:
@tilgovi @berkerpeksag thanks for the feedback. Updated per comments.

tilgovi:
Looks great. Thanks!

PR #2024, opened 2019-04-30
Emphasize that full greenlet support might need additional work

aberres:
When I read about the async worker the first time I had no idea about greenlets and stuff.
I think it would have helped me to get some additional context as the async worker is obviously no silver bullet.

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -46,6 +46,13 @@ Gevent_). Greenlets are an implementation of cooperative multi-threading for
 Python. In general, an application should be able to make use of these worker
 classes with no changes.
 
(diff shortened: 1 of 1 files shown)


benoitc:
we can add an expectation / workers imo

tilgovi:
It's okay for me. I might take out context variables. I'm just thinking about what might be most likely to become out of date. Making specific recommendations is tricky.

aberres:
I agree that the documentation might benefit from some additional context and maybe examples. At least for the moment I do not have the time to add anything else though.

I agree about not referencing context variables. I removed the clause.

tilgovi:
This is great. Thank you for helping make the documentation better!

PR #2041, opened 2019-05-20
Proposal: Use threading.Semaphore instead of nr_conns counter.

c-bata:
Hi maintainers! Thanks for maintaining this awesome project. This project helps me a lot.

For my study purpose, I read the implementation of this. Then I thought it's able to refactor the implementation of ThreadWorker using threading.Semaphore. May I have your opinions about this refactor changes?

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -17,6 +17,7 @@ import selectors
 import socket
 import ssl
 import sys
+import threading
 import time
 from collections import deque
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


benoitc:
whhy do you want to use a semaphore though ? What would be the advantage about using the counter there?

I think our current usage of the counter issafe there. Otherwise i would prefer to go using an atomic integer instead of a semaphore. but let us know.

c-bata:
As you said, it might be enough to use atomic integer. I couldn't show the great advantage.

PR #2042, opened 2019-05-20
Fix typo: connnection to connection

c-bata:


tilgovi:
Thank you for all of these!

PR #2043, opened 2019-05-20
Avoid using __import__ and eval

c-bata:
See https://docs.python.org/3/library/functions.html#__import__

```
$ python -V
Python 3.7.2
$ gunicorn examples.echo:app
[2019-05-20 23:09:16 +0900] [26697] [INFO] Starting gunicorn 19.9.0
[2019-05-20 23:09:16 +0900] [26697] [INFO] Listening at: http://127.0.0.1:8000 (26697)
[2019-05-20 23:09:16 +0900] [26697] [INFO] Using worker: sync
[2019-05-20 23:09:16 +0900] [26700] [INFO] Booting worker with pid: 26700

```

PR #2046, opened 2019-05-26
Remove execfile_, use SourceFileLoader instead

c-bata:
We can use importlib because gunicorn drops Python2 support. In this PR, I remove `gunicorn/_compat.py` and use importlib.SourceFileLoader instead.

Notice: To import wsgi application, we've already used `importlib` (See https://github.com/benoitc/gunicorn/pull/2043 for details).


gunicorn/app/base.py
```diff
@@ -93,22 +95,18 @@ def get_config_from_filename(self, filename):
...
+            loader = SourceFileLoader(module_name, filename)
+            loader.exec_module(mod)
         except Exception:
             print("Failed to read config file: %s" % filename, file=sys.stderr)
             traceback.print_exc()
             sys.stderr.flush()
             sys.exit(1)
 
-        return cfg
+        return mod.__dict__
```
berkerpeksag: Can't we use `vars(mod)` here?

gunicorn/app/base.py
```diff
@@ -5,13 +5,15 @@
 import os
 import sys
 import traceback
+import types
+from importlib.machinery import SourceFileLoader
```
berkerpeksag: Nit: Can you import it as `import importlib.machinery` and alphabetize all imports?

c-bata:
Thanks for reviewing. I've just pushed a commit to apply your feedbacks.

berkerpeksag:
Thanks!

PR #2047, opened 2019-05-27
Remove util.import_module

c-bata:
Remove `util.import_module`. I think this is the last cleanup around module importing functionalities.

refs #2046 and #2043


berkerpeksag:
Thanks!

PR #2049, opened 2019-05-27
Remove SimpleNamespace class

c-bata:
types.SimpleNamespace is new in version 3.3.
https://docs.python.org/3/library/types.html#types.SimpleNamespace


berkerpeksag:
Thanks!

PR #2053, opened 2019-05-29
Updated systemd documentation to link the Socket with the Service

budgester:
When restarting the service the socket doesn't always get recreated, by linking the socket with the service, the socket file is always available to nginx.

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -252,6 +252,9 @@ to the newly created unix socket:
 
     [Unit]
     Description=gunicorn socket
+    # Restart the socket as PartOf the service so we never lose the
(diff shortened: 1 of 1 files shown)


tilgovi:
I don't think I understand why this is necessary or why it solves the problem described in the comment, but I am not very well versed in systemd. Why do we need to restart the socket when gunicorn restarts? Isn't the point of socket activation that the socket can exist even without the service?

budgester:
Yup, your right, I have my socket and pid file in a subdirectory of /run e.g. /run/myapp/app.pid & app.socket, the behaviour I was seeing was on a service restart the sub-dir was getting recreated and losing the socket file, and hence nginx wasn't able to find it anymore.

Issue #2075, opened 2019-07-10
Define a __repr__ for gunicorn.config.Setting descendants


aawilson:
I did search old issues for this, so if it's already been accepted and rejected, please direct all blame at Github.

Sometimes it's helpful for debugging purposes to be able to just dump all the settings at a go. Currently, all Settings subclasses use the default object `__repr__` (which looks like `<[module.classname] object at [memory location]>`), which isn't particularly useful. I propose that it either prepend self.value to the default (which has the advantage of showing that it is a gunicorn.config.Settings type), or that it just use the repr of self.value (which has the advantage of showing what you probably actually care about).

benoitc:
merged. Thanks!

PR #2076, opened 2019-07-10
Add a __repr__ to config to include its value

aawilson:
It's sometimes helpful to be able to trivially dump all the config values
for debugging purposes. This commit defines a repr for that. See #2075

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -308,6 +308,15 @@ class Setting(object):
                 self.order < other.order)
     __cmp__ = __lt__
 
+    def __repr__(self):
+        return "<%s.%s object at %x with value %r>" % (
+            self.__class__.__module__,
+            self.__class__.__name__,
+            id(self),
+            self.value,
+        )
+
+
 Setting = SettingMeta('Setting', (Setting,), {})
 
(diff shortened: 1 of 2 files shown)


aawilson:
The test is basically "does this not explode", so if anyone thinks of something better to exercise, let me know. The functionality is pretty trivial, so I don't really expect much there.

aawilson:
This test failure is bizarre:

https://github.com/benoitc/gunicorn/pull/2076

```
$ python3 --version
Python 3.5.2
$ python3
>>> from gunicorn import util
>>> util.parse_address('127.0.0.1:test')
  File "<path>/util.py", line 238, in parse_address
    port = int(port)
ValueError: invalid literal for int() with base 10: 'test'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "<path>/util.py", line 240, in parse_address
    raise RuntimeError("%r is not a valid port number." % port)
RuntimeError: 'test' is not a valid port number.
```
>>>

aawilson:
Happy to ignore them. I can't merge, though, so I don't have much choice ^.^

PR #2077, opened 2019-07-11
Logging: Handle auth type case insensitively

MartinBasti:
According RFC-7617 (inherited from RFC-2978) schema and parameter names are handled
case insensitively:
```
Note that both scheme and parameter names are matched case-
insensitively.
```

Signed-off-by: Martin Bašti <mbasti@redhat.com>

### DIFF
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -29,15 +29,15 @@ def test_parse_address(test_input, expected):
 
 
 def test_parse_address_invalid():
-    with pytest.raises(RuntimeError) as err:
+    with pytest.raises(RuntimeError) as exc_info:
         util.parse_address('127.0.0.1:test')
-    assert "'test' is not a valid port number." in str(err)
+    assert "'test' is not a valid port number." in str(exc_info.value)
 
 
 def test_parse_fd_invalid():
-    with pytest.raises(RuntimeError) as err:
+    with pytest.raises(RuntimeError) as exc_info:
         util.parse_address('fd://asd')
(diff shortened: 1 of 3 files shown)


tests/test_util.py
```diff
@@ -29,15 +29,15 @@ def test_parse_address(test_input, expected):
 
 
 def test_parse_address_invalid():
-    with pytest.raises(RuntimeError) as err:
+    with pytest.raises(RuntimeError) as exc_info:
         util.parse_address('127.0.0.1:test')
-    assert "'test' is not a valid port number." in str(err)
+    assert "'test' is not a valid port number." in str(exc_info.value)
```
berkerpeksag: `str()` shouldn't be needed when we use `ExceptionInfo.value` as it already contains a string (`ExceptionInfo` is basically a wrapper around `sys.exc_info()`:

```py
class ExceptionInfo:
    """ wraps sys.exc_info() objects and offers
        help for navigating the traceback.
    """

    ...

    @property
    def value(self):
        """the exception value"""
        return self._excinfo[1]
```

MartinBasti: Fails without `str()`, AFAIK .value contains exception instance
```
>       assert 'No module' in exc_info.value
E       TypeError: argument of type 'ModuleNotFoundError' is not iterable
```

berkerpeksag: Good point!

MartinBasti:
Is there anything I can do to get this PR merged?

berkerpeksag:
Thanks!

MartinBasti:
Thank you!

PR #2081, opened 2019-07-30
refactor to catch multiple execption

jun0jang:
removing duplicated try catch code :)


tilgovi:
I'll wait to see if anyone has a strong opinion to keep them separate for any reason, but I'm fine with this.

berkerpeksag:
Thanks!

Issue #2080, opened 2019-07-29
Hardcoded six


urishalit:
Hi There,
There is a known [issue](https://github.com/spotify/luigi/issues/1706) when using pinject and six < 1.9.0. When running with gunicorn there is the problem that six.py is under gunicorn package and not as a dependency. I see in the latest version 19.9.0 it still is the case, but in master it isn't the case anymore.

Is a version expected to remove this dependency? Why is it there from the first place and not list it as a dependency?

Basically, I can't use gunicorn & pinject together currently.

urishalit:
I forked the repository and removed the hardcoded dependency of `six` and it seems to work, opened a PR to merge it back.

https://github.com/benoitc/gunicorn/pull/2083

benoitc:
#2083 has been closed, closing the issue since no activity happened since. Feel free to reopen it if needed. Thanks for the PR also :)

PR #2083, opened 2019-08-01
Remove six dependency

urishalit:
I removed the file `gunicorn/six.py` and added `six` as a requirement for the project.

The motivation for this was [this issue](https://github.com/benoitc/gunicorn/issues/2080).

I ran the tests locally and that ones that pass before change pass after it and vice versa, so I am sure the ones that are failing are due to a problematic test configuration (Or you have failing tests merged) - I tested this on our environment and it seems to work.


urishalit:
Yes, the problem is _winreg.
Why isn't this the correct solution? 

Why is it too late?

tilgovi:
> Why is it too late?

We would like to consider 19 in maintenance mode now, with a focus on releasing 20. Changes should be kept to a minimum to fix bugs. Updating our vendor copy of `six` is fine, but adding a dependency is more risky.

Gunicorn has `six` as a vendored dependency so that Gunicorn does not restrict what version of `six` applications can use and is not affected by any version of `six` used by an application running under Gunicorn. I believe we should keep it that way.

I think we could accept a PR to update the vendored version of `six`, assuming there are no changes that would case any fear of breaking something in Gunicorn.

berkerpeksag:
What @tilgovi has said. I'd be happy to review and merge a PR that moves the _winreg import to the correct place.

PR #2090, opened 2019-08-14
Make http request body file-like

lwzm:
When use request body in stream text mode, it should have these methods:

* readable
* writable
* seekable
* closed

ref: https://docs.python.org/3/library/io.html#io.IOBase

### DIFF
--- a/gunicorn/http/body.py
+++ b/gunicorn/http/body.py
@@ -174,10 +174,11 @@ class EOFReader(object):
         return ret
 
 
-class Body(object):
+class Body(io.IOBase):
     def __init__(self, reader):
         self.reader = reader
         self.buf = io.BytesIO()
+        super().__init__()
 
     def __iter__(self):
         return self
@@ -188,6 +189,9 @@ class Body(object):
             raise StopIteration()
         return ret
     next = __next__
+
+    def readable(self):
+        return True
 
     def getsize(self, size):
         if size is None:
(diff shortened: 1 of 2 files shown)


lwzm:
My test app:
```
import io

def application(environ, start_response):
    status = '200 OK'
    headers = [('Content-type', 'text/plain; charset=utf-8')]
    src = io.TextIOWrapper(environ['wsgi.input'])
    for line in src:
        print(repr(line))
    start_response(status, headers)
    return []
```

lwzm:
another test app:
```
import io

import falcon


class Patch:
    def on_post(self, req, resp, target):
        src = io.TextIOWrapper(req.stream)
        for line in src:
            print(repr(line))


application = falcon.API()

application.add_route('/{target}', Patch())

```

and client:
```
curl --data-binary @x.txt localhost:8000/foo
```

gunicorn/http/body.py
```diff
@@ -174,10 +174,11 @@ def read(self, size):
         return ret
 
 
-class Body(object):
+class Body(io.IOBase):
     def __init__(self, reader):
         self.reader = reader
         self.buf = io.BytesIO()
+        super().__init__()
```
tilgovi: Does this make sense? `IOBase` has no constructor.

lwzm: eh... I don't known how to avoid pylint warning.

lwzm:
This is a "quick fix", don't mind.
I'll attention your progress.

gunicorn/http/body.py
berkerpeksag: We can ignore that pylint warning since it doesn't make any sense in most cases.

Issue #2085, opened 2019-08-02
setproctitle isn't in extra_requires


jwhitlock:
The [FAQ says](https://docs.gunicorn.org/en/latest/faq.html#how-can-i-name-processes):

> If you install the Python package setproctitle Gunicorn will set the process names to something a bit more meaningful. This will affect the output you see in tools like ``ps`` and ``top``. This helps for distinguishing the master process as well as between masters when running more than one app on a single machine. See the proc_name setting for more information.

If this was added to ``extras_require`` in ``setup.py``, then ``pip install gunicorn[gevent,setproctitle]`` or similar could pick up this optional dependency, and advertise it a little more: 

```
extras_require = {
    'gevent':  ['gevent>=0.13'],
    'eventlet': ['eventlet>=0.9.7'],
    'tornado': ['tornado>=0.2'],
    'gthread': [],
    'setproctitle': ['setproctitle'],
}
```

tilgovi:
A PR would be most welcome. Thanks!

berkerpeksag:
PR #2094 has been merged. Closing this as 'fixed'. Thank you for the report and for the PR!

PR #2094, opened 2019-08-18
Add setproctitle to extras_require

jwhitlock:
This allows you to specify that you want ``setproctitle`` installed so that ``gunicorn`` can set meaningful process names at install time (``pip install gunicorn[gevent,setproctitle]``) or in a requirements file (``gunicorn[gevent,setproctitle]==19.9.1``). Fixes issue #2085.


jwhitlock:
I added a section to the Installation doc on all the ``[extras]`` options, and added a fix in the Custom Application document for a Sphinx warning.

PR #2105, opened 2019-09-06
Test under PyPy3

johnthagen:
One of the advantages `gunicorn` has is PyPy3 support (see uWSGI issue https://github.com/unbit/uwsgi/issues/869). This PR adds CI testing for PyPy3 so users can be confident that support does not regress.


berkerpeksag:
Thanks!

Issue #1140, opened 2015-11-02
Switch to use SSLContext


berkerpeksag:
We need to use SSLContext in Python 3 (and perhaps 2.7.9+ too) to make Gunicorn users like easier. The old API is buggy and causes weird edge case bugs like reported in #1135.

benoitc:
@berkerpeksag wanna want to do it? :)

berkerpeksag:
Will do, but I'm busy with job interviews these days so it can take a while :)

tilgovi:
Here's a diff that begins this work for anyone who wants to get started on this:

    diff --git a/gunicorn/workers/sync.py b/gunicorn/workers/sync.py
    index 15ac084..5fd5730 100644
    --- a/gunicorn/workers/sync.py
    +++ b/gunicorn/workers/sync.py
    @@ -122,8 +122,16 @@ class SyncWorker(base.Worker):
             req = None
             try:
                 if self.cfg.is_ssl:
    -                client = ssl.wrap_socket(client, server_side=True,
    -                    **self.cfg.ssl_options)
    +                context = ssl.SSLContext(self.cfg.ssl_options['ssl_version'])
    +                context.set_default_verify_paths()
    +                #context.load_verify_locations(self.cfg.ssl_options['ca_certs'])
    +                context.load_cert_chain(
    +                    self.cfg.ssl_options['certfile'],
    +                    keyfile=self.cfg.ssl_options['keyfile'])
    +                context.set_ciphers(self.cfg.ssl_options['ciphers'])
    +                context.options |= ssl.OP_CIPHER_SERVER_PREFERENCE
    +                context.verify_mode = self.cfg.ssl_options['cert_reqs']
    +                client = context.wrap_socket(client, server_side=True)
     
                 parser = http.RequestParser(self.cfg, client)
                 req = six.next(parser)

The `ssl.OP_CIPHER_SERVER_PREFERENCE` should probably be made a configuration option. I'm also not sure we _should_ set the default verify paths or not and I have commented out the `load_verify_locations` call because I didn't fully digest the parameters here and wanted to be sure I was loading the CA certificates for client authentication. This was a rush, rough draft I did one evening last weekend just to start getting familiar with the `ssl` module.

We can also probably do this somewhere else, maybe in `config.py`, so that the `SSLContext` setup code can be shared by all the workers.

benoitc:
SSLContext is available to python 2.7:

[...]

PR #2118, opened 2019-09-26
3Reload listeners if ssl configs have changed

TylerLubeck:
Hi! 

I'm doing some work with certificate rotation, and it would be useful to be able to reload gunicorn when ssl configurations have changed.

The more interesting part is reloading gunicorn if the contents of the various certificate files have changed. This allows us to automate the process of certificate rotation - we have a separate process that creates new certificate bundles and puts them in a standard location and then signals other applications to reload them.

Let me know what you think?

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -430,6 +430,8 @@ class Arbiter(object):
 
     def reload(self):
         old_address = self.cfg.address
+        old_ssl_options = self.cfg.ssl_options
+        old_certificate_mtimes = self.cfg.certificate_mtimes
 
         # reset old environment
         for k in self.cfg.env:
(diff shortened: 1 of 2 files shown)


tilgovi:
Does Python cache the `certfile` argument? Gunicorn calls the deprecated `wrap_socket()` method (there is issue #1140 about switching to `SSLContext`) with the certfile and keyfile arguments when a client connects. At the very least, I would expect spawning new workers to use a new certificate even if each worker caches the certificate by filename between client accepts.

sindhuravi-github:
Hi @TylerLubeck just curious, is there any update on this or any alternatives to reloading the SSLContext for usecases like automating cert rotation?

Issue #1466, opened 2017-02-19
Serving http+https


binrush:
Hello. gunicorn supports both listening on multiple ports and serving https. My question is, if it possible to:

1. Serve the same app on both http and https in single gunicorn instance?
2. Use custom `SSLContext` for https?

Seems like it should be possible with Custom applicaton, but I really can not figure the way out from the docs.

stefaang:
I'm using nginx the load balancer to achieve your first bullet, with a redirect from http to https.
[Tutorial](https://www.digitalocean.com/community/tutorials/how-to-set-up-django-with-postgres-nginx-and-gunicorn-on-ubuntu-16-04)
But maybe that is not what you're looking for.

benoitc:
such support can be added inside gunicorn. As the moment it's not possible.

Cl1608Ho:
bump. could this be implemented as extension to #444 to also accept `-b http://` or `-b https://` respective, to determine whether a certain socket should work with https or not, if none are used (like legacy support) this should default to "use https if cert and key are given" or however it is handled right now.

dumblob:
The effort in https://github.com/benoitc/gunicorn/pull/2120 seems to currently wait for someone having some experience with gunicorn, but otherwise seems to be a leap forwards.

javabrett:
There's a small overlap here with #1929 where an SSL-request sent to a non-SSL port results in a blocking timeout waiting for the first CR/LF of a HTTP request.

Personally I'm still weighing the merits of this change, but wanted to mention that related issue - presumably if we were detecting HTTP/S that early by peeking at bytes or whatever, #1929 could be solved.

benoitc:
Please join #2237 to discuss the API for listening on multiple interfaces. I think once done it should be straightforward to detect which socket is given. to the request.

kennysong:
Serving the same app on both http and https in Gunicorn would be helpful for us.

You can do this with nginx, but it's a bit overkill for our use case (basically ML inference) since we're not exposing a public facing web app. 

You can also do this by starting two Gunicorn masters, but then we need to worry about consistency between the two instances.

benoitc:
it's easy to run 2 gunicorn each listening on one interface. This remove the complexity. but we can add support of it quite easily in gunicorn modulo the need to handle supervisors like systemd that provides their own sockets.

jwaxy:
Helloi I came here from a google search. Is there any plan in adding this feature into gunicorn?

[...]

PR #2120, opened 2019-09-26
Dual http https

TylerLubeck:
This takes a stab at solving #1466. It would be beneficial for some work I'm involved in where we need to convert all our apps to https communication, and would like to avoid adding technologies (nginx in front of the app, running two separate application pools, etc). 

*It doesn't quite work yet*, but I wanted to ask for an earlier look at methodology and implementation.

The idea is that there's a new flag `--bind-no-ssl` where, regardless of other configurations, ssl settings will not be applied. I chose to go this route to minimize compatibility problems.

Some examples:

* `gunicorn --bind-no-ssl 127.0.0.1:8001`: 
    Works the same as `gunicorn --bind 127.0.0.1:8001`

* `gunicorn --bind 127.0.0.1:8001 --bind-no-ssl 127.0.0.1:8002`: 
    Works the same as `gunicorn --bind 127.0.0.1:8001 --bind 127.0.0.1:8002`

* `gunicorn --certfile ./mycert --keyfile ./mykey --bind 127.0.0.1:8001 --bind-no-ssl 127.0.0.1:8002`: 
    Serves on https://127.0.0.1:8001 and http://127.0.0.1:8002

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -154,7 +154,11 @@ class Arbiter(object):
 
             self.LISTENERS = sock.create_sockets(self.cfg, self.log, fds)
 
-        listeners_str = ",".join([str(l) for l in self.LISTENERS])
+        l_strs = []
+        for l, is_ssl in self.LISTENERS:
+            l_strs.append('%s:%s' % ('ssl' if is_ssl else 'plaintext', str(l)))
+
+        listeners_str = ",".join(l_strs)
         self.log.debug("Arbiter booted")
         self.log.info("Listening at: %s (%s)", listeners_str, self.pid)
         self.log.info("Using worker: %s", self.cfg.worker_class_str)
@@ -379,7 +383,7 @@ class Arbiter(object):
             and not self.systemd
             and not self.cfg.reuse_port
         )
-        sock.close_sockets(self.LISTENERS, unlink)
+        sock.close_sockets([l for l, _ in self.LISTENERS], unlink)
 
         self.LISTENERS = []
         sig = signal.SIGTERM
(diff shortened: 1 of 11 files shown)


apollo13:
@TylerLubeck Is it important for you that you'd bind to different ports? Otherwise one could make the socket listen for both protocols. (This would work since the client always sends bytes first and SSL should be easy enough to identify, plenty of servers do that and then tell you "tried http on ssl port")

TylerLubeck:
Honestly I didn't know you could do that. When you say "SSL should be easy enough to identify", I assume that means with code in gunicorn? 

That implementation would work for me, but I'm not sure how to proceed with it.

tilgovi:
I'm not sure there's a common use case for optional SSL on the same port. I think whatever approach we take should preserve secure defaults where providing certificates to Gunicorn enforces SSL on the socket. I sort of like how explicit `--bind-no-ssl` is!

apollo13:
On Wed, Nov 13, 2019, at 01:13, Tyler Lubeck wrote:
> When you say "SSL should be 
> easy enough to identify", I assume that means with code in gunicorn?

Yes, but I just checked and it seems to be everything but easy to implement in Python 🙄

dumblob:
Any progress on this front? I think I have a valid use case for having optional SSL on the same port (in my case the Flask application shall decide based on the app's internal state). Currently I have to spawn two instances of gunicorn (one with certificates, i.e. forced SSL and one without any certificates, i.e. plain HTTP) - each having it's own port which has some downsides. Particularly the reason is, that the Flask app provides user interface and an API. For API SSL is always required, but the user interface can accommodate (i.e. show just less sensitive information) for HTTP if SSL wasn't requested by the client.

benoitc:
That's a good idea! Though the patch may be  over complicated. I would try for example to just wrap the `gunicorn.Socket` object to add an `is_ssl` method and check on accept. 

Anyway  I would like to discuss the API before going further I created #2237 for it. Please join the discussion :)

benoitc:
Closing due to inactivity. This draft PR has been open for over 4 years and discussion was redirected to issue #2237. If this feature is still needed, please feel free to open a fresh PR against current master. Thank you for your contribution!

PR #2128, opened 2019-10-10
Ensure header value is string before conducting regex search.

jeffgamedev:
When integrating with WebApp2 3.0.0b1 (latest, Python 3) and WebOb 1.8.5 (latest), header value was being passed as byte string and resulted in the exception: "TypeError: cannot use a string pattern on a bytes-like object". Regex search fails due to this value. Current gunicorn implementation parses the value as string, but only after regex search. Changing the casting of value to string before regex search ensures this exception will not occur. This change fixes integration with these libraries.

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -253,7 +253,8 @@ class Response(object):
             if HEADER_RE.search(name):
                 raise InvalidHeaderName('%r' % name)
 
-            value = str(value)
+            if not isinstance(value, str):
+                raise TypeError('%r is not a string' % value)
 
             if HEADER_VALUE_RE.search(value):
                 raise InvalidHeader('%r' % value)
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -253,10 +253,12 @@ class Response(object):
             if HEADER_RE.search(name):
                 raise InvalidHeaderName('%r' % name)
 
+            value = str(value)
+
             if HEADER_VALUE_RE.search(value):
                 raise InvalidHeader('%r' % value)
 
-            value = str(value).strip()
+            value = value.strip()
             lname = name.lower().strip()
             if lname == "content-length":
                 self.response_length = int(value)


tilgovi:
Even though it might be nice to be lenient, I think maybe we can leave this as is. PEP 3333 is pretty unambiguous about headers being `str` type: https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types

jeffgamedev:
> Even though it might be nice to be lenient, I think maybe we can leave this as is. PEP 3333 is pretty unambiguous about headers being `str` type: https://www.python.org/dev/peps/pep-3333/#a-note-on-string-types

I agree, but this change _would_ increase the amount of cases the code works for. Whatever you guys feel is best :)

tilgovi:
I'm not sure that I'm opposed, but I would like to see this code get more consistent. We raise, explicitly, for non-`str`-typed header name, but not for values. It doesn't make much sense to decode a bytes instance to a string here with the default encoding if we later encode as `latin-1` when we convert it back to bytes to send.

I think I'd rather keep it strict and simple, but if it would have helped your debugging I think a PR to raise a reasonable error before executing the regex, like we do for header names, would be a fine change.

tilgovi:
In other words, I sort of like this change as an alternative:

``` diff
diff --git a/gunicorn/http/wsgi.py b/gunicorn/http/wsgi.py
index 32e7a2a..3524471 100644
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -253,10 +253,13 @@ class Response(object):
             if HEADER_RE.search(name):
                 raise InvalidHeaderName('%r' % name)
 
+            if not isinstance(value, str):
+                raise TypeError('%r is not a string' % value)
+
             if HEADER_VALUE_RE.search(value):
                 raise InvalidHeader('%r' % value)
 
-            value = str(value).strip()
+            value = value.strip()
             lname = name.lower().strip()
             if lname == "content-length":
                 self.response_length = int(value)
```

jeffgamedev:
> In other words, I sort of like this change as an alternative:
> 
> ```diff
> diff --git a/gunicorn/http/wsgi.py b/gunicorn/http/wsgi.py
> index 32e7a2a..3524471 100644
> --- a/gunicorn/http/wsgi.py
> +++ b/gunicorn/http/wsgi.py
> @@ -253,10 +253,13 @@ class Response(object):
>              if HEADER_RE.search(name):
>                  raise InvalidHeaderName('%r' % name)
>  
> +            if not isinstance(value, str):
> +                raise TypeError('%r is not a string' % value)
> +
>              if HEADER_VALUE_RE.search(value):
>                  raise InvalidHeader('%r' % value)
>  
> -            value = str(value).strip()
> +            value = value.strip()
>              lname = name.lower().strip()
>              if lname == "content-length":
>                  self.response_length = int(value)
> ```

Thanks! I've made these changes to the pull request.

tilgovi:
Thank you, @jeffgamedev!

Issue #2130, opened 2019-10-15
Inotify recursive reload


talham7391:
When reloading use polling, we achieve a recursive reload behaviour (i.e. if we are watching `/src` and `/src/test/index.py` changes then gunicorn will relaunch the worker).

This is not the case when we are using inotify as the reloader. Inotify supports recursive reloading we just need to enable it in gunicorn.

PR #2131, opened 2019-10-15
Fixeds #2130

talham7391:
When gunicorn is configured to reload using polling, it will recognize any changes in nested folders of the directories being watched. When gunicorn is configured to use inotify as the reloader, it does not reload when there are changes in nested folders. 

This PR configures `InotifyReloader` to reload on nested folder changes to match the default polling behaviour.

### DIFF
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -58,7 +58,7 @@ class Reloader(threading.Thread):
 has_inotify = False
 if sys.platform.startswith('linux'):
     try:
-        from inotify.adapters import Inotify
+        from inotify.adapters import InotifyTrees
         import inotify.constants
         has_inotify = True
     except ImportError:
@@ -78,7 +78,7 @@ if has_inotify:
             self.setDaemon(True)
             self._callback = callback
             self._dirs = set()
-            self._watcher = Inotify()
+            self._watcher = InotifyTrees(mask=self.event_mask)
 
             for extra_file in extra_files:
                 self.add_extra_file(extra_file)
@@ -89,7 +89,7 @@ if has_inotify:
             if dirname in self._dirs:
                 return
 
-            self._watcher.add_watch(dirname, mask=self.event_mask)
+            self._watcher.__load_trees([dirname])
             self._dirs.add(dirname)
 
         def get_dirs(self):
@@ -105,7 +105,7 @@ if has_inotify:
             self._dirs = self.get_dirs()
 
             for dirname in self._dirs:
-                self._watcher.add_watch(dirname, mask=self.event_mask)
+                self._watcher.__load_trees(dirname)
 
             for event in self._watcher.event_gen():
                 if event is None:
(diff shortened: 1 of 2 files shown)


gunicorn/reloader.py
```diff
@@ -89,7 +89,7 @@ def add_extra_file(self, filename):
             if dirname in self._dirs:
                 return
 
-            self._watcher.add_watch(dirname, mask=self.event_mask)
+            self._watcher.__load_trees([dirname])
```
connor-tophat: Why is a list `[dirname]` passed here but just `dirname` below?

talham7391: InotifyTrees expects an array of paths

talham7391:
Linting is failing for files I did not touch - what is expected resolution in this case?

gunicorn/reloader.py
connor-tophat: Then why isn't it an array below?

gunicorn/reloader.py
```diff
@@ -105,7 +105,7 @@ def run(self):
             self._dirs = self.get_dirs()
 
             for dirname in self._dirs:
-                self._watcher.add_watch(dirname, mask=self.event_mask)
+                self._watcher.__load_trees(dirname)
```
connor-tophat: Don't we need `self.event_mask` any more?

gunicorn/reloader.py
```diff
@@ -105,7 +105,7 @@ def run(self):
             self._dirs = self.get_dirs()
 
             for dirname in self._dirs:
-                self._watcher.add_watch(dirname, mask=self.event_mask)
```
tilgovi: It doesn't feel great calling a method that looks private. I noticed that `InotifyTrees` takes a list of paths in its constructor. Can we defer instantiation until this `run` method and pass `self._dirs` to the constructor, or do we need `_watcher` to exist earlier for some reason?

gunicorn/reloader.py
talham7391: Yes - this slipped past - I will push a commit

gunicorn/reloader.py
talham7391: If we want to support `add_extra_file` then we will have to call  `__load_trees` after instantiation. So basically even if I defer to the constructor in the `run` method at some point we will still need to call `__load_trees`.

We can avoid calling this method if we copy paste its logic but I think in this case it should be fine.

talham7391:
@connor-tophat latest commit addresses your comment

gunicorn/reloader.py
tilgovi: I'm not sure there's any reason we don't defer calling `self.reloader.start()` in `gunicorn.workers.base` until after `self.load_wsgi()`. If we defer starting the reloader we can maybe assume that all files are added by the time we call `self.reloader.start()`.

I wonder whether the use case of adding extra files after starting the application is needed.

talham7391: We don't need to reload extra files after startup and your proposed solution would work for us. So I think that is the better way to go.

tilgovi:
@talham7391 I think we should still merge this!

PR #2134, opened 2019-10-16
Start the reloader after the wsgi app has been loaded.

talham7391:
When launching Gunicorn with the `--reload` option, workers will start the Reloader and then load the wsgi app. This is problematic because user files are loaded when the wsgi app is loaded, and the reloader depends on the user files to be loaded so that it can watch them.

So when the Reloader starts, it isn't watching any user files. This isn't an issue when using the default polling reloader as it updates what files it is watching every time it polls but this is an issue with the inotify reloader.

This PR makes sure the reloader is started after wsgi app has been loaded so that the InotifyReloader is able to pickup and watch the proper files/folders.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -163,6 +163,7 @@ Stephen DiCato <Locker537@gmail.com>
 Stephen Holsapple <sholsapp@gmail.com>
 Steven Cummings <estebistec@gmail.com>
 Sébastien Fievet <zyegfryed@gmail.com>
+Talha Malik <talham7391@hotmail.com>
 TedWantsMore <TedWantsMore@gmx.com>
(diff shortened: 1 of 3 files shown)


tilgovi:
Looks good to me. I'll wait a day or so in case others have concerns.

I contributed the original reloader code, taken from meebo/greins. In that project, multiple Gunicorn applications could exist in the same process and they would be loaded by a background thread. In Gunicorn, I'm not aware of any reason that we need locking on the list of files in the reloader, nor any reason to support adding extra files after the reloader starts.

👍 for this. You're welcome to remove the locking in the reloader, too.

talham7391:
@tilgovi Can we merge this?

PR #2144, opened 2019-10-24
[py38] Python 3.8 is now official, use it by default

Djailla:


Djailla:
@tilgovi any update required on this PR ?

PR #2146, opened 2019-10-29
Fix fdopen RuntimeWarning in 3.8

Junnplus:
Fix https://github.com/benoitc/gunicorn/issues/2091

### DIFF
--- a/gunicorn/workers/workertmp.py
+++ b/gunicorn/workers/workertmp.py
@@ -32,7 +32,9 @@ class WorkerTmp(object):
         try:
             if not IS_CYGWIN:
                 util.unlink(name)
-            self._tmp = os.fdopen(fd, 'w+b', 1)
+            # In Python 3.8, open() emits RuntimeWarning if buffering=1 for binary mode.
+            # Because we never write to this file, pass 0 to switch buffering off.
+            self._tmp = os.fdopen(fd, 'w+b', 0)
         except:
             os.close(fd)
             raise


gunicorn/workers/workertmp.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, cfg):
         try:
             if not IS_CYGWIN:
                 util.unlink(name)
-            self._tmp = os.fdopen(fd, 'w+b', 1)
+            self._tmp = os.fdopen(fd, 'w+b', 0)
```
berkerpeksag: It would be nice to add a comment to explain why we pass 0 here as explained in https://github.com/benoitc/gunicorn/issues/2091#issuecomment-522344617.

Junnplus: thank you for you suggestion, i try to add a comment for explain

talhajunaidd:
I'm using gunicorn latest version and python v3.8.5
```
>>> gunicorn --version
gunicorn (version 20.0.4)
>>> python -V
Python 3.8.5
```
But still getting the mentioned warning,
I've tried running the application with flask builtin server to verify that if it's something else, I did not get the warning for flask

saru2020:
In my case, the ports were used and it worked after killing the process running on the port.

Command to kill the process running on the port:
**_sudo lsof -t -i tcp:12000 -s tcp:listen | sudo xargs kill_** 

[Replace the port number with the one from the error]_

jalsaman:
Hi All, 

I am trying to run autopkg and my process if getting stuck at this prompt. /Library/AutoPkg/Python3/Python.framework/Versions/3.10/lib/python3.10/subprocess.py:959: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stdout = io.open(c2pread, 'rb', bufsize)
/Library/AutoPkg/Python3/Python.framework/Versions/3.10/lib/python3.10/subprocess.py:964: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  self.stderr = io.open(errread, 'rb', bufsize)

There is no error but the process isn't completed at all. 

My Primary command was

Issue #2176, opened 2019-11-15
Gunicorn Vulnerable to HTTP Desync / CL-TE Attack


ajones1982:
It seems like gunicorn is vulnerable to a CL-TE attack when used with something like Amazon’s ECS service + a load balancer. It looks like gunicorn parses a malformed transfer-encoding header for this attack.

See this for an example https://nathandavison.com/blog/haproxy-http-request-smuggling

A suggested fix might be to add an option to fail a request if the request has Content-Length and  Transfer-Encoding that is not identity.

snovity:
An alternative approach would be to ignore `Transfer-Encoding` header when Amazon Load Balancer ignores it, i.e. when the header value has a new line before it.

snovity:
Go's fix for this https://github.com/golang/go/commit/6e6f4aaf70c8b1cc81e65a26332aa9409de03ad8. Looks like this has to be handled on `http.server` level

benoitc:
Thanks for the report. I will take care of it ASAP? We have a security mailing list for it why didn't you use it? http://docs.gunicorn.org/en/stable/community.html#security-issues

kovalevvlad:
We are new to contributing to gunicorn. We are working on a fix in parallel. Our approach was to replicate Go team's solution by leaving the header names intact (we dropped rstrip call from message.py:93). We'll try this approach against a vulnerability scanner and report back. Thanks for looking into it.

benoitc:
@kovalevvlad thanks. would be cool to  make a release over the we with a fix

kovalevvlad:
We'll try to send a PR over the weekend (assuming it works)

Sytten:
Any progress? I will start working on it otherwise, because this is a critical vulnerability. I wonder if the new option (`routing.http.drop_invalid_header_fields.enabled`) on AWS ALB might mitigate the issue.

snovity:
@Sytten this is how far we got https://github.com/benoitc/gunicorn/compare/19.x...ryderdamen:19.x Feel free to take over. Will try to test today if this change or `routing.http.drop_invalid_header_fields.enabled` fixes the issue.

Sytten:
This seems like a first step, but we need to test all the variations of the attack. I will enable this field in prod for now and I will work on the fix today. Do we need to backport it to v19 or we only do it for v20 @benoitc?

benoitc:
[...]

PR #2181, opened 2019-11-19
Fix/http desync

Sytten:
This is a first try at fixing https://github.com/benoitc/gunicorn/issues/2176
I have looked at multiple server implementation and frankly I have seen a lot of different things especially concerning the TE parsing.
In summary the fixes are:
- Don't rstrip the names of headers
    - Go server does that
    - RFC 7230 says `No whitespace is allowed between the header field-name and colon`
- Only accept one Content-Length
    - Node does that
    - RFC 7230 says `duplicate Content-Length header fields have been generated or combined by an upstream message processor, then the recipient MUST either reject the message as invalid or replace the duplicated field-values with a single valid Content-Length`
    - It's easier to just deny that case than try to handle it and most likely induce a vulnerability
- Only accept `identity` and `chunked` Transport-Encoding
    - In this implementation, the order does not matter (it probably should). The Go implementation only uses the first value of the header.
    - Seems to be in sync with the behaviour of AWS ALB
    - All other valid (gzip, compress, etc.) and invalid TE will return a 501, since we don't have readers for them I figured this was the right move, but feel free to correct me

Future improvements:
- Handle the duplicates of headers earlier in the parsing
- Use a `Dict[str, List[str]]` instead of a `List[Tuple[str, str]]` for headers
- Drop the chunked TE for WSGI to app, this is what Waitress decided to do [here](https://github.com/Pylons/waitress/blob/e247c93ce932c0184c5242d451e5776b17a20d21/waitress/parser.py#L217) and I think it's a good idea since we already buffer the whole thing in the server anyway

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -12,7 +12,7 @@ from gunicorn.http.unreader import SocketUnreader
 from gunicorn.http.body import ChunkedReader, LengthReader, EOFReader, Body
 from gunicorn.http.errors import (InvalidHeader, InvalidHeaderName, NoMoreData,
     InvalidRequestLine, InvalidRequestMethod, InvalidHTTPVersion,
-    LimitRequestLine, LimitRequestHeaders, UnsupportedTransferEncoding)
+    LimitRequestLine, LimitRequestHeaders)
 from gunicorn.http.errors import InvalidProxyLine, ForbiddenProxyRequest
 from gunicorn.http.errors import InvalidSchemeHeaders
 from gunicorn.util import bytes_to_str, split_request_uri
@@ -135,13 +135,8 @@ class Message(object):
                     raise InvalidHeader("CONTENT-LENGTH", req=self)
                 content_length = value
             elif name == "TRANSFER-ENCODING":
-                normalized_value = value.lower()
-                if normalized_value == "identity":
-                    pass
-                elif normalized_value == "chunked":
+                if value.lower() == "chunked":
                     chunked = True
-                else:
-                    raise UnsupportedTransferEncoding(normalized_value)
             elif name == "SEC-WEBSOCKET-KEY1":
                 content_length = 8
 
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -12,7 +12,7 @@ from gunicorn.http.unreader import SocketUnreader
 from gunicorn.http.body import ChunkedReader, LengthReader, EOFReader, Body
 from gunicorn.http.errors import (InvalidHeader, InvalidHeaderName, NoMoreData,
     InvalidRequestLine, InvalidRequestMethod, InvalidHTTPVersion,
-    LimitRequestLine, LimitRequestHeaders)
+    LimitRequestLine, LimitRequestHeaders, UnsupportedTransferEncoding)
 from gunicorn.http.errors import InvalidProxyLine, ForbiddenProxyRequest
 from gunicorn.http.errors import InvalidSchemeHeaders
 from gunicorn.util import bytes_to_str, split_request_uri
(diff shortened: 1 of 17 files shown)


Sytten:
@benoitc all yours!

gunicorn/http/message.py
```diff
@@ -128,9 +131,17 @@ def set_body_reader(self):
         content_length = None
         for (name, value) in self.headers:
             if name == "CONTENT-LENGTH":
+                if content_length is not None:
+                    raise InvalidHeader("CONTENT-LENGTH", req=self)
                 content_length = value
             elif name == "TRANSFER-ENCODING":
-                chunked = value.lower() == "chunked"
+                normalized_value = value.lower()
+                if normalized_value == "identity":
```
benoitc: Gunicorn is a gateway so it shouldn't have a strong opinion about the transfer encoding. I think it's better to only test if the transfer encoding header is chuned or not.

```python
if value.lower() == "chunked":
  chunked = True
```

and let it pass for other cases

benoitc: This removes the need of `UnsupportedTransferEncoding` also

gunicorn/http/message.py
```diff
@@ -90,7 +90,10 @@ def parse_headers(self, data):
             if curr.find(":") < 0:
                 raise InvalidHeader(curr.strip())
             name, value = curr.split(":", 1)
-            name = name.rstrip(" \t").upper()
+            if self.cfg.strip_header_spaces:
+                name = name.rstrip(" \t").upper()
```
benoitc: shouldn't we return an error if it has "spaces" at the end? We need to check the spec about it.

benoitc:
Thanks for the patch! Seems to handle the case. Can you do th erequested changes? It would be good to make a release today.

gunicorn/http/message.py
Sytten: Well this is the option (disabled by default) to still it if some people really need it. The Normal case (in the else) will not strip the same, the regex after it will fail and yield an error.

gunicorn/http/message.py
Sytten: Your proposal can still be attacked by the following desync:
```
Content-Length: 5
Transfer-Encoding: chunked
Transfer-Encoding: Identity
```
Most proxy will use the chunked, but we will use the CL.
My understanding of Transfer-Encoding is that its a [hop-by-hop header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding), meaning that we should parse it, use it and strip it before passing the data down the line.
For example, AWS ALB will not accept anything that is not identity or chunked. If you read my improvements section above, I suggest that we do something similar to other WSGI server: parse the request to the best of our ablity and only transfer data to the application using a CL. We do buffer the data in this application anyway and we don't support gzip, compress and deflate.

Sytten: reading a bit more about PEP333. It says: `WSGI servers must handle any supported inbound "hop-by-hop" headers on their own, such as by decoding any inbound Transfer-Encoding, including chunked encoding if applicable.` There is a great thread here to read: https://github.com/pallets/flask/issues/367

benoitc: Gunicorn has a long history of supporting chunked encoding and last 20 is having the support of ` wsgi.input_terminated`, i'm not inclined to change this actually.

if I understand your concern, in the case of your example

```
Content-Length: 5
Transfer-Encoding: chunked
Transfer-Encoding: Identity
```

TE should be Identity and not chunked? Or do we want it to be chunked (which is really weird).

 My main concern is to make sure we support the last TE given whatever it is (there are other cases than "Identity" in some apis), so we must not return an error, if it's different.

PR #2188, opened 2019-11-20
fix wsgi.input_terminated

benoitc:
Sometimes both TRANSFER-ENCODING=chunked and CONTENT_LENGTH are set. Since gunicorn prioritise (following the HTTP1.1 spec) chunked encoding we should make sure in this case to signal to th application that the input is terminated by the server.

Without the change gunicorn were always setting wsg.input_terminated to False when a CONTENT_LENGTH header was present ignoring that Gunicorn was afaik handling the termination.

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -33,6 +33,7 @@ class Message(object):
         self.version = None
         self.headers = []
         self.trailers = []
+        self.terminated = True
         self.body = None
         self.scheme = "https" if cfg.is_ssl else "http"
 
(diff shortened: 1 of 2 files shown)


benoitc:
- [ ] add a test

tilgovi:
I still think we should always set `True` because Gunicorn always ensures the reader terminates. It's okay that we don't set it when the content length is known, but it would still be accurate to say that it is terminating.

benoitc:
that's fine for me i will make the change.

benoitc:
i reverted the change and fixed it via 8a270c9329ec3cc8054659117b2ffd613a3d8113 . Now it is always set to true.

tilgovi:
Perfect 👍

PR #2193, opened 2019-11-22
fix attributes when importing config module

### LINKED ISSUE #2182: [ver.20.0.0] The gunicorn config files cannot using '__file__' constants.
I configured app_root that base on the relative path in gunicorn files.

`app_root = os.path.abspath(os.path.join(__file__, '..', '..'))`

now I have occurred when upgrade gunicorn version to 20.0.0 and starting with config file.

`$ venv/bin/gunicorn -c /path/to/app/conf/gunicorn.conf.py wsgi:app`

> Failed to read config file: /path/to/app/conf/gunicorn.conf.py
  File "/path/to/app/venv/lib/python3.7/site-packages/gunicorn/app/base.py", line 102, in get_config_from_filename
    loader.exec_module(mod)
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/path/to/app/conf/gunicorn.conf.py", line 9, in <module>
    app_root = os.path.abspath(os.path.join(__file__, '..', '..'))

Thanks to help me.

benoitc:
current import of the config module doesn't add the `__file__` attribute to the imported module. 

Fix is done by loading the module as suggested in the Python docs winds up setting __file__ as expected (https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly).

Since this change only work on Python >= 3.5 and Python 3.4 is unsupported we also remove its support.

fix #2182

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ matrix:
       env: TOXENV=lint
       dist: xenial
       sudo: true
-    - python: 3.4
-      env: TOXENV=py34
     - python: 3.5
       env: TOXENV=py35
     - python: 3.6
--- a/README.rst
+++ b/README.rst
@@ -28,7 +28,7 @@ The documentation is hosted at http://docs.gunicorn.org.
 Installation
 ------------
 
-Gunicorn requires **Python 3.x >= 3.4**.
+Gunicorn requires **Python 3.x >= 3.5**.
 
 Install from PyPI::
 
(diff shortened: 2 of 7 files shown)


benoitc:
This change is actually removing the support of python 3.4 which means we are slightly out of the scope of a "20.0.1" though I don't think it's really an issue. But let me know. Otherwise we have 2 choices:

* add support of 3.4 (but deprecate it)
* bump to 20.1

Let me known,

benoitc:
@berkerpeksag @jamadden @tilgovi please test and let me know

PR #2202, opened 2019-11-25
Fix socketfromfd

### LINKED ISSUE #2201: 20.0.2 raise OSError
I use `kill -USR2 ` to reload process, but it raise error.

```
[2019-11-25 11:33:44 +0800] [3967] [INFO] Starting gunicorn 20.0.2
[2019-11-25 11:33:44 +0800] [3967] [INFO] Listening at: http://127.0.0.1:8000 (3967)
[2019-11-25 11:33:44 +0800] [3967] [INFO] Using worker: sync
[2019-11-25 11:33:44 +0800] [3970] [INFO] Booting worker with pid: 3970
[2019-11-25 11:33:51 +0800] [3967] [INFO] Handling signal: usr2
[2019-11-25 11:33:51 +0800] [3975] [INFO] Starting gunicorn 20.0.2
  File "/Users/xx/project/py37/bin/gunicorn", line 10, in <module>
    sys.exit(run())
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 58, in run
    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/app/base.py", line 219, in run
    super().run()
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/app/base.py", line 71, in run
    Arbiter(self).run()
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/arbiter.py", line 198, in run
    self.start()
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/arbiter.py", line 155, in start
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/sock.py", line 171, in create_sockets
    sock = fromfd(fd)
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/socketfromfd.py", line 95, in fromfd
    family = _raw_getsockopt(fd, socket.SOL_SOCKET, SO_DOMAIN)
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/socketfromfd.py", line 74, in _raw_getsockopt
    ctypes.byref(optval), ctypes.byref(optlen))
  File "/Users/xx/project/py37/lib/python3.7/site-packages/gunicorn/socketfromfd.py", line 44, in _errcheck_errno
    raise OSError(errno, os.strerror(errno))
OSError: [Errno 42] Protocol not available

```

benoitc:
fix cases where the type of socket is not correctly detected on mac.

fix #2201

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -122,3 +122,11 @@ util/unlink.py
 --------------
 
 backport frop python3 Lib/test/support.py
+
+
+gunicorn.socketfromfd
+---------------------
+
+Under Apache License 2
+
+Copyright (C) 2016  Christian Heimes
--- a/gunicorn/socketfromfd.py
+++ b/gunicorn/socketfromfd.py
@@ -87,8 +87,8 @@ def _raw_getsockopt(fd, level, optname):
 
 def _raw_getsockname(fd):
     sockaddr = SockAddr()
-    len = ctypes.c_int(ctypes.sizeof(sockaddr))
-    _libc_getsockname(fd, sockaddr, len)
+    sockaddrlen = ctypes.c_int(ctypes.sizeof(sockaddr))
+    _libc_getsockname(fd, sockaddr, sockaddrlen)
     return sockaddr.sa_family
 
 def fromfd(fd, keep_fd=True):
--- a/gunicorn/socketfromfd.py
+++ b/gunicorn/socketfromfd.py
@@ -16,10 +16,6 @@ from .util import find_library
 
 __all__ = ('fromfd',)
 
-SO_DOMAIN = getattr(socket, 'SO_DOMAIN', 39)
-SO_TYPE = getattr(socket, 'SO_TYPE', 3)
-SO_PROTOCOL = getattr(socket, 'SO_PROTOCOL', 38)
-
 _libc_name = find_library('c')
 if _libc_name is not None:
     if sys.platform.startswith("aix"):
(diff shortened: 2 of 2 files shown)


benoitc:
we can also probably just use the python method directly on python >= 3.8 , what do you think?

jamadden:
In Python 3.5 and 3.6, the `socket.socket` documentation says that " the values for *family*, *type*, and *proto*" are ignored when a *fileno* is given. In 3.7, that text was changed to explicitly state:

>  the values for *family*, *type*, and *proto* are auto-detected from the specified file descriptor. Auto-detection can be overruled by calling the function with explicit family, type, or proto arguments. **This only affects how Python represents e.g. the return value of socket.getpeername() but not the actual OS resource**.

All versions document `fromfd` with "Address family, socket type and protocol number are as for the socket() function."

So I would think that on 3.7 and above we could skip all of our auto-detection and let Python handle it. (The less ctypes, the better!) `return socket.socket(fileno=fd)`

Maybe we could skip it everywhere since they're documented as being ignored? I haven't looked at why gunicorn is doing this. (`fromfd` doesn't have defaults for those three arguments so we could either `dup()` the `fd` ourself and call `socket.socket()`, or pass in default values to be ignored.)

benoitc:
@jamadden I agree, but I'm not sure for 3.5, 3.6. In doubt i would just use python `fromfd` on python >+ 3.7 and continue to use this change for now.   I think the main reason for it was to make sure the new y created socket was using the correct class.

benoitc:
@jamadden the proto and type are the default used by python if none is set, only the family seems to be the mandatory, so this PR seems to be right.

benoitc:
reading https://bugs.python.org/issue28134  it seems that this module is needed since autodetection is not working on python =< 3.7 alsowe need to make sure to keep the fd around which is only guaranted using fromfd.

jamadden:
Ah, I forgot about that Python bug. So yeah, this looks good for Python <= 3.6.

PR #2206, opened 2019-11-25
config file extension should be a python extension

### LINKED ISSUE #2198: Configuration file requires .py extension since 20.0.1
With gunicorn 20.0.1 and 20.0.2 running on PyPy3.6 (7.1.1), I get startup failures like this (some paths sanitized):

```
gunicorn[9371]: Failed to read config file: /home/.../gunicorn.cfg
gunicorn[9371]: Traceback (most recent call last):
gunicorn[9371]:   File "/home/.../virtualenv/site-packages/gunicorn/app/base.py", line 100, in get_config_from_filename
gunicorn[9371]:     mod = importlib.util.module_from_spec(spec)
gunicorn[9371]:   File "<frozen importlib._bootstrap>", line 558, in module_from_spec
gunicorn[9371]: AttributeError: 'NoneType' object has no attribute 'loader'
```

In a quick scan I couldn't find the commit that changes it, but I think this is caused by "fixed the way the config module is loaded. `__file__` is now available" in 20.0.1. Downgrading to 20.0.0 allows my server to start up normally.

benoitc:
This change make it clear what is configuration file for Gunicorn.
Using a filename with an extension different than a python extension
was never supported. Also it gives us some room for a proper config file.

fix #2198

### DIFF
--- a/docs/source/configure.rst
+++ b/docs/source/configure.rst
@@ -53,9 +53,9 @@ Configuration File
 
 The configuration file should be a valid Python source file with a **python
 extension** (e.g. `gunicorn.conf.py`). It only needs to be readable from the
-file system. More specifically, it does not need to be importable. Any Python
-is valid. Just consider that this will be run every time you start Gunicorn
-(including when you signal Gunicorn to reload).
+file system. More specifically, it does not have to be on the module path
+(sys.path, PYTHONPATH). Any Python is valid. Just consider that this will be
+run every time you start Gunicorn (including when you signal Gunicorn to reload).
 
 To set a parameter, just assign to it. There's no special syntax. The values
 you provide will be used for the configuration values.
--- a/docs/source/configure.rst
+++ b/docs/source/configure.rst
@@ -51,10 +51,11 @@ isn't mentioned in the list of :ref:`settings <settings>`.
 Configuration File
 ==================
 
-The configuration file should be a valid Python source file. It only needs to
-be readable from the file system. More specifically, it does not need to be
-importable. Any Python is valid. Just consider that this will be run every time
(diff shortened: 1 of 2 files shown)


benoitc:
We may still need to put back support of files without python extensions for  a slow deprecation. Lot of  examples on SO and others describe a configuration file without `.py` which is is unfortunate. Thoughts?

docs/source/configure.rst
```diff
@@ -51,10 +51,11 @@ isn't mentioned in the list of :ref:`settings <settings>`.
 Configuration File
 ==================
 
-The configuration file should be a valid Python source file. It only needs to
-be readable from the file system. More specifically, it does not need to be
-importable. Any Python is valid. Just consider that this will be run every time
-you start Gunicorn (including when you signal Gunicorn to reload).
+The configuration file should be a valid Python source file with a **python
+extension** (e.g. `gunicorn.conf.py`). It only needs to be readable from the
+file system. More specifically, it does not need to be importable. Any Python
```
tilgovi: By "does not need to be importable" do you mean it does not have to be on the module path (`sys.path`, `PYTHONPATH`)? If so, saying that more explicitly might be clearer.

tilgovi: By "does not need to be importable" do you mean it does not have to be on the module path (`sys.path`, `PYTHONPATH`)? If so, saying that more explicitly might be clearer.

benoitc: i think this is what it mean, i didn't change  that part. I agree anyway i am making the change

benoitc:
@tilgovi I preferred the explicit way so we make sure that "None" is not something else :)

benoitc:
@tilgovi i made the doc more explicit. If OK I will merge it.

PR #2208, opened 2019-11-26
Revert "Fix socketfromfd"

benoitc:
Reverts benoitc/gunicorn#2202
 

This change break the usage on linux platform: `sockaddr` strut is different between BSDs,Macs and Linux.


benoitc:
@jamadden so i reviewed the patch that trigger the inclusion of socketfromfd: https://github.com/benoitc/gunicorn/commit/ee7af1247b8cc289af7b4bcfd2ff93ecfaf8c9df#diff-cd1a6450c95ae0e121da73674dc3e908 

I'm not sure we need it at all. Using socket.fromfd should be enough according my reading since it's socket.AF_UNIX had the larger length. Thoughts?

benoitc:
@jamadden let me know what you think about c695873a0bd35f9718281288039280388474fbd1 I will do more tests on different platforms

benoitc:
@jamadden at least USR2 on my mac with python 3.5 doesn't fail for me. At C level I agree that shouldn't be much an issue. I think we are OK enough. for that's? Let's wait a little to see if if someone come with an issue. If not I will release a new version today.

ThePrez:
Smoke-tested on IBM i. Looks good! Thanks!

PR #2214, opened 2019-12-06
added the parameter 'min_requests_jitter' in order to overwrite the d…

dblyon:
**The parameter I've added is supposed to only be optional even when using "max_requests_jitter".**

added the parameter 'min_requests_jitter' in order to overwrite the default (0) in 'jitter = randint(0, cfg.max_requests_jitter)'

**First of all many many thanks for your efforts!!!**

I have a python-flask app that takes ~8min to start-up (reading large files and allocating contiguous memory, etc.) and have several (e.g. 3) workers. Workers should re-spawn after "max_requests" + "jitter". But there can be situations when "lots of requests" go to gunicorn, and all workers reach their max_requests in a short amount of time. Which causes one after the other to re-spawn, but they are not fast enough (due to long start-up) and therefore there will be down-time in which requests are not answered until the workers are up again. This new parameter would enable me to set 
e.g.
max_requests = 10000 
min_requests_jitter = 5000
max_requests_jitter = 50000
which would give a process enough time to be ready in time. I can of course set "max_request" and "max_requests_jitter" to very high values without "min_requests_jitter" but then there could still be a situation where "randint" picks a small number and the issue described above occurs. 

Please let me know if I've missed a setting that would take care of this situation otherwise. 

Kind regards,
David

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -716,6 +716,23 @@ class MaxRequests(Setting):
         restarts are disabled.
         """
 
+class MinRequestsJitter(Setting):
+    name = "min_requests_jitter"
+    section = "Worker Processes"
+    cli = ["--min-requests-jitter"]
+    meta = "INT"
+    validator = validate_pos_int
+    type = int
+    default = 0
+    desc = """\
+        The minimum jitter to add to the *max_requests* setting.
+
+        The jitter causes the restart per worker to be randomized by
+        ``randint(minimum_requests_jitter, max_requests_jitter)``. This is intended to stagger worker
+        restarts to avoid all workers restarting at the same time.
+
+        Minimum_requests_jitter can optionally be used if the default (0) should be overwritten.
+        """
 
 class MaxRequestsJitter(Setting):
     name = "max_requests_jitter"
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -53,7 +53,7 @@ class Worker(object):
         self.nr = 0
 
         if cfg.max_requests > 0:
-            jitter = randint(0, cfg.max_requests_jitter)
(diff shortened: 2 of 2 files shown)


tilgovi:
How is this different from increasing max_requests?

dblyon:
Max_requests defines the minimum amount of requests before a worker gets re-spawned. If jitter offset is a random value between 0 and max_requests_jitter there is a chance that the jitter offset is small (close to 0, even if high max_requests_jitter is set to a large number). Imagine a scenario where all workers are close to their max_requests and many requests are hitting gunicorn. Now due to the jitter offset the workers will re-spawn at varying times (max_requests + jitter offest), but if workers take a long time to spawn (upstart time of the python app) and the offset is a low number (which can't be controlled if there is no min_requests_jitter), **then it can happen that there is not a single worker available to answer a request**.
This can be prevented by having a min_requests_jitter parameter and setting this at a higher value. 
e.g. 
max_requests = 5000
min_requests_jitter = 10000
max_requests_jitter = 15000
Another way to achieve the behavior I need is to have the possibility to keep at least one worker alive while others re-spawn and wait until the new workers are ready until killing/re-spawning the last one. I've not found such a parameter and therefore thought min_requests_jitter would be the most simple way to implement what I need.

dblyon:
I see the error in my own logic. You are correct this additional parameter doesn't change anything. Sorry for the mistake! 
Is there a way to achieve the behavior I've described above?

tilgovi:
It sounds like you want a guarantee that at least one worker is running at all times. There is no way to guarantee that using max requests right now.

What is your reason for using max requests? Do you have a memory leak? If so, can you resolve it?

If you must cycle your workers, can you script a cron job to do so using the USR2 signal, followed by a shutdown of the old master process some minutes later after there's been time for the new workers to start?

PR #2219, opened 2019-12-13
Update FAQ - Workers Silently Killed.

monobaila:
Fixes/Documents Issue 2215.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -205,3 +205,30 @@ Check the result::
     tmpfs              65536     0     65536   0% /mem
 
 Now you can set ``--worker-tmp-dir /mem``.
+
+Why are Workers Silently Killed?
+--------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


tilgovi:
Looks good to me!

benoitc:
merged. Thanks all :)

rollokb:
@monobaila hey you mention in your merge request that you can configure the OOM to do something else rather than sigkill?

I'm searching for a way to configure this, andy inks?

monobaila:
> @monobaila hey you mention in your merge request that you can configure the OOM to do something else rather than sigkill?
> 
> I'm searching for a way to configure this, any links?

I've had to read through again to remind myself. I'm afraid I think my wording was a bit clumsy here - It's possible to tell OOM killer not to "SIGKILL by default" - but they're not great alternatives which are system panic instead or disable oom killer, both pretty drastic actions. Their might have been some other possibility but I can't remember exactly. In hindsight this sentence just leads to confusion - apologies.

PR #2220, opened 2019-12-13
Fix sample command syntax highlighting in run.rst

tahnok:
Extra `:` was causing the syntax highlighting to mess up, resulting in the docs containing the `..code-block:: text` markup. Fix it by removing the offending colon

Screenshot from 2019-12-13 15-00-45


berkerpeksag:
Good catch, thanks!

PR #2230, opened 2020-01-04
Use socket.sendfile() instead of os.sendfile().

### LINKED ISSUE #2223: PyPy: module 'os' has no attribute 'sendfile'
#2172  seems to cause the following error with PyPy.

```
  File "xxx/pypyenv/site-packages/gunicorn/workers/ggevent.py", line 33, in <module>
    def _gevent_sendfile(fdout, fdin, offset, nbytes, _os_sendfile=os.sendfile):
AttributeError: module 'os' has no attribute 'sendfile'
```

jamadden:
Fixes #2223.

Unfortunately, eventlet doesn't implement GreenSocket.sendfile, so we have to do it for it.

Add gevent and eventlet to tox.ini and add tests to make sure we can at least import the workers. Some tests that this actually functions would be nice...

Update the gevent and eventlet setup extras to require the versions that are enforced in their worker modules.

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -360,12 +360,6 @@ class Response(object):
             offset = os.lseek(fileno, 0, os.SEEK_CUR)
             if self.response_length is None:
                 filesize = os.fstat(fileno).st_size
-
-                # The file may be special and sendfile will fail.
-                # It may also be zero-length, but that is okay.
-                if filesize == 0:
-                    return False
-
                 nbytes = filesize - offset
             else:
                 nbytes = self.response_length
@@ -378,12 +372,7 @@ class Response(object):
             chunk_size = "%X\r\n" % nbytes
             self.sock.sendall(chunk_size.encode('utf-8'))
 
-        sockno = self.sock.fileno()
-        sent = 0
-
-        while sent != nbytes:
-            count = min(nbytes - sent, BLKSIZE)
-            sent += os.sendfile(sockno, fileno, offset + sent, count)
+        self.sock.sendfile(respiter.filelike, count=nbytes)
 
         if self.is_chunked():
             self.sock.sendall(b"\r\n")
--- a/gunicorn/workers/geventlet.py
+++ b/gunicorn/workers/geventlet.py
@@ -4,8 +4,6 @@
 # See the NOTICE for more information.
 
 from functools import partial
-import errno
-import os
 import sys
 
 try:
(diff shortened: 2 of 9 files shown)


setup.py
```diff
@@ -76,8 +76,8 @@ def run_tests(self):
 ]
 
 extras_require = {
-    'gevent':  ['gevent>=0.13'],
-    'eventlet': ['eventlet>=0.9.7'],
+    'gevent':  ['gevent>=1.4.0'],
+    'eventlet': ['eventlet>=0.24.1'],
```
tilgovi: Good catch.

gunicorn/http/wsgi.py
```diff
@@ -360,12 +360,6 @@ def sendfile(self, respiter):
             offset = os.lseek(fileno, 0, os.SEEK_CUR)
             if self.response_length is None:
                 filesize = os.fstat(fileno).st_size
-
-                # The file may be special and sendfile will fail.
-                # It may also be zero-length, but that is okay.
-                if filesize == 0:
```
tilgovi: Why don't we want this anymore? Is it because the implementations in the stdlib, gevent, and this PR all fall back to reading and sending chunks, rather than calling the underlying OS implementation?

gunicorn/workers/geventlet.py
```diff
@@ -19,22 +17,49 @@
...
-                trampoline(fdout, write=True)
-            else:
-                raise
+def _eventlet_socket_sendfile(self, file, offset=0, count=None):
+    # Based on the implementation in gevent which in turn is slightly
+    # modified from the standard library implementation.
+    if self.gettimeout() == 0:
+        raise ValueError("non-blocking sockets are not supported")
+    if offset:
+        file.seek(offset)
```
tilgovi: Do we have to worry about this not being implemented, or do we want this to raise in the event we try to send from an offset of something unseekable? Alternatively, maybe the argument is here for completeness but is never triggered?

gunicorn/http/wsgi.py
jamadden: Right. In the case that `_sendfile_use_sendfile` gets used (the version of `socket.sendfile` that uses `os.sendfile`) the standard library and gevent implementations handle the 0 size case automatically. That's part `socket.sendfile` being a higher-level API than `os.sendfile.` The implementation provided here for eventlet, of course, doesn't use `os.sendfile`.

gunicorn/workers/geventlet.py
jamadden: Right, that's just matching the `socket.sendfile` signature. gunicorn will never use it.

In the event that some application software calls it on an eventlet socket, it needs to match  (prior to this, eventlet would copy `socket.sendfile` to itself and call that directly, so we need to maintain the signature and semantics as much as possible; of course, it wouldn't actually *work* because the socket was non-blocking, but now that it does and the `ValueError` isn't raised the rest of it should raise the standard exceptions or work the standard way.)

PR #2238, opened 2020-01-13
Added pip requirements files for examples/frameworks

javabrett:
This adds a set of 5 one-liner pip requirements files, plus a sixth which references them all using `-r requirements_...`.  Although the required modules are largely obvious, this documents them and makes them easy to install as a collection when testing the frameworks examples.

Flask, CherryPy, Pyramid, Tornado and WebPy are covered.  Django in a subdirectory is not currently, as I don't know how exactly to test it, input welcome.

Note that tornado requires pinning to `tornado<6` due to #2239 .


tilgovi:
Another option would be to put them all in their own directories with a "requirements.txt".

javabrett:
> Another option would be to put them all in their own directories with a "requirements.txt".
Agreed - if I return here for further changes I'll look into that.

Issue #2244, opened 2020-01-21
Syntax errors crash the process even with --reload


LiquidLemon:
Running `gunicorn --reload` doesn't work as intended. Syntax errors cause it to crash instead of emitting the error and watching for further updates.

```
[2020-01-21 21:01:06 +0100] [19910] [INFO] Starting gunicorn 20.0.4
[2020-01-21 21:01:06 +0100] [19910] [INFO] Listening at: http://127.0.0.1:8000 (19910)
[2020-01-21 21:01:06 +0100] [19910] [INFO] Using worker: sync
[2020-01-21 21:01:06 +0100] [19913] [INFO] Booting worker with pid: 19913
[2020-01-21 21:01:10 +0100] [19913] [INFO] Worker reloading: /home/liquid/tmp/gunicorn/app.py modified
[2020-01-21 21:01:10 +0100] [19913] [INFO] Worker exiting (pid: 19913)
[2020-01-21 21:01:10 +0100] [19931] [INFO] Booting worker with pid: 19931
[2020-01-21 21:01:10 +0100] [19931] [ERROR] invalid syntax (app.py, line 4)
Traceback (most recent call last):
  File "/home/liquid/tmp/gunicorn/venv/lib/python3.7/site-packages/gunicorn/workers/base.py", line 144, in load_wsgi
    self.wsgi = self.app.wsgi()
  File "/home/liquid/tmp/gunicorn/venv/lib/python3.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
    self.callable = self.load()
  File "/home/liquid/tmp/gunicorn/venv/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 49, in load
    return self.load_wsgiapp()
  File "/home/liquid/tmp/gunicorn/venv/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 39, in load_wsgiapp
    return util.import_app(self.app_uri)
  File "/home/liquid/tmp/gunicorn/venv/lib/python3.7/site-packages/gunicorn/util.py", line 358, in import_app
    mod = importlib.import_module(module)
  File "/usr/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
[...]

Issue #2133, opened 2019-10-16
Worker does not pick up proper files to watch for reloading when rebooting


talham7391:
When launching Gunicorn with the `--reload` option, workers will start the Reloader and then load the wsgi app. This is problematic because user files are loaded when the wsgi app is loaded, and the reloader depends on the user files to be loaded so that it can watch them.

So when the Reloader starts, it isn't watching any user files. This isn't an issue when using the default polling reloader as it updates what files it is watching every time it polls but this is an issue with the inotify reloader.

trendels:
The [change in 3054b18](https://github.com/benoitc/gunicorn/commit/3054b184dea1b07581293c203a1c546c7ea3ebca) broke  the default reloader unfortunately. When a file has a syntax error, gunicorn now crashes when reloading, with this exception:

```
Traceback (most recent call last):
  File "/home/st/gunicorn_reload_test/venv/lib/python3.7/site-packages/gunicorn/arbiter.py", line 583, in spawn_worker
    worker.init_process()
  File "/home/st/gunicorn_reload_test/venv/lib/python3.7/site-packages/gunicorn/workers/base.py", line 119, in init_process
    self.load_wsgi()
  File "/home/st/gunicorn_reload_test/venv/lib/python3.7/site-packages/gunicorn/workers/base.py", line 160, in load_wsgi
    self.reloader.add_extra_file(exc_val.filename)
AttributeError: 'NoneType' object has no attribute 'add_extra_file'
```
The reason is that `load_wsgi()` calls `self.reloader.add_extra_file()` [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L158). The commit linked above moved the initialization of `self.reloader` after the call to `load_wsgi()`, so it is not properly initialized.

tilgovi:
@trendels I'm truly sorry I missed that in my reviews. I've opened #2248. Feel free to pick up the issue yourself, if you're so inclined, and I'm happy to review. I'll try to make time before our next release, if you cannot.

trendels:
@tilgovi thanks for creating the issue. I have a fix ready but I'm still figuring out how to write a good test for it. I'll make a pull request later today.

PR #2249, opened 2020-01-26
Issue 2244

trendels:
This PR contains a test and fix for #2244 

The "chicken and egg problem" with the (inotify-based) reloader depending on the wsgi app to be loaded first, and `load_wsgi` depending on the reloader to be present is solved by initializing the reloader first, but only starting it after the wsgi app was loaded.

I've added an integration test that runs `gunicorn --reload` in a tmpdir, replaces the WSGI app and checks for changes in the HTTP response via a unix domain socket. In order to not slow down the tests too much, I've made the default poll interval for the poll-based reloader (1 second) overridable via an (undocumented) [environment variable](https://github.com/benoitc/gunicorn/pull/2249/commits/f1f0d011aec16adcb7b28b7fb5b0f3bb391de2fe). I hope this is OK.

Initially, the test was very flaky and failed most of the time even with the fix (but only when running under `pytest`): gunicorn would pick up the changed file and restart the worker, but the new worker would still be running the old code! It took me a while to figure out that this must have something to do with the compiled bytecode (maybe a timestamp issue?). [Starting gunicorn with writing of bytecode disabled](https://github.com/benoitc/gunicorn/pull/2249/commits/6923709396afca7f8966bc19528094bd5af566d0#diff-3d4ef4fd9b978e57e60dfb39ca4c5baeR27-R31) fixed the issue for me. I was not able to reproduce this outside of pytest.

There was also an issue where the inotify-based reloader would throw an exception if a directory it tried to watch did not exist, which happens when using `--reload` and `--chdir` together (the poll-based reloader already ignores all missing/unreadable files).

### DIFF
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -104,7 +104,8 @@ if has_inotify:
             self._dirs = self.get_dirs()
 
             for dirname in self._dirs:
-                self._watcher.add_watch(dirname, mask=self.event_mask)
+                if os.path.isdir(dirname):
+                    self._watcher.add_watch(dirname, mask=self.event_mask)
 
             for event in self._watcher.event_gen():
                 if event is None:
--- a/gunicorn/workers/base.py
+++ b/gunicorn/workers/base.py
@@ -117,8 +117,6 @@ class Worker(object):
 
         self.init_signals()
 
-        self.load_wsgi()
-
         # start the reloader
         if self.cfg.reload:
             def changed(fname):
@@ -132,6 +130,9 @@ class Worker(object):
             reloader_cls = reloader_engines[self.cfg.reload_engine]
             self.reloader = reloader_cls(extra_files=self.cfg.reload_extra_files,
                                          callback=changed)
+
+        self.load_wsgi()
+        if self.reloader:
             self.reloader.start()
 
         self.cfg.post_worker_init(self)
--- a/tests/test_reload.py
+++ b/tests/test_reload.py
@@ -1,0 +1,68 @@
+import unittest.mock as mock
+
+from gunicorn.app.base import Application
+from gunicorn.workers.base import Worker
+from gunicorn.reloader import reloader_engines
+
+
+class ReloadApp(Application):
+    def __init__(self):
+        super().__init__("no usage", prog="gunicorn_test")
+
+    def do_load_config(self):
+        self.load_default_config()
+        self.cfg.set('reload', True)
+        self.cfg.set('reload_engine', 'poll')
+
+
+class SyntaxErrorApp(ReloadApp):
+    def wsgi(self):
+        error = SyntaxError('invalid syntax')
+        error.filename = 'syntax_error_filename'
+        raise error
+
+
+class MyWorker(Worker):
+    def run(self):
+        pass
+
+
+def test_reload_on_syntax_error():
+    """
+    Test that reloading works if the application has a syntax error.
+    """
+    reloader = mock.Mock()
+    reloader_engines['poll'] = lambda *args, **kw: reloader
+
+    app = SyntaxErrorApp()
+    cfg = app.cfg
+    log = mock.Mock()
+    worker = MyWorker(age=0, ppid=0, sockets=[], app=app, timeout=0, cfg=cfg, log=log)
+
+    worker.init_process()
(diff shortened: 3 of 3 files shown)


trendels:
I see there are still some failures on Travis CI. I'm not sure what the issue is. Possibly the same one I was seeing earlier. The tests pass when I run them locally (except pypy which I don't have installed):

```
$ python3 -m venv tox-venv
$ tox-venv/bin/pip install --upgrade tox coverage
$ tox-venv/bin/tox --recreate
[...]
--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

________________________________________________ summary ________________________________________________
  py35: commands succeeded
  py36: commands succeeded
  py37: commands succeeded
  py38: commands succeeded
ERROR:  pypy3: InterpreterNotFound: pypy3
  lint: commands succeeded
```

trendels:
I've ran the tests in CI some more, and they [remain flaky unfortunately](https://travis-ci.org/trendels/gunicorn/builds), even with slightly increased timeouts.

I don't think the failures are real errors, but that the detection of filesystem changes can be arbitrarily slow in the CI environment. Maybe the best way would be to disable this test in CI?

gunicorn/workers/base.py
```diff
@@ -132,7 +130,9 @@ def changed(fname):
             reloader_cls = reloader_engines[self.cfg.reload_engine]
             self.reloader = reloader_cls(extra_files=self.cfg.reload_extra_files,
                                          callback=changed)
-            self.reloader.start()
+
+        self.load_wsgi()
+        self.reloader.start()
```
tilgovi: This needs to be wrapped in a conditional in case the reloader doesn't exist, right?

gunicorn/reloader.py
```diff
@@ -12,10 +12,11 @@
 import threading
 
 COMPILED_EXT_RE = re.compile(r'py[co]$')
+POLL_INTERVAL = float(os.getenv('GUNICORN_RELOADER_POLL_INTERVAL', '1'))
```
tilgovi: Should we expose this on the CLI somehow? Or is this intended just for the tests?

gunicorn/workers/base.py
trendels: You're right, I've fixed it.

gunicorn/reloader.py
trendels: This is just intended for the tests, so the test does not take multiple seconds each time.

tilgovi:
If you'd like to squash up the commits, feel free to do so!

gunicorn/workers/base.py
```diff
@@ -132,6 +130,9 @@ def changed(fname):
             reloader_cls = reloader_engines[self.cfg.reload_engine]
             self.reloader = reloader_cls(extra_files=self.cfg.reload_extra_files,
                                          callback=changed)
+
+        self.load_wsgi()
+        if self.reloader:
```
berkerpeksag: Why do we need this branch? Can't we just simply revert https://github.com/benoitc/gunicorn/commit/3054b184dea1b07581293c203a1c546c7ea3ebca?

trendels: That commit was meant as a fix for #2133 (inotify reloader does not pick up files), but also introduced the regression in  #2244 (syntax errors crash the process on reload). So a simple revert would fix the regression but also re-break the inotify reloader unfortunately.

Doing things in this particular order (load_wsgi, then start the reloader) fixes the original problem, but without the regression.

trendels:
> I wonder if it would be possible to add a test case for original problem that broke reloading of files with syntax error? See #2133 for the original report.

The cause for this was an uninitialized variable, so I guess I can add a unit test for that.

I suggest that I do that and remove the integration test instead. While I think it would be very useful to have such a test, I continue to see spurious failures when running it, and I don't want to introduce a flaky test to the test suite.

trendels:
I've replaced the flaky test with more targeted unit tests for the specific issues #2244 and #2133 that this PR addresses, and squashed everything into two commits (two because f39b065 fixes an unrelated issue that I noticed during testing).

tilgovi:
This is great, @trendels. Thanks for being responsive on the tests and feedback. Feel free to nudge me for followup if I'm being slow.

Issue #2247, opened 2020-01-23
Gunicorn changes its listening interfaces when PORT is in environment


kmichel2:
Hi,

When the `PORT` variable is in the environment, [the default value of the --bind option changes](https://github.com/benoitc/gunicorn/blob/73ae195068d99ea8c9f26a999178a2cfb8fb057f/gunicorn/config.py#L562) from `127.0.0.1` to `0.0.0.0`.

I can see how this can be useful on some deployments (such as on Heroku) but this behaviour might be surprising and possibly a security issue when suddenly you're listening on a public interface in addition to localhost.

Changing this might be hard and break too many things but I think it should at least be documented : I haven't found any trace of this in the documentation except for an [2012 entry in the changelog](https://docs.gunicorn.org/en/stable/2012-news.html#id3), specifically mentioning Heroku.

[The Configuration documentation](https://docs.gunicorn.org/en/stable/configure.html) very clearly enumerates the three places when gunicorn can read its configuration but does not include the environment.

[The Settings list](https://docs.gunicorn.org/en/stable/settings.html) does mention a few environment variables: `GUNICORN_CMD_ARGS`, `PYTHONUNBUFFERED`, `SENDFILE`, `FORWARDED_ALLOW_IPS`, `WEB_CONCURRENCY` but not the `PORT` variable. The default value of `--bind` is documented as only `['127.0.0.1:8000']`

The automatic binding to the public interface makes [Honcho](https://github.com/nickstenning/honcho) harder to use safely as a local dev server since Honcho will always set the PORT variable.

I can work on improving the documentation and/or removing the behaviour (maybe for a new major version) with a documented migration path (such as reading the environment in the configuration file), depending on what you prefer.

PS: There are 4 more undocumented environment dependencies, but not really specific to gunicorn:
  - `LD_LIBRARY_PATH` in [util.py](https://github.com/benoitc/gunicorn/blob/73ae195068d99ea8c9f26a999178a2cfb8fb057f/gunicorn/util.py#L656) but it seems to be dead code.
  - `NOTIFY_SOCKET` in [systemd.py](https://github.com/benoitc/gunicorn/blob/73ae195068d99ea8c9f26a999178a2cfb8fb057f/gunicorn/systemd.py#L62)
[...]

PR #2253, opened 2020-01-29
Document environment variables and $PORT

kmichel-sereema:
As described in #2247, this documents the effect of the `$PORT` environment variable on the default value of `--bind` and documents the impact and priority of the environment variables in the configuration overview.

~The dead code using the environment variable `LD_LIBRARY_PATH` is also removed.~

### DIFF
--- a/docs/source/configure.rst
+++ b/docs/source/configure.rst
@@ -6,15 +6,16 @@ Configuration Overview
 
 Gunicorn reads configuration information from five places.
 
-Gunicorn first reads environment variables for some
-configuration :ref:`settings <settings>`.
+Gunicorn first reads environment variables for some configuration
+:ref:`settings <settings>`.
 
-Gunicorn will then read configuration from a framework
-specific configuration file. Currently this only affects Paster applications.
+Gunicorn then reads configuration from a framework specific configuration
+file. Currently this only affects Paster applications.
 
-The third source of configuration information is a configuration file that is
-optionally specified using command line arguments. Anything specified in the
-Gunicorn config file will override any framework specific settings.
+The third source of configuration information is an optional configuration file
+``gunicorn.conf.py`` searched in the current working directory or specified
+using a command line argument. Anything specified in this configuration file
+will override any framework specific settings.
 
 The fourth place of configuration information are command line arguments
 stored in an environment variable named ``GUNICORN_CMD_ARGS``.
--- a/docs/source/configure.rst
+++ b/docs/source/configure.rst
@@ -4,23 +4,35 @@
 Configuration Overview
 ======================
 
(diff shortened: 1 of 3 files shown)


benoitc:
@kmichel-sereema Thanks for the patch!

Can you move the commit "Remove dead code" in a separate PR since the concern is different? I will look on the rest asap.

kmichel-sereema:
Fixed, the dead code change is in #2254 :)

docs/source/configure.rst
```diff
@@ -4,23 +4,35 @@
 Configuration Overview
 ======================
 
-Gunicorn pulls configuration information from three distinct places.
+Gunicorn pulls configuration information from five distinct places.
```
tilgovi:
```suggestion
Gunicorn reads configuration information from five places.
```

docs/source/configure.rst
```diff
@@ -4,23 +4,35 @@
...
 
-The first place that Gunicorn will read configuration from is the framework
+Gunicorn will first read environment variables for the default value of some
```
tilgovi:
```suggestion
Gunicorn first reads environment variables for some
```

docs/source/configure.rst
```diff
@@ -4,23 +4,35 @@
...
+configuration :ref:`settings <settings>`.
+
+The second place that Gunicorn will read configuration from is the framework
```
tilgovi:
```suggestion
Gunicorn will then read configuration from a framework
```

docs/source/configure.rst
```diff
@@ -4,23 +4,35 @@
...
 specific configuration file. Currently this only affects Paster applications.
 
-The second source of configuration information is a configuration file that is
-optionally specified on the command line. Anything specified in the Gunicorn
-config file will override any framework specific settings.
+The third source of configuration information is a configuration file that is
+optionally specified using command line arguments. Anything specified in the
+Gunicorn config file will override any framework specific settings.
```
tilgovi: Default is `gunicorn.conf.py`. If it's present, it is used implicitly. Maybe we can make that clear here.

docs/source/settings.rst
```diff
@@ -1164,6 +1164,10 @@ Multiple addresses can be bound. ex.::
 will bind the `test:app` application on localhost both on ipv6
 and ipv4 interfaces.
 
+If the ``PORT`` environment variable is defined, the default
+is ``['0.0.0.0:$PORT']``. If it is not defined, the default
+is ``['127.0.0.1:8000']``.
+
```
tilgovi: 👍 thank you!

docs/source/configure.rst
kmichel-sereema: That's a good idea, I added documentation for this.

PR #2257, opened 2020-02-03
Replace and run the sitemap generator

kmichel-sereema:
This replaces the very old sitemap generator which was over 2kloc and
only compatible with Python 2.

According to the stored lastmod, the generator wasn't used since 2010.

The minimal replacement script scan the static site for html files and
uses git to deduce the last modification date of each page.

The sitemap xmlns version was updated to the latest 0.9 from
sitemaps.org .

The index page was given a higher priority since the other pages
are just redirects to the index with anchors.

The output file is pretty printed to help with diffs.

Static assets (css, images...) aren't listed in the sitemap anymore.


kmichel-sereema:
Yeah, I wrote the small replacement in the spirit of keeping feature parity during a cleanup but the site doesn't really need a sitemap to be indexed by search engines.

This is especially true on gunicorn.org where almost all pages are actually just redirects to the home page.

tilgovi:
Ready to merge?

benoitc:
thanks!

Issue #2268, opened 2020-02-13
ModuleNotFoundError: No module named '/src/myproject'


TZubiri:
In order to pass python code to gunicorn, one has to use this command, as per the documentation:

    gunicorn myproject:app

by this they mean that you need to have a myproject.py file with an app class in it. Why remove the .py I'm not sure. Of course if we run this command from a different directory than the myproject.py file an error we occur, so naturally we run:

    gunicorn /src/myproject:app

This doesn't work and yields the following error:

    ModuleNotFoundError: No module named '/src/myproject'

/src/myproject.py didn't work either, it's basically down to the fact that there is no absolute path support, one needs to be on the same directory, so one needs to use cd, if one is using docker, we need to use the workdir directive. This is horrible because we don't want to introduce state dependent operations, but we must now.

There's also the existence of --chdir, I haven't looked much into it, but whoever implemented this option absolutely missed implementing absolute path support instead. Now a user has 3 subpar ways of working around this issue. 

Sometimes you don't need to add a feature, you just need to fix a bug.

TZubiri:
Here is the Dockerfile I'm using. For deterministic testing:

FROM python:3.8-buster

RUN pip3 install flask
RUN pip3 install gunicorn

ADD aegis.py /src/aegis.py

ENTRYPOINT gunicorn /src/aegis:app

Removing the word

jamadden:
This is not a bug.

gunicorn uses Python's standard [import mechanism](https://docs.python.org/3/library/importlib.html) to [locate the application](https://github.com/benoitc/gunicorn/blob/7d8c92f48a6d9cee6b15fbdade6981721182d073/gunicorn/util.py#L57). That means that Python's [sys.path](https://docs.python.org/3/library/sys.html#sys.path) must contain the module.

The exact details to arrange that will depend on the exact deployment scenario. It the simplest case on the command line, one could achieve that with by setting an appropriate [PYTHONPATH environment variable](https://docs.python.org/3/using/cmdline.html?highlight=pythonpath#envvar-PYTHONPATH), for example, `$ PYTHONPATH=/src gunicorn myproject:app`. More often, the project is installed in a virtual environment that also contains gunicorn so that they automatically share the same value for `sys.path`.

TZubiri:
Hello, thank you for the quick response.

[...]

PR #2269, opened 2020-02-14
Module not a path error

TZubiri:
https://github.com/benoitc/gunicorn/issues/2268

I'm adding a specific error message when the user tries to pass a path instead of a module.


TZubiri:
P.s: Currently the error message is passed as an exception that raises an exception.

The first step would be to collapse this into a single exception.

Ideally the error message would be displayed to stdout like the rest of the messages:
[2020-02-13 21:18:09 -0300] [32678] [ERROR] Exception in worker process

TZubiri:
There was a whitespace issue. I didn't run the test suite so this was kinda expected, I'll create a new pull request if there's a positive response in the issue opened.

PR #2277, opened 2020-02-25
Do not raise and crash worker on ENOTCONN error

### LINKED ISSUE #1913: Sockets
Service is running on a kubernetes pod, and out of nowhere and without any specific causes, it happens off and on:
```
  File "/usr/local/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 134, in handle
    req = six.next(parser)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/parser.py", line 41, in __next__
    self.mesg = self.mesg_class(self.cfg, self.unreader, self.req_count)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 181, in __init__
    super(Request, self).__init__(cfg, unreader)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 54, in __init__
    unused = self.parse(self.unreader)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 230, in parse
    self.headers = self.parse_headers(data[:idx])
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 74, in parse_headers
    remote_addr = self.unreader.sock.getpeername()
OSError: [Errno 107] Transport endpoint is not connected
[2018-11-04 17:57:55 +0330] [31] [ERROR] Socket error processing request.
  File "/usr/local/lib/python3.6/site-packages/gunicorn/workers/sync.py", line 134, in handle
    req = six.next(parser)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/parser.py", line 41, in __next__
    self.mesg = self.mesg_class(self.cfg, self.unreader, self.req_count)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 181, in __init__
    super(Request, self).__init__(cfg, unreader)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 54, in __init__
    unused = self.parse(self.unreader)
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 230, in parse
    self.headers = self.parse_headers(data[:idx])
  File "/usr/local/lib/python3.6/site-packages/gunicorn/http/message.py", line 74, in parse_headers
    remote_addr = self.unreader.sock.getpeername()
OSError: [Errno 107] Transport endpoint is not connected
```

JordanP:
A couple of socket operations can fail with ENOTCONN error if the
other side of the connection is not connected anymore. In that case,
let's not crash the whole worker and give a chance to accept new
connections.

In my case, the operation that sometimes fails is a "getpeername()",
which was introduced in b07532be752668be5eb5dbd0a8303abf5c219c99
(v19.8.0). Someone in https://github.com/benoitc/gunicorn/issues/1913
metionned that v19.7.1 was working fine so it matches.

Fixes #1913

### DIFF
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -73,11 +73,13 @@ class AsyncWorker(base.Worker):
                 self.log.debug("Error processing SSL request.")
                 self.handle_error(req, client, addr, e)
         except EnvironmentError as e:
-            if e.errno not in (errno.EPIPE, errno.ECONNRESET):
+            if e.errno not in (errno.EPIPE, errno.ECONNRESET, errno.ENOTCONN):
                 self.log.exception("Socket error processing request.")
             else:
                 if e.errno == errno.ECONNRESET:
                     self.log.debug("Ignoring connection reset")
+                elif e.errno == errno.ENOTCONN:
+                    self.log.debug("Ignoring socket not connected")
                 else:
                     self.log.debug("Ignoring EPIPE")
         except Exception as e:
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -285,11 +285,13 @@ class ThreadWorker(base.Worker):
                 self.handle_error(req, conn.sock, conn.client, e)
 
         except EnvironmentError as e:
-            if e.errno not in (errno.EPIPE, errno.ECONNRESET):
+            if e.errno not in (errno.EPIPE, errno.ECONNRESET, errno.ENOTCONN):
                 self.log.exception("Socket error processing request.")
             else:
                 if e.errno == errno.ECONNRESET:
                     self.log.debug("Ignoring connection reset")
+                elif e.errno == errno.ENOTCONN:
+                    self.log.debug("Ignoring socket not connected")
                 else:
                     self.log.debug("Ignoring connection epipe")
         except Exception as e:
--- a/gunicorn/workers/sync.py
+++ b/gunicorn/workers/sync.py
@@ -146,11 +146,13 @@ class SyncWorker(base.Worker):
                 self.log.debug("Error processing SSL request.")
                 self.handle_error(req, client, addr, e)
         except EnvironmentError as e:
-            if e.errno not in (errno.EPIPE, errno.ECONNRESET):
+            if e.errno not in (errno.EPIPE, errno.ECONNRESET, errno.ENOTCONN):
                 self.log.exception("Socket error processing request.")
             else:
                 if e.errno == errno.ECONNRESET:
                     self.log.debug("Ignoring connection reset")
+                elif e.errno == errno.ENOTCONN:
+                    self.log.debug("Ignoring socket not connected")
                 else:
                     self.log.debug("Ignoring EPIPE")
         except Exception as e:


jethrogillgren:
Hi all, thanks for this PR. I just wanted to report that we're successfully using it to work around https://github.com/benoitc/gunicorn/issues/1913 in a kubernetes install, gevent worker class, --timeout specified.
Thanks for the fix. Would love to see it merged in!

JordanP:
@jethrogillgren good to know thanks ! 

I don't think this is going to be merged soon. The maintainers don't seem to see the issue, which is weird because it's clearly documented that `getpeername()` (like some other networking operations) can raise a `ENOTCONN`. But somehow they decided not to catch that exception and that that exception could tear down the whole worker...

SaschaHeyer:
will this be merged soon?

JordanP:
@tilgovi thanks !

@benoitc if you have a moment, it would be great to have this reviewed.

JordanP:
@benoitc the `ENOTCONN` error is discussed in great length and more here https://stackoverflow.com/questions/900042/what-causes-the-enotconn-error 

Now, I am not a TCP expert and I suspect half of the respondents neither. But what seems to be agreed upon is that a `ENOTCONN` can pop here and there. The question is: what should Gunicorn do about it ? For which I don't know the best answer, but please do not let it crash the whole worker, i.e do not let the exception bubble up too much.

benoitc:
> @benoitc the `ENOTCONN` error is discussed in great length and more here https://stackoverflow.com/questions/900042/what-causes-the-enotconn-error
> 
> Now, I am not a TCP expert and I suspect half of the respondents neither. But what seems to be agreed upon is that a `ENOTCONN` can pop here and there. The question is: what should Gunicorn do about it ? For which I don't know the best answer, but please do not let it crash the whole worker, i.e do not let the exception bubble up too much.

Do you have any other cases that trigger that error apart the `getpeername` call? Such error should really not happen when reading/writing. If it is then it means we have a bug to fix. The other case I see it can happen is when closing the socket or forcing it with a shutdown , but then it would mean we don't have the order of actions correct, and in any case that should be fixed at that place.

What I'm suggesting is to carefully list the case where we say it happens today and case by case see how to solve it. Instead of simply catching and ignoring it. Do you already have such list? Maybe your patch could be redone in different commits to handle each cases. What do you think?

JordanP:
In my system, it seems it only happens with `getpeername()`, but I guess it's because it's the first network call that is attempted on a "TCP connection in a weird state".  

Attached is one example of a stacktrace, but they all look the same afaik:
image

Sorry for the screenshot, I know it sucks :p 

One thing to mention is that I have `worker_class = "gevent"` and `keepalive = 630` in my Gunicorn config.

I can rework the patch to "except" closer to the `getpeername()` call, but I am not sure what I should do in the `except` block: just log or raise another exception (if so, which one ?)

LukasDeco:
Will this get merged soon? I'm experiencing the same `getpeername()` issue and I'd like to resolve it. Will this merge resolve it? Thanks!

tilgovi:
FreeBSD documents `getpeername()` as returning `ENOTCONN` and `ECONNRESET`, depending on the reason that the socket is closed. The Linux man pages I can find seem to only document `ENOTCONN`.

I think it's possible to get `ENOTCONN` from other calls, but only if Gunicorn itself has shut down its end of the socket (calling `shutdown` or `close`). We raise `StopIteration` in the case that we shut the socket down so we should never end up in the parser with a socket that is shut down.

We could handle `ENOTCONN` explicitly in the message parser, but that would be inconsistent with the rest of our error handling. The message class itself is not concerned with iteration, the parser handles no errors, and the parser can read from any iterable (not just a socket).

I'm going to merge this change. I don't see a better place to handle this error and I think we should handle it here with `ECONNRESET`.

There may be room in the future to make a deeper change and get all the socket-related error handling to raise a single end of request / request abort exception, maybe in the `SocketUnreader`. That would get all this error handling out of the various worker base classes and into a single place that neatly encapsulates the socket, but this would be a much bigger change.

benoitc:
I am not sure I share totally  this analysis. enoconn should never happen when the socket is accepted unless it's unblocked. so decorating getpeername would have been probably enough. I propose to add a debug log when calling this function to check it.

Anyway that kind of a breaking change i will bump and make a release with it today.

tilgovi:
> econn should never happen when the socket is accepted

You may be right. I think the code in the Linux kernel for this is here: https://github.com/torvalds/linux/blob/d652d5f1eeeb06046009f4fcb9b4542249526916/net/ipv4/af_inet.c#L773

I should have checked this more closely. I cannot find the bug where Gunicorn would close the socket and then try to read another request, though.

tilgovi:
In the backtrace provided above the error occurs during the keepalive loop. That could be explained by the application code closing the socket even if Gunicorn does not. However, in #1913 there is a backtrace for the sync worker, which has no keepalive loop. I cannot see any way Gunicorn would have closed that socket.

I can only conclude that there is some way to get ENOTCONN from the kernel after `accept()`, even if I cannot see it.

Whatever the reason for the connection being closed, there is nothing for Gunicorn to do here. Again, I think it would be nice to handle all socket exceptions in `SocketUnreader`, but we don't do that right now. We do handle similar errors in the worker base classes here, like ECONNRESET and EPIPE, so I think it makes sense to add this. I don't think this is a breaking change. I think it is a fixing change.

tilgovi:
I've opened #2479 with changes to remove the call to `getpeername`. I still think the change from this PR is worth keeping so that workers don't crash if the application closes the socket.

PR #2282, opened 2020-03-01
Fix simple typo: convertable -> convertible

timgates42:
There is a small typo in examples/websocket/gevent_websocket.py, examples/websocket/websocket.py.
Should read `convertible` rather than `convertable`.


jamadden:
Thanks!

PR #2294, opened 2020-03-19
Bug #1913: Prevent OSError bubbling when socket disconnected

asantoni:
Here's a fix for Bug #1913. I'm not sure if this is the best way to do it, but it works.


JordanP:
@asantoni I think this is similar to https://github.com/benoitc/gunicorn/pull/2277

Issue #2303, opened 2020-04-10
Gevent worker won't exit if keepalive is set and requests keep getting handled


jkemp101:
I feel like I might be missing something but while investigating another issue it looks like this loop will never exit for a gevent worker if using a keepalive and requests keep coming in within the keepalive period. Shouldn't the `while True:` be `while self.alive:`?  Why would we want to stay in this loop if we know the process should exit? I think when this happens the worker eventually gets killed which means server hooks like `worker_exit()` won't get called.

https://github.com/benoitc/gunicorn/blob/8cb2bd2329a94c6f22f35f3ee5f1d725ab798101/gunicorn/workers/base_async.py#L45-L55

PR #2304, opened 2020-04-10
Exit async keepalive request loop when self.alive is False

jkemp101:
Fixes https://github.com/benoitc/gunicorn/issues/2303

### DIFF
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -42,7 +42,7 @@ class AsyncWorker(base.Worker):
                 else:
                     # keepalive loop
                     proxy_protocol_info = {}
-                    while True:
+                    while self.alive:
                         req = None
                         with self.timeout_ctx():
                             req = next(parser)


benoitc:
LGTM . I will test .

tilgovi:
Ooh, I merged this too quickly, I think. Isn't the `self.alive` check in `handle_request` enough to eventually cause this to exit? I think this loop shouldn't terminate because that might cut the connection in the middle of a client sending the next keepalive request on the same socket. Instead, we want the loop to just force a "Connection: close" header on the next request.

tilgovi:
Then, after the `handle_request` loop determines that the next request should close, and sends back the header to tell the client that it will close, it raises `StopIteration` to exit this outer loop.

tilgovi:
I reverted the merge. Sorry for the noise. I'm happy to re-apply if my assessment is wrong. Thank you for the PR, @jkemp101! Please let me know if I missed something.

jkemp101:
@tilgovi It looks like you are handling the case where you hit max requests in https://github.com/benoitc/gunicorn/commit/ebb41da4726e311080e69d1b76d1bc2769897e78 but my original issue was just having the process exit cleanly after receiving a SIGTERM while it keeps getting requests. Don't we still need to find a way out of that loop for that case?

tilgovi:
@jkemp101 would this resolve both?

``` python
diff --git a/gunicorn/workers/base_async.py b/gunicorn/workers/base_async.py
index 0aaaa4d..6e7ae3f 100644
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -99,9 +99,8 @@ class AsyncWorker(base.Worker):
                 if self.alive:
                     self.log.info("Autorestarting worker after current request.")
                     self.alive = False
-                resp.force_close()
 
-            if not self.cfg.keepalive:
+            if not self.alive or not self.cfg.keepalive:
                 resp.force_close()
 
             respiter = self.wsgi(environ, resp.start_response)
diff --git a/gunicorn/workers/gthread.py b/gunicorn/workers/gthread.py
index c7f64b6..5cd3e8a 100644
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -313,7 +313,7 @@ class ThreadWorker(base.Worker):
                     self.alive = False
                 resp.force_close()
 
-            if not self.cfg.keepalive:
+            if not self.alive or not self.cfg.keepalive:
                 resp.force_close()
             elif len(self._keep) >= self.max_keepalived:
                 resp.force_close()
```

tilgovi:
@jkemp101 opened #2313. Your review would be appreciated!

PR #2313, opened 2020-04-20
Disable keepalive during graceful shutdown

tilgovi:
Fix graceful shutdown behavior so that clients receive notice to close
the connection.

After #2304 and the follow-up in ebb41da, Joe Kemp noticed that, while
the new behavior would successfully indicate a connect close after
hitting the maximum request limit for a worker during graceful shutdown,
the worker would not indicate a connection close if if it had not hit
the maximum request limit.

This change ensures that the worker exits gracefully when hitting the
request limit and also indicates to clients that the connection should
close once the shutdown begins.

### DIFF
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -99,9 +99,8 @@ class AsyncWorker(base.Worker):
                 if self.alive:
                     self.log.info("Autorestarting worker after current request.")
                     self.alive = False
-                resp.force_close()
 
-            if not self.cfg.keepalive:
+            if not self.alive or not self.cfg.keepalive:
                 resp.force_close()
 
             respiter = self.wsgi(environ, resp.start_response)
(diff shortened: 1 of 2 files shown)


jkemp101:
This looks really good to me. I tested with gevent on my local Mac using the test script below. I reproduced with and without max-requests using a commit from last month before your recent commits. Sending a SIGTERM would wait until the graceful timeout expired and then kill the workers. The client would never get a `'Connection': 'close'` header. Using this branch everything worked as expected. The worker exited quickly and sent the appropriate `'Connection': 'close'` header. 


```
import urllib3
import time

http = urllib3.PoolManager()

while True:
    r = http.request("GET", "http://localhost:8000")
    hp = http.connection_from_host("localhost", 8000)
    print(r.data, r.headers, len(hp.pool.queue))
    time.sleep(1)
```
`PYTHONPATH=. ./g --chdir examples/frameworks flaskapp:app --keep-alive 10 -k gevent -w 1`

tilgovi:
@jkemp101 thanks for the speedy testing!

PR #2314, opened 2020-04-20
Safe Worker Reaping

tilgovi:
See individual commits:

- Reap workers on the main execution context.
- Remove workers from the workers dictionary only when reaping.
- Adjust iterations of the workers dictionary because there is no longer any concurrent modification.


tilgovi:
This is not safe as is due to this code that might cause Gunicorn to miss signals:

``` python
    def signal(self, sig, frame):
        if len(self.SIG_QUEUE) < 5:
            self.SIG_QUEUE.append(sig)
            self.wakeup()
```

tilgovi:
Ahh, I think I may have identified why we had SIGCHLD separately handled?

Is it because some systems exhibit a behavior where not calling `os.waitpid` in the signal execution context causes infinite recursion? 

http://poincare.matf.bg.ac.rs/~ivana/courses/ps/sistemi_knjige/pomocno/apue/APUE/0201433079/ch10lev1sec7.html

tilgovi:
Python actually checks for the situation I linked to, taking care not to re-install the signal handler for `SIGCHLD` automatically if it might cause infinite recursion.

https://github.com/python/cpython/blob/62183b8d6d49e59c6a98bbdaa65b7ea1415abb7f/Modules/signalmodule.c#L334

I've modified the first commit to address that and commented the code to indicate this case.

gunicorn/arbiter.py
```diff
@@ -636,11 +640,5 @@ def kill_worker(self, pid, sig):
             os.kill(pid, sig)
         except OSError as e:
             if e.errno == errno.ESRCH:
-                try:
```
benoitc, line 639: why removing it?

gunicorn/arbiter.py
```diff
@@ -186,7 +186,6 @@ def init_signals(self):
         # initialize all signals
         for s in self.SIGNALS:
             signal.signal(s, self.signal)
-        signal.signal(signal.SIGCHLD, self.handle_chld)
```
benoitc, line 189: the reason we want to have the CGHLD signal handle separately is that we don't want to queue on it and ignore it  or having to reinstall it.  afaik originally it was just waking the arbiter to do asynchronous reaping. Maybe we should remove instead the `reap_workers` call there and let it async? That way we maximise the change to not loose a signal in the queue, since it's processed before.

benoitc:
made some comment, but i'm still unsure to understand the need of this patch. Can you clarify what is it  trying to solve? Do we have a way to reproduce the issue?

benoitc (commented):
made some comment, but i'm still unsure to understand the need of this patch. Can you clarify what is it  trying to solve? Do we have a way to reproduce the issue?

gunicorn/arbiter.py
tilgovi: It duplicates code in `reap_workers`. If we can guarantee we don't miss the CHLD signal, then we will not need this.

gunicorn/arbiter.py
tilgovi: I'm re-reading the links I posted and I don't think we need to reinstall the handler. I think Python will do it automatically because it's safe on all platforms we support (it won't lead to infinite loop). That is why we never needed to reinstall it before, and that should not change. I can remove that part. But if we want to be extra safe, I can keep the reinstall, but it is only safe to do on those platforms _after_ reaping.

> Maybe we should remove instead the reap_workers call there and let it async?

That's what I made this PR to do. The current way we handle that is with the signal queue and a wakeup. If we are afraid of losing the signal, we can either have a separate flag for CHLD or always append the CHLD signal no matter how long the queue is.

gunicorn/arbiter.py
tilgovi: Also, this is called from `murder_workers`. By removing this code, we remove the concurrent modification and looping, and we no longer need to call `list()`.

tilgovi:
> Can you clarify what is it trying to solve? Do we have a way to reproduce the issue?

Just code cleanup. It avoids the need to call `list()` on the workers list when we loop over it because it removes the concurrent modification, by removing duplicate code.

tilgovi:
I reworked this with a separate indicator flag that the arbiter uses to mark that it needs to reap workers in the next main loop iteration. This change ensures that the arbiter will always reap workers even if the signal queue is full. I also added better explanation for why it is good to re-install the handler.

Please take another look.

benoitc:
The use of use of `list(self.WORKERS.items()` was mainly there because `.items()` was not considered safe. When a worker die the list is updated concurrently. 

Anyway the patch looks fine. I need to test it on BSDs systems, in particularly OpenBSD where I remember the special handling of the CHLD signal was added for. 

Can it wait next release (let's make it next month?) ?

tilgovi:
> The use of use of list(self.WORKERS.items() was mainly there because .items() was not considered safe. When a worker die the list is updated concurrently.

Not any more because that's what this patch changes!

> Anyway the patch looks fine. I need to test it on BSDs systems, in particularly OpenBSD where I remember the special handling of the CHLD signal was added for.

I don't think any BSD has this problem. From what I read it's mos likely some versions of Solaris.[1] In any case, setting the handler when it is already set is not a problem, so I'm not worried about regression here. The only possible change here would to fix behavior on a system where Gunicorn was previously broken.

> Can it wait next release (let's make it next month?) ?

Absolutely.

[1]: http://poincare.matf.bg.ac.rs/~ivana/courses/ps/sistemi_knjige/pomocno/apue/APUE/0201433079/ch10lev1sec7.html

gunicorn/arbiter.py
```diff
@@ -186,10 +187,11 @@ def init_signals(self):
...
```
tilgovi, line 189: This part is a mistake. We still need to install this handler separately, and it should not be in the list of `SIGNALS` above, if we are not going to use the queue for it.

tilgovi: Sorry, that's wrong. The point of this patch is to handle reaping in the main thread.

What's missing is a check for `self.SIGCHLD` in the `run` method, with a call `self.handle_chld` from the main loop.

gunicorn/arbiter.py
```diff
@@ -186,10 +187,11 @@ def init_signals(self):
...
 
     def signal(self, sig, frame):
-        if len(self.SIG_QUEUE) < 5:
+        if sig is signal.SIGCHLD:
+            self.SIG_CHLD = True
```
tilgovi, line 193: One consequence of not using the signal queue, as compared to the old implementation, is that we're not reaping immediately. We could call `self.wakeup()` here.

Alternatively, is there any reason for the limit, below, to 5 signals in the queue? If not, we could have no special case for `SIGCHLD`.

tilgovi:
@benoitc would you take another look here, please? I think it's a good idea to reap on the main thread. I think this approach is easier to understand, safer, reduces duplicate code, and eliminates concurrent modification of the workers list.

gunicorn/arbiter.py
```diff
@@ -519,6 +535,12 @@ def reap_workers(self):
                 if self.reexec_pid == wpid:
                     self.reexec_pid = 0
                 else:
+                    worker = self.WORKERS.pop(wpid, None)
+                    if not worker:
+                        continue
+
+                    worker.tmp.close()
```
tilgovi, line 542: I thought it was best to ensure that we actually only do anything below here if this is actually our worker. But I can make this a separate PR if that's desired.

gunicorn/arbiter.py
```diff
@@ -42,10 +42,10 @@ class Arbiter(object):
     # I love dynamic languages
     SIG_QUEUE = []
     SIGNALS = [getattr(signal, "SIG%s" % x)
-               for x in "HUP QUIT INT TERM TTIN TTOU USR1 USR2 WINCH".split()]
+               for x in "CHLD HUP QUIT INT TERM TTIN TTOU USR1 USR2 WINCH".split()]
     SIG_NAMES = dict(
         (getattr(signal, name), name[3:].lower()) for name in dir(signal)
-        if name[:3] == "SIG" and name[3] != "_"
+        if name[:3] == "SIG" and name[3] != "_" and name[3:] != "CLD"
```
piskvorky, line 48: This condition gave me a pause, not obvious.

```suggestion
        if name[:3] == "SIG" and name[3] != "_" and name[3:] != "CLD"  # SIGCLD is an obsolete name for SIGCHLD
```

At least according to https://www.gnu.org/software/libc/manual/html_node/Job-Control-Signals.html

benoitc:
@tilgovi I will check this patch. But one of the reason we handled CHLD differently is that montoring  is separate from handling signals targetting the process itself. Having all in in the same signal loop means that this signal wil be in the same queue.

gunicorn/arbiter.py
tilgovi: I thought I encountered an error without this (on Linux) but I'll double check.

sylt:
I wasn't aware of this pr until now, but FYI, I've posted #3148 which also aims to handle SIGCHLD on the main thread, although in a slightly different way.

piskvorky:
I've been running & testing #3148 (@sylt's PR), so far without problems. It seems more complete – perhaps the two PRs could be merged, gaining the benefit of both?

And then there's https://github.com/benoitc/gunicorn/pull/2908/files too of course.

gunicorn/arbiter.py
piskvorky: @sylt's solution in #3148 seems a bit cleaner.

tilgovi:
> @tilgovi I will check this patch. But one of the reason we handled CHLD differently is that montoring  is separate from handling signals targetting the process itself. Having all in in the same signal loop means that this signal wil be in the same queue. 
> 
> 

We could put the SIGCHLD signal on the front of the queue, if what you want is just to handle it with higher priority, but I think it's a good idea to make the handler itself short and defer reaping to the main thread.

I also had a version of this with just a boolean flag for whether to reap on wakeup, if for some reason you really don't want SIGCHLD in the queue at all. But I don't think I understand in what way this signal is for "monitoring" or different from "signals targeting the process itself."

Can you explain anymore what concerns you?

tilgovi:
I see on the other PR that it seems like one hesitation you have is that you don't want to wait for children when exiting Gunicorn, but that already happens. While we reap children in the signal context, handling other signals is blocked.

Issue #1937, opened 2018-12-17
Workers silently exit when allocating more memory than the system allows.


jonathanlunt:
Problem Description
--------------------
Gunicorn workers silently exit when allocating more memory than the system allows. This causes the master to enter an infinite loop where it keeps booting new workers unsuccessfully. No hook/logging exists to detect or handle this behavior.

```
[2018-12-16 23:29:45 +0000] [1] [INFO] Starting gunicorn 19.9.0
[2018-12-16 23:29:45 +0000] [1] [INFO] Listening at: http://127.0.0.1:8080 (1)
[2018-12-16 23:29:45 +0000] [1] [INFO] Using worker: sync
[2018-12-16 23:29:45 +0000] [10] [INFO] Booting worker with pid: 10
post_fork 10
child_exit 10
[2018-12-16 23:29:46 +0000] [11] [INFO] Booting worker with pid: 11
post_fork 11
child_exit 11
[2018-12-16 23:29:47 +0000] [12] [INFO] Booting worker with pid: 12
post_fork 12
child_exit 12
[2018-12-16 23:29:47 +0000] [13] [INFO] Booting worker with pid: 13
post_fork 13
child_exit 13
```

Files
-----
These file will allow you to to reproduce the behavior (clone-able version here: https://github.com/jonathanlunt/gunicorn-memory-example). `docker` is used to artificially constrain system resources.

**app.py**: Generic "hello world" from gunicorn documentation
```python
def app(environ, start_response):
    data = b"Hello, World!\n"
    start_response("200 OK", [
        ("Content-Type", "text/plain"),
        ("Content-Length", str(len(data)))
    ])
    return iter([data])
```

**config.py**: This includes a `post_fork` function that allocates 100MB
```python
bind = '%s:%s' % ('127.0.0.1', '8080')
workers = 1

def post_fork(server, worker):
    """Allocate ~100MB of integers"""
    print('post_fork', worker.pid)
    items = list(range(4300800))

def child_exit(server, worker):
    print('child_exit', worker.pid)
```

**Dockerfile**
```dockerfile
FROM python:3.6-alpine3.8

RUN pip install gunicorn

COPY *.py /opt/
WORKDIR /opt

ENTRYPOINT ["gunicorn", "-c", "config.py", "app:app"]
```

Usage
------
The run command limits the container to allow only 50MB in memory
```
docker build -t gunicorn-example .
docker run -m 50mb -t gunicorn-example
```

Proposed Solutions
[...]

PR #2315, opened 2020-04-21
More verbose exit logging

tilgovi:
- Add logging at error level for abnormal worker exit codes.
- Make arbiter shutdown log at error level for abnormal exits.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -520,6 +520,8 @@ class Arbiter(object):
                     # that it could not boot, we'll shut it down to avoid
                     # infinite start/stop cycles.
                     exitcode = status >> 8
+                    if exitcode != 0:
+                        self.log.error('Worker (pid:%s) exited with code %s', wpid, exitcode)
                     if exitcode == self.WORKER_BOOT_ERROR:
                         reason = "Worker failed to boot."
                         raise HaltServer(reason, self.WORKER_BOOT_ERROR)
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -340,9 +340,12 @@ class Arbiter(object):
     def halt(self, reason=None, exit_status=0):
         """ halt arbiter """
         self.stop()
-        self.log.info("Shutting down: %s", self.master_name)
+
+        log_func = self.log.info if exit_status == 0 else self.log.error
+        log_func("Shutting down: %s", self.master_name)
         if reason is not None:
-            self.log.info("Reason: %s", reason)
+            log_func("Reason: %s", reason)
+
         if self.pidfile is not None:
             self.pidfile.unlink()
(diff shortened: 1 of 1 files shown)


nakulcg:
I installed gunicorn from this branch, but I dont see 
```
'Worker (pid:%s) exited with code %s'
```
but I do see 
```
Out of memory: Kill process 14666 (gunicorn) score 319 or sacrifice child
```
in dmesg in debian buster container

gunicorn/arbiter.py
```diff
@@ -340,9 +340,12 @@ def wakeup(self):
     def halt(self, reason=None, exit_status=0):
         """ halt arbiter """
         self.stop()
-        self.log.info("Shutting down: %s", self.master_name)
+
+        log_func = self.log.info if exit_status == 0 else self.log.error
+        log_func("Shutting down: %s", self.master_name)
         if reason is not None:
-            self.log.info("Reason: %s", reason)
+            log_func("Reason: %s", reason)
```
benoitc, line 347: shouldn't we keep the line "Shutting down ..."  ?

benoitc (changes requested):

gunicorn/arbiter.py
tilgovi: I did. It moved above the if statement so that it's always logged.

tilgovi: Oh, sorry, it was always logged in both cases. I moved it down so that it logs at error level when the status is non-zero.

tilgovi:
@benoitc any changes you still want here? I responded to your earlier feedback.

kosciCZ:
Bump, I'd really like to see this merged.

tilgovi:
@benoitc can you please take another look at this one?

harto:
Hello, this would be useful for us too :)

sp1rs:
@tilgovi Should it pass the `exitcode` to `child_exit` function? This will give client more flexibility to play around. In my case I would like to log the exitcode as well as based on certain logic also predict if it is OOM or not.
Also I like the [proposal](https://github.com/benoitc/gunicorn/issues/1937#issue-391520093) to add the `exit code` to `worker`.

gunicorn/arbiter.py
```diff
@@ -340,9 +340,12 @@ def wakeup(self):
     def halt(self, reason=None, exit_status=0):
         """ halt arbiter """
         self.stop()
-        self.log.info("Shutting down: %s", self.master_name)
+
+        log_func = self.log.info if exit_status == 0 else self.log.error
+        log_func("Shutting down: %s", self.master_name)
```
benoitc, line 345: why not simply adding the log error as a second line ? I think displaying "shutting down" in info logs is useful. (sorry to not have take that earlier).

benoitc (changes requested):

Issue #2285, opened 2020-03-07
[Contribution]Example : Integration of Gunicorn with Kafka & Gevent


mohammedvaghjipurwala:
Example to demonstrate the integration of Gunicorn with Kafka streaming and its worker as gevent.

I have not found any relevant documentation for these technologies.

I had integration issues like the consumer getting hanged, Kafka connectivity issues, etc. I finally resolved this issue and I would like to contribute this solution with an example to this repository.

tilgovi:
That would be great. Please open a PR to add something to the `examples` directory and we'll review it!

mohammedvaghjipurwala:
Thanks @tilgovi , I have created an MR for the Example.
https://github.com/benoitc/gunicorn/pull/2316

benoitc:
no activity since awhile. closing feel free to create a new ticket if needed.

PR #2316, opened 2020-04-21
Example : Integration of Gunicorn with Kafka & Gevent

mohammedvaghjipurwala:
#2285 
Example added for Gunicorn implementation with Kafka & Gevent.

  This Example demonstrates the Kafka consumer & producer with Gunicorn and worker as Gevent
  Monkey Patching the Gevent will create a continuous Kafka consumer and resolve the issue of the consumer getting idle.

### DIFF
--- a/examples/frameworks/flaskapp_gevent_kafka.py
+++ b/examples/frameworks/flaskapp_gevent_kafka.py
@@ -34,7 +34,7 @@ def read_kafka():
     for message in get_consumer():
         message = message.value
         sent = producer.send('test_producer',
-                             dumps(message).encode())
+                             message)
         return sent.get()
 
 
--- a/examples/frameworks/flaskapp_gevent_kafka.py
+++ b/examples/frameworks/flaskapp_gevent_kafka.py
@@ -16,20 +16,84 @@
 
 from json import loads, dumps
 from flask import Flask
-from gevent import monkey
+from gevent import pool
 from kafka import KafkaConsumer, KafkaProducer
 
-app = Flask(__name__)
+APP = Flask(__name__)
 
-monkey.patch_all(socket=False, thread=False)
+WORKER_THREADS = 10
+POOL = pool.Pool(WORKER_THREADS)
 
-@app.route("/")
+@APP.route("/")
 def hello():
+    """
+    ########################################
+    ## Method to flask app
+    ########################################
+    """
     return "Hello World!"
 
 
-@app.route("/kafka/start-reading")
+def get_status(green_thread):
+    """
+    ########################################
+    ## Method to get the status of greenlet
+    ########################################
+    """
+    total = 0
+    running = 0
+    completed = 0
+    succeeded = 0
+    yet_to_run = 0
+    failed = 0
+    total += 1
+    if bool(green_thread):
+        running += 1
+    elif green_thread.ready():
+        completed += 1
+        if green_thread.successful():
+            succeeded += 1
+        else:
+            failed += 1
+    else:
+        yet_to_run += 1
+
+    return dict(total=total,
+                running=running,
+                completed=completed,
+                succeeded=succeeded,
+                yet_to_run=yet_to_run,
+                failed=failed)
+
+@APP.route("/kafka/start-reading")
+def process_records():
+    """
+    ########################################
+    ## Starter method to trigger the reader
+    ## & producer method in async mode using
+    ## greenlets
+    ########################################
+    """
+    # Creating green threads(Greenlets)
+    greenlet = POOL.apply_async(read_kafka)
+    response = get_status(greenlet)
+    print(response)
+    while True:
+        response = get_status(greenlet)
+        print(response)
+        if response['running'] == response['total']:
+            break
+        print("retrying...")
+    return_str = "Greenlet started Successfully !"
+    return return_str
+
 def read_kafka():
+    """
+    ########################################
+    ## Method to read data from topic and
+    ## produce on another topic
+    ########################################
+    """
     producer = get_producer()
     for message in get_consumer():
         message = message.value
@@ -39,6 +103,11 @@ def read_kafka():
 
 
 def get_consumer():
+    """
+    ########################################
+    ## Method to generate consumer object
+    ########################################
+    """
     return KafkaConsumer(
         'test_consumer',
         bootstrap_servers=['localhost:9092'],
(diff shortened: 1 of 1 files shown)


examples/frameworks/flaskapp_gevent_kafka.py
```diff
@@ -0,0 +1,53 @@
...
+    return "Hello World!"
+
+
+@app.route("/kafka/start-reading")
+def read_kafka():
+    producer = get_producer()
+    for message in get_consumer():
+        message = message.value
+        sent = producer.send('test_producer',
+                             json.dumps(data).encode())
```
tilgovi, line 37: `data` is not bound here.

examples/frameworks/flaskapp_gevent_kafka.py
```diff
@@ -0,0 +1,53 @@
...
+###############################
+
+from flask import Flask
+from gevent import monkey
+from json import loads, dumps
+from kafka import KafkaConsumer, KafkaProducer
+
+app = Flask(__name__)
+
+monkey.patch_all(socket=False, thread=False)
```
jamadden, line 24: I would be reluctant to recommend passing any non-default arguments to `monkey.patch_all()`.

You write:
>  I found that Gevent blocks consumers from reading messages continuously. So Monkey patching Gevent with disabling socket & thread.

Can you elaborate more on what you saw? That's not how gevent should behave, at all. Something like that is often a sign of monkey-patching too late, or a conflict with native code. (Here, we're definitely patching later than recommended as there are numerous imports before the patch; that's a [general problem in gunicorn too](https://github.com/benoitc/gunicorn/issues/1566), at least in some configurations.) Does gevent's [blocking monitor](http://www.gevent.org/monitoring.html#blocking) report anything when you see this behaviour?

examples/frameworks/flaskapp_gevent_kafka.py
```diff
@@ -0,0 +1,53 @@
...
+                    'test_consumer',
+                     bootstrap_servers=['localhost:9092'],
+                     auto_offset_reset='earliest',
+                     enable_auto_commit=True,
+                     group_id='my-group',
+                     value_deserializer=lambda x: loads(x.decode('utf-8')))
+
+def producer():
+    return KafkaProducer(bootstrap_servers=['localhost:9092'],
+                         value_serializer=lambda x:
```
tilgovi, line 52: I'm not familiar with the kafka library for Python,  but does this serializer mean that you can directly call `producer.send(data)` rather than `producer.send(json.dumps(data).encode())`?

examples/frameworks/flaskapp_gevent_kafka.py
tilgovi: `dumps(data)` and not `json.dumps(data)` since there is `from json import loads, dumps`

examples/frameworks/flaskapp_gevent_kafka.py
mohammedvaghjipurwala: Agreed. I have changed it to 
`
producer.send(data)
`

examples/frameworks/flaskapp_gevent_kafka.py
mohammedvaghjipurwala: Hi @jamadden , 

I'm working on a product that fetches near to realtime data(statistics) from the server/3rd party tools, parse the input data, run some AI analytics on it and provide insights & solutions.

For real-time data transmission between the different services, we are using Kafka messaging broker.
When we were running in the sync mode the transmission of data was flawless. However, for performance benefits, we added Gevent as our worker on Gunicorn.
Post this, we started seeing that the consumer read only one message and then stopped and moved to the hanged state.

After some research, we found that the Gevent was blocking the consumer thread to read the messages continuously. Hence, we implemented Monkey patching disabling socket & thread.

This solution worked like a charm and resolved this issue permanently !!!

jamadden: I'm sorry, I wasn't clear. I'm the maintainer of gevent, and I wasn't asking for details about your system, I was asking for details about your research and conclusion. Exactly *how* did you determine that "that the Gevent was blocking the consumer thread to read the messages continuously" and *why* was it that "disabling socket & thread" was a solution? As I said, the behaviour you described is not something that gevent should ever do; if it does, it's a misconfiguration, an incompatibility, or a bug. So I'm trying to understand which of those it is, and exactly what's going on.

Disabling monkey-patching of threads and sockets should rarely ever be a solution, and I don't want to recommend it without knowing a lot more than we do — that's almost equivalent to simply not monkey-patching at all. Those are usually the two most important things to monkey-patch in existing code; without those, all network operations will be blocking and things usually just get worse. (Unless, of course, you're directly using gevent APIs like `gevent.socket`, but that's not in this example.) 

Besides, when your WSGI app is imported into gunicorn, gunicorn is going to run `monkey.patch_all()` *again*, and that's going to patch, you guessed it, threads and sockets. (Most likely with a warning about using `patch_all` twice with different arguments.)

As a guess, a problem like you described can happen when there is native code (a C extension) involved that directly makes blocking operating system calls; gevent cannot patch those, and they don't cooperate with the event loop, so things might appear to hang. Putting the calls to that code into a native thread can solve the problem. But the way to do that is usually not to avoid monkey-patching thread, it's to use the threadpool, and/or queues and/or async watchers. Is there any evidence that there is native code doing the blocking?

mohammedvaghjipurwala: Apologies for misunderstanding you query, The solution was implemented in the product almost 5-7 months back. Hence I donot have the exact logs at that moment.However, from what i recall we added debug loggers at kafka level first and then got to know that the after reading the first message it moved to the hang state it was able to get the heart_beat but no messages were consumed.
I forgot to add on that we were using async threads for starting multiple Consumers
After this, we tried in sync mode again and it worked without any glitches.
Then, we checked on the internet for why is it being blocked. we found few articles from which we assummed that gevent might be blocking the consumer thread.
So, we started the research to resolve the blocking issue and found some url which mentioned using `monkey.patch_all(socket=False, thread=False)`. This is how the issue got solved. Later we never tried changing this settings.

we then switched the async threads with greenlets without removing the patch_all and we are not seeing the error of consumer getting hanged.

As per your recommendation, we did not check for any blocking OS calls however, I would try again without monkey patching since we are already using the using greenlets.

mohammedvaghjipurwala: @jamadden 

Apologies for the delay. As per your recommendation, I have removed the monkey patch from the code as we already have the greenlets for async processing.
I have updated the file with the above changes.

Could you please review and merge. 

Thanks

benoitc:

benoitc:
a late thanks for your contribution . I am closing that change as i since this is a little bit outdated

PR #2325, opened 2020-05-01
Upper bound gevent to <20.5.0

darwinyip:
#2324


jamadden:
Thanks for the PR. As indicated in the referenced issue, this is a problem with the particular system, not gevent+gunicorn.

Issue #2335, opened 2020-05-20
gunicorn + gevent + statsd causes errors.


elendiastarman:
Reposting https://github.com/benoitc/gunicorn/issues/1673 as it does not seem to be fixed.

benoitc:
no activity since awhile. closing feel free to create a new ticket if needed.

PR #2336, opened 2020-05-20
Fixed two bugs related to gevent + gunicorn + statsd.

elendiastarman:
Fixes https://github.com/benoitc/gunicorn/issues/2335. Changes are simply decoding `status` (when logging access for statsd or logging requests for gevent) from bytes to str if necessary. Tested this locally with Python 3.7 and gevent 1.4.0, and it works just fine.


benoitc:
thanks!

PR #2344, opened 2020-05-29
✨ Add support for python -m gunicorn

tiangolo:
✨ Add support for `python -m gunicorn`.

This just adds a simple `__main__.py` replicating the script defined in `setup.py`.

That adds support for `python -m gunicorn`, as described [here in the Python docs](https://docs.python.org/3.8/library/__main__.html).

It is useful when `python` or equivalent is available as a command but `gunicorn` is not, even though it is installed and available to import. And where the idea is still to call it from the CLI and not importing and running from code... :man_shrugging: 

I know that sounds contrived, but the specific use case is Pex: https://pex.readthedocs.io/

It is based on "zip applications" from this [PEP 441](https://www.python.org/dev/peps/pep-0441/).

So, it would be called like:

```console
$ ./some-binary-pex-file -m gunicorn
```

### DIFF
--- a/gunicorn/__main__.py
+++ b/gunicorn/__main__.py
@@ -1,0 +1,7 @@
+# -*- coding: utf-8 -
+#
+# This file is part of gunicorn released under the MIT license.
+# See the NOTICE for more information.
+
+from gunicorn.app.wsgiapp import run
+run()


jamadden:
I think this is a great idea too!

Should this be documented somewhere?

tilgovi:
Good point, @jamadden :)

tiangolo:
Thanks! :rocket: 

---

About documenting it, let me know if you would like it to be explicitly stated, probably as a small note somewhere. Right now it would just work like e.g. `pip` or `python -m pip`.

I also understand that `python -m gunicorn` would not be the "canonical form" to call it, and I admit it is a rather "exotic" use case :sweat_smile: , so you might also want to just leave it out of the docs, just to not confuse users (I've done that in [FastAPI](https://github.com/tiangolo/fastapi) :man_shrugging: ), but let me know if you prefer to have it explicit somewhere.

---

About tests, I didn't find something to test the CLI behavior to try and include this there, the option I would think of would be a subprocess calling Python directly and checking the string from something like:

```console
python -m gunicorn --version
__main__.py (version 20.0.4)
```

That's what I do for testing some corner cases in [Typer](https://github.com/tiangolo/typer). But that could mean an extra layer of complexity and a bit of time increase for the tests (probably just a couple of secs though). If that sounds fine for you, I can sketch a PR with it. On the other side, do you have an idea of a better/different approach that you would prefer?

di:
+1 to documenting this somewhere, at the very least in [the changelog](https://github.com/benoitc/gunicorn/blob/master/docs/source/news.rst) (but it looks like this just gets updated on release?)

PR #2346, opened 2020-06-01
Unread method must add data to the beginning

mapogolions:
Steps to reproduce:
-  `git checkout master`
- go to the [line](https://github.com/benoitc/gunicorn/blob/4bed9e7b192ae211b7d1f6c14065c373c9e1c0aa/gunicorn/http/body.py#L103) and change it like this ```sh data = unreader.read(8)```
- `make test`
On my machine, the tests failed

### DIFF
--- a/gunicorn/http/unreader.py
+++ b/gunicorn/http/unreader.py
@@ -50,8 +50,14 @@ class Unreader(object):
         return data[:size]
 
     def unread(self, data):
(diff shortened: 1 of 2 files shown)


benoitc:
The underlying bug was fixed in commit 0032c797 (via a cleaner implementation from #2915). Thank you for identifying and documenting this issue!

PR #2348, opened 2020-06-01
Proposed as a solution for issue #1658

rbwinslow:
A new timeout-start-after value is added to config, granting workers an
initial grace period to do time-consuming initialization, after which
they are held to timeout stricture.

Previous attempts of mine to form an acceptable pull request have failed
linting in CI, all in code I didn't touch. This branch is newly minted
from the current upstream master.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -1363,6 +1363,23 @@ For the non sync workers it just means that the worker process is still
 communicating and is not tied to the length of time required to handle a
 single request.
 
+.. _timeout-start-after:
+
+timeout_start_after
+~~~~~~~~~~~~~~~~~~~
+
+* ``--timeout-start-after INT``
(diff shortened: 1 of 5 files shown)


benoitc:
why not simply increasing the timeout if you know it will take longer on startup?

rbwinslow:
We need the timeout value to be small because we want its strict enforcement of request duration after initialization is done and the worker is handling requests. We just need a longer period at startup, as each worker sets up process state it will need to process requests.

benoitc:
thanks for the patch. closing as won't do. The reasonning is that we already give a graceful time to close the worker. Once  exit is sent workers must be closed asap.

rbwinslow:
The grace time wasn't at closure, it was at startup. Some workers need more
time just at the start, to initialize state that is worker-specific.

PR #2353, opened 2020-06-05
If inotify is not installed the error is opaque. Fix that.

### LINKED ISSUE #2352: Correct misleading error with --reload-engine
When running with the `--reload-engine inotify` flag and without having previously installed `inotify` the error message "You must have the inotify module installed to use the inotify reloader" should show up to quickly reorient the user. However, it actually fails with the opaque and misleading error:

```
  File "/usr/local/lib/python3.7/site-packages/gunicorn/arbiter.py", line 583, in spawn_worker
    worker.init_process()
  File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/base.py", line 133, in init_process
    callback=changed)
TypeError: __init__() got an unexpected keyword argument 'extra_files'
```

This is because of the parameter mismatch from the call site [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L131), to the fake supposed-to-error class [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/reloader.py#L120).

Given that the fake `InotifyReloader` class has one purpose - to raise that error message to the user - the function spec could be changed to the below, which will work, and will be future-proof under changes to the real `InotifyReloader` class's spec changes.

```python
def __init__(self, *args, **kwargs):
```

dunk:
As per the ticket, running `--reload-engine inotify` without inotify installed is supposed to cause an obvious error, but due to the different function specs on the `__init__` it just blows up with an error. This fix makes the actual error visible in the output, and future-proofs the init of the fake `InotifyReloader`.

Fixes #2352

### DIFF
--- a/tests/test_reload.py
+++ b/tests/test_reload.py
@@ -1,4 +1,6 @@
+import pytest
 import unittest.mock as mock
+
 
 from gunicorn.app.base import Application
 from gunicorn.workers.base import Worker
@@ -66,3 +68,24 @@ def test_start_reloader_after_load_wsgi():
         mock.call.load_wsgi(),
         mock.call.reloader_start(),
     ])
+
+
+def test_reload_errors_with_import_error_when_inotify_not_installed():
+    reloader = mock.Mock()
+    # NOTE: Don't override the reloader_engines, or the test does nothing!
+    # reloader_engines['inotify'] = lambda *args, **kw: reloader
+
+    app = ReloadApp()
+    cfg = app.cfg
+    log = mock.Mock()
+    worker = MyWorker(age=0, ppid=0, sockets=[], app=app, timeout=0, cfg=cfg, log=log)
+
+    cfg.set('reload_engine', 'inotify')
+
+    worker.load_wsgi = mock.Mock()
+    mock_parent = mock.Mock()
+    mock_parent.attach_mock(worker.load_wsgi, 'load_wsgi')
+    mock_parent.attach_mock(reloader.start, 'reloader_start')
+
+    with pytest.raises(ImportError):
+        worker.init_process()
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -118,7 +118,7 @@ if has_inotify:
 else:
 
     class InotifyReloader(object):
-        def __init__(self, callback=None):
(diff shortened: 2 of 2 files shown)


dunk:
I just force pushed to rename the test, which I had left as some temporary thing by accident. I think that there is a small risk with this one, depending on who is running the tests and how they run them. inotify isn't in the requirements, so any normal, clean test env would work correctly. If someone tested in an unclean environment - like on their local machine, or perhaps in a virtualenv that they have set up to run the code, then because the test relies on actual import behaviour, the test would operate differently for them. I tried to fix this by overriding the import behaviour at test time, but that's not possible because this code uses the parse-time to actually run the code, so the tests can't inject the appropriate mock. That should probably be fixed, but it's more systemic and project-wide. `importlib.util.find_spec` could be used for that purpose, as it allows the code to programmatically detect what packages are present, without trying and catching exceptions.

tests/test_reload.py
```diff
@@ -66,3 +68,24 @@ def test_start_reloader_after_load_wsgi():
...
+    cfg = app.cfg
+    log = mock.Mock()
+    worker = MyWorker(age=0, ppid=0, sockets=[], app=app, timeout=0, cfg=cfg, log=log)
+
+    cfg.set('reload_engine', 'inotify')
+
+    worker.load_wsgi = mock.Mock()
+    mock_parent = mock.Mock()
+    mock_parent.attach_mock(worker.load_wsgi, 'load_wsgi')
+    mock_parent.attach_mock(reloader.start, 'reloader_start')
```
trendels, line 88: These lines are not needed here (probably copied over from the previous test case).

humphrey:
Could we update the documentation for this also, just to make it clear that it's the Python inotify package that needs to be installed?  Since inotify is a system thing, I misread that as having to have an apt-get package installed, and it wasn't until I read through the source code that I realised it was a python package.

At https://docs.gunicorn.org/en/latest/settings.html#reload

I suggesting a simple change to the following line from:

> In order to use the inotify reloader, you must have the inotify package installed.

to 

> In order to use the inotify reloader, you must have the python [inotify](https://pypi.org/project/inotify/) package installed.

benoitc:

PR #2354, opened 2020-06-08
Typo/grammar in Deploying Gunicorn docs

ssanjaysubramaniam:


tilgovi:
This change is probably a minor improvement. Both forms are common, but I think the "to + infinitive" form may be better only with a noun phrase.

"We recommend *you* to use Gunicorn behind a proxy server."
"We recommend using Gunicorn behind a proxy server."

https://www.grammaring.com/to-infinitive-or-gerund-advise-recommend-allow-permit-forbid-require

PR #2376, opened 2020-07-11
Update docs to mention nginx's proxy_http_version

dgilman:
Nginx added the `proxy_http_version` option to enable HTTP 1.1 proxy requests. The option was added in nginx 1.1.4, released on September 20th, 2011 so hopefully all of your users have upgraded by now.

The FAQ entry was written maybe a year before nginx 1.1.4 was released, so although it was correct at the time it is no longer relevant and I've taken it out. I also updated the example proxy configurations to enable HTTP 1.1.

Hopefully this will save some poor hacker some troubleshooting down the road when they've configured Nginx by copying out of the example from the docs, used an async worker and configured gunicorn's keepalive support and can't figure out why keepalive still isn't working.

### DIFF
--- a/docs/source/faq.rst
+++ b/docs/source/faq.rst
@@ -43,13 +43,15 @@ in tools like ``ps`` and ``top``. This helps for distinguishing the master
 process as well as between masters when running more than one app on a single
 machine. See the proc_name_ setting for more information.
 
-Why is there no HTTP Keep-Alive?
---------------------------------
+Why is there no HTTP Keep-Alive header?
+---------------------------------------
 
-The default Sync workers are designed to run behind Nginx which only uses
-HTTP/1.0 with its upstream servers. If you want to deploy Gunicorn to
-handle unbuffered requests (ie, serving requests directly from the internet)
-you should use one of the async workers.
+You need to enable Keep-Alive support in both gunicorn and your load balancer.
+To enable Keep-Alive support in gunicorn you must use an async worker class,
+which is any of the worker classes that aren't the default sync class. To
+enable Keep-Alive with nginx you must enable its HTTP 1.1 support. Add
+``proxy_http_version 1.1;`` to your nginx configuration to enable HTTP 1.1
+proxy requests.
 
 .. _Hey: https://github.com/rakyll/hey
 .. _setproctitle: https://pypi.python.org/pypi/setproctitle
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -29,6 +29,7 @@ To turn off buffering, you only need to add ``proxy_buffering off;`` to your
 
   ...
   location @proxy_to_app {
+      proxy_http_version 1.1;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header Host $http_host;
       proxy_redirect off;
--- a/examples/nginx.conf
+++ b/examples/nginx.conf
@@ -56,6 +56,7 @@ http {
     }
 
     location @proxy_to_app {
+      proxy_http_version 1.1;
       proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
       proxy_set_header X-Forwarded-Proto $scheme;
       proxy_set_header Host $http_host;


benoitc:
I would rather see the warning documentation message changed for something eventually letting the users know they can use this option since it's not the default in NGINX.  Can you change your PR accordingly?

tilgovi:
I agree with Benoit. Keep the update to the example, but modify the FAQ to reflect that async workers support keep-alive while sync workers do not. Running behind nginx, async workers need proxy_http_version to support keep-alive connects to the gunicorn upstream.

dgilman:
Review feedback has been addressed, can we get this merged?

docs/source/faq.rst
```diff
@@ -43,13 +43,13 @@ in tools like ``ps`` and ``top``. This helps for distinguishing the master
 process as well as between masters when running more than one app on a single
 machine. See the proc_name_ setting for more information.
 
-Why is there no HTTP Keep-Alive?
+Why is there no HTTP Keep-Alive header?
 --------------------------------
```
jamadden, line 47:
```suggestion
Why is there no HTTP Keep-Alive header?
---------------------------------------
```

The adornment needs to match the length of the title, otherwise there's warnings from the document processors.

jamadden (commented):
I don't see anything that addressed @tilgovi's last comment about keeping the information about sync vs async workers. That still seems to be removed.

dgilman:
One last round of reviews?

docs/source/faq.rst
```diff
@@ -43,13 +43,15 @@ in tools like ``ps`` and ``top``. This helps for distinguishing the master
...
 
-The default Sync workers are designed to run behind Nginx which only uses
-HTTP/1.0 with its upstream servers. If you want to deploy Gunicorn to
-handle unbuffered requests (ie, serving requests directly from the internet)
-you should use one of the async workers.
+You need to enable Keep-Alive support in both gunicorn and your load balancer.
+To enable Keep-Alive support in gunicorn you must use an async worker class,
+which is any of the worker classes that aren't the default sync class. To
+enable Keep-Alive with nginx you must enable its HTTP 1.1 support. Add
+``proxy_http_version 1.1;`` and ``proxy_set_header Connection "";`` to your
```
tilgovi, line 53: Is it definitely good to add `proxy_set_header Connection ""`? Downstream clients should not be sending `Connection: close` if they want to make a keep-alive request. If the advice is just about how to get keep-alive between a reverse proxy and Gunicorn, then this makes sense, but it also prohibits `Connection: upgrade` for websockets requests, doesn't it?

tilgovi:
@dgilman just catching up on Gunicorn issues, sorry to be slow again. This looks great, I just have one more question.

I am going to clear my notification clutter tonight, so hopefully I am more responsive in the future! :pray:

docs/source/faq.rst
dgilman: Possibly. This is getting out of my familiarity with HTTP, I don't know what a good suggestion would be here.

tilgovi: I think let's remove the `proxy_set_header` and I will merge this.

benoitc (commented):
we are loosing some information I. This pr. I think it’s good to keep the notion of unbuffered proxy and that the gunixorn sync work doesn’t handle keep alive by itself

tilgovi:
@dgilman I'm sorry it's been so long that this has been in review here. Do you want to make a small change to your update to incorporate the feedback from Benoit and we'll merge this. I think it's a good contribution and I'd like to merge it.

dgilman:
I do not understand benoitc's feedback. I think it would be better if a gunicorn maintainer could finish this out and find phrasing that is acceptable to everyone. This whole thing is a pretty minor change and you all have much more familiarity with it than I do.

benoitc:
> I do not understand benoitc's feedback. I think it would be better if a gunicorn maintainer could finish this out and find phrasing that is acceptable to everyone. This whole thing is a pretty minor change and you all have much more familiarity with it than I do.

I am speaking about this part : 

```
 If you want to deploy Gunicorn to
 handle unbuffered requests (ie, serving requests directly from the internet)
 you should use one of the async workers.

```

This is not only related to keepalive request. Any other kind of unbuffered requests or streams require an async worker. I agree it may need some clarification, but we shouldn't loose this information. What about `To enable Keep-Alive support , or to handle streams (websockets, or any unbuffered requests)  in gunicorn you must use an async worker class ...`

Issue #2374, opened 2020-07-09
Type stubs / Type annotations / mypy


MartinThoma:
I'm currently annotating types for my codebase. While doing so, I was wondering what the types of the parameters of [glogging.access](https://github.com/benoitc/gunicorn/blob/master/gunicorn/glogging.py) are:

    def access(self, resp, req, environ, request_time): ...

Maybe it would even be possible to add a type stub file?

tilgovi:
This has been covered a bit in #1393. I would welcome and champion contributions to add types to Gunicorn. The code is Python 3 only so there is nothing to prevent it.

tilgovi:
I'm going to close this because I believe the question of whether we would add a type stub (or types directly) has been answered already, but so far no one has taken this initiative. Thank you for bumping the issue, though. It's good to know that it's wanted.

foobacca:
For anyone who lands here looking for type hints, they are now in typeshed, as mentioned in https://github.com/benoitc/gunicorn/issues/3420

The direct link to the folder in typeshed, if you want to browse for types, is https://github.com/python/typeshed/tree/main/stubs/gunicorn/gunicorn

Issue #1393, opened 2016-11-17
mypy/PEP-484 annotations?


surrealcristian:
Are you interested in add mypy/PEP-484 annotations to gunicorn?

tilgovi:
I have no objection. It's an accepted PEP after all! I cannot speak for everyone, though.

tilgovi:
If we do this, we should do it in the [comment style](https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code) so that we maintain Python 2 compatibility.

benoitc:
what would be the benefits for the project?

tilgovi:
It functions as a form of documentation and mypy/pytype/pylint either support it or have plans to. It may help contributors this way. It also might work to gently discourage confusing behavior like functions returning multiple types depending on the code path, which I find more error prone and difficult to reason about.

Mostly, I see it as having no negative effect at all, and maybe some gentle positives.

benoitc:
if someone want to take care of it please let me know. Otherwise let's drop it, there is no need to keep an issue open if no one want to take care of it.

tilgovi:
I don't feel a need to take a hard stance like insisting new code has type annotations. However, now that we're on Python 3 only there's nothing preventing it, so we should accept code with proper type annotations. I may try to do it myself when making changes.

tilgovi:
Closed because I don't think it's realistic to expect that the current contributors will put in significant time to add types to code where it doesn't already exist. If anyone would like to make PRs to add types to any modules, please feel free.

PR #2377, opened 2020-07-12
Start adding type annotations to gunicorn

MartinThoma:
See issue #2374 and issue #1393


MartinThoma:
Hi guincorn developers! First of all: A big thank you for gunicorn 🤗 I've used it for several projects and never had to think a lot about that part :-) I'm currently annotating a code base and noticed that gunicorn does not have type annotations, sooo ... here you are :-) 

As I'm not familiar with the codebase, this will require some work and help of people who know gunicorn.

In case you want to read more about type annotations: [I wrote an article :-)](https://medium.com/analytics-vidhya/type-annotations-in-python-3-8-3b401384403d)

## How to get type annotations

This is how I would like to go forward and add types to gunicorn:

1. Make sure mypy doesn't crash anymore. *Decide on stub files vs in-line annotations.* (In the PR you see the in-line annotations example; let me know if you think stub files would be the better idea)
2. *Integrate mypy in CI pipeline*: Let me know if you prefer not to do that; note that this would only fail if mypy discovers a new problem - when new code without annotations is added, it is not checked and thus does NOT fail. Those settings can be adjusted, of course.
3. *Increase typed parts*
    * Try to get rid of "type[ignore]" / Any / untyped generics
    * Type untyped functions (currently, I mainly added annotations for the config.py)


## Impossible at the moment

* Using "real" annotations everywhere (variable annotations were introduced in Pyhton 3.6)
* Have "nice" annotations when a class returns itself (`from __future__ import annotations` was introduced in Python 3.7)
* Using TypedDict instead of Dict (introduced in Python 3.8)

MartinThoma:
I see a lot of linting errors. I'm used to `black`, but that would change quite some lines. Would that be ok? Do you have other autoformatters?

benoitc:
> I see a lot of linting errors. I'm used to the [`black` formatter](https://github.com/psf/black). Would it be ok to apply it? Do you have other autoformatters?

No autoformatter is and will be used. If you see any pep8 issue, let fix them in a separate commit :)

gunicorn/config.py
```diff
@@ -20,14 +20,21 @@
...
+KNOWN_SETTINGS = []  # type: List[Type[Setting]]
 PLATFORM = sys.platform
 
 
-def make_settings(ignore=None):
-    settings = {}
-    ignore = ignore or ()
+def make_settings(ignore=None) -> Dict[str, "Setting"]:
+    settings = {}  # type: Dict[Any, Any]
+    ignore = ignore or ()  # type: ignore
```
benoitc, line 37: where do you define the type ignore?

gunicorn/config.py
```diff
@@ -110,7 +117,7 @@ def worker_class_str(self):
         return uri
 
     @property
-    def worker_class(self):
+    def worker_class(self) -> Type[SyncWorker]:
```
benoitc, line 120: maybe better to use the Base worker class or create a type for it? this function will indeed return different type of worker class

gunicorn/config.py
MartinThoma: `type: ignore` tells mypy to ignore this line. I've added it in a couple of places if I didn't know directly why mypy was complaining. Those ignores should all be resolved at some point. For now, I first want to get this in general to work to get things started.

MartinThoma: In this specific case, very likely the problem is that the variable `ignore` first has some type, but if that value converts to bool it becomes a Tuple. Most of the time, one can resolve this by telling mypy all the possible types upfront: `type: Union[whatever it can be, Tuple]`

gunicorn/config.py
MartinThoma: Cool, I change that :-) Should I add it to this commit (and force-push) or would you rather like separate commits?

benoitc: at this stage i think it's always bette to add it as a commit. It will always be time later to fix the history.

benoitc:
> > I see a lot of linting errors. I'm used to the [`black` formatter](https://github.com/psf/black). Would it be ok to apply it? Do you have other autoformatters?
> 
> No autoformatter is and will be used. (one readon is to keep some flexibility). If you see any pep8 issue, let's fix them in a separate commit :)

I meant at the project level.  I need to publish some style rules so anyone can adapt their auto-formatter in a way we keep the code readable and not increasing that much the number of lines like some formatter do... If you use one in your env, just make sure to apply the change on a separate commit :)

gunicorn/config.py
benoitc: fair enough. Thanks for the explanation!

MartinThoma:
I applied `black` to the config.py to solve the linting issues and fixed the remaining issues. Black mostly has done two types of changes:

* Single quotes to double quotes
* The way of line breaks for long parametrized funtions

Current:

    def foo(parameter_a,
            parameter_b,
            parameter_c):
       body

Black:

    def a_function(
        parameter_a,
        parameter_b,
        parameter_c
    ):
        body

I think [yapf](https://github.com/google/yapf/issues/251) would support the current style, but I'm not familiar with yapf.

MartinThoma:
@benoitc I guess you are busy with other task, but just to make it clear: I'm waiting for feedback. From my side, this PR could be merged. Once it is merged, I will try to add mypy to the CI pipeline :-)

MartinThoma:
@benoitc Did you have the time to have a look at the PR?

tilgovi (commented):
I support the idea of adding type annotations. It would be nice to reduce the number of Any declarations. For example, the workers dict is `Dict[int, Worker]` and the pipe pair is a `[socket, socket]` tuple.

It is always preferable to separate style changes from other changes, so it would be nice to take the auto-formatting commit out of this PR. I support auto formatting, but we shouldn't apply it in this PR, and it sounds like @benoitc would need some more convincing.

MartinThoma:
@tilgovi I've added the more specific type annotations you mentioned and undid the changes black did to config. I've only applied black to that file and that is also where most of the type annotations were added.

MartinThoma:
@tiangolo Is there anything blocking this PR?

MartinThoma:
@benoitc Is there anything blocking this PR?

tilgovi:
I still like this idea and would love to add type annotations. I would prefer to use the annotation syntax, rather than comments, now that Gunicorn is Python 3 only. I'll ping people directly to see if I can get some consensus from folks about starting to add these types.

stdedos:
_Pls don't forget the `py.typed` marker_ 🙏 

Also, partial stuff (like merging this MR) is better than no stuf

Also related to https://github.com/benoitc/gunicorn/issues/2833?

stdedos:
That is ... very respected ❤️

benoitc:
closing this issue. While I appreciate the effort  there won't be type annotation in gunicorn.

Issue #2328, opened 2020-05-06
--access-logformat for iso date with microseconds


thenewguy:
The docs do not give a format option for iso date with microseconds.  Is this possible?

thenewguy:
Or any format with resolution greater or equal to what you can do with nginx?  Can't currently order logs correctly with logstash since nginx can output logs with msecs but gunicorn logs are only to the nearest whole second

benoitc:
you can pass your own access log pattern using the [access_log_format](https://docs.gunicorn.org/en/stable/settings.html#access-log-format) setting. request Time can be given in microseconds using the `D` format. Dant can't be changed yet  until you use your own [logging_class](https://docs.gunicorn.org/en/stable/settings.html#logger-class). 

I guess we can make this https://github.com/benoitc/gunicorn/blob/master/gunicorn/glogging.py#L353 customisable though. Would you mind open ing a PR for it?

benoitc:
no activity since awhile. closing feel free to create a new ticket if needed.

PR #2379, opened 2020-07-14
Custom date formatting for logging

johnlawsharrison:
This is my naive first approach to #2328 and I'd like some feedback to make sure I'm heading in the right direction.

- Am I properly addressing the scope of the issue in #2328? Should the date format be configurable for all handlers (using`cfg.logconfig_dict.formatters` ), or would we rather keep the hardcoded default (Apache Common) for everything except access logs?
- The author of the original issue desires microsecond precision for formatted dates, however `now()` currently makes use of `time.strftime()` which [doesn't support microsecond resolution] (https://docs.python.org/3/library/time.html#time.strftime). To support this we would have to switch to using `datetime.now().strftime()`
- I'll introduce some more comprehensive testing once I have a clearer sense of where to go from here

Thanks!

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -175,7 +175,7 @@ class Logger(object):
     loglevel = logging.INFO
 
     error_fmt = r"%(asctime)s [%(process)d] [%(levelname)s] %(message)s"
-    datefmt = r"[%Y-%m-%d %H:%M:%S %z]"
+    default_datefmt = r"[%Y-%m-%d %H:%M:%S %z]"
 
     access_fmt = "%(message)s"
     syslog_fmt = "[%(process)d] %(message)s"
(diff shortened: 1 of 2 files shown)


gunicorn/glogging.py
```diff
@@ -352,7 +353,7 @@ def access(self, resp, req, environ, request_time):
 
     def now(self):
         """ return date in Apache Common Log Format """
```
johnlawsharrison, line 355: TODO: update this docstring to reflect the configurable nature of the format

gunicorn/glogging.py
```diff
@@ -198,6 +198,7 @@ def setup(self, cfg):
         self.loglevel = self.LOG_LEVELS.get(cfg.loglevel.lower(), logging.INFO)
         self.error_log.setLevel(self.loglevel)
         self.access_log.setLevel(logging.INFO)
+        self.datefmt = cfg.logconfig_dict.get('datefmt', self.default_datefmt)
```
johnlawsharrison, line 201: Setting this here seemed reasonable considering that `self.datefmt` is used as early as [line 212](https://github.com/benoitc/gunicorn/pull/2379/files#diff-3f0ccdab4e1ca658f432ac7feb652ccaR212) to set up the error handler. However, is it more appropriate to have this override occur later on [line 232](https://github.com/benoitc/gunicorn/pull/2379/files#diff-3f0ccdab4e1ca658f432ac7feb652ccaR232)?

johnlawsharrison: Ah, I'm re-reading [the python logging docs](https://docs.python.org/3/library/logging.config.html#logging-config-dictschema) and see that the datefmt for each handler will need to be set from dicts in `cfg.logconfig_dict.formatters`

PR #2390, opened 2020-07-29
Support IP networks in allow lists

### LINKED ISSUE #1485: Make it possible to allow a whole subnet in forwarded_allow_ips
I don't know the my reverse proxy's exact IP, but I know its subnet. It'd be great if I could set `--forwarded-allow-ips=10.2.41.0/24`.

For now a workaround is to use a config file like this:

```python
from ipaddress import ip_network
forwarded_allow_ips = '127.0.0.1,' + ','.join(str(ip) for ip in ip_network(u'10.2.40.0/22').hosts())
```

dbnicholson:
Use python's ipaddress module to support IP networks in
`forwarded_allow_ips` and `proxy_allow_ips`. If an address is provided,
it will be converted to an IPv4 /32 or IPv6 /128 network as appropriate.

Fixes #1485.


benoitc:
Applied this feature in commit 2ff3eaf0. forwarded_allow_ips and proxy_allow_ips now support CIDR networks (e.g., 192.168.0.0/16). Thank you for the contribution!

PR #2392, opened 2020-07-30
Log access even when connection is closed

### LINKED ISSUE #1060: Access Log line missing when client times out
I noticed that when running gunicorn with access log enabled, if the client times out and disconnects gunicorn doesn't log anything to the access or error logs. This seems like a bug, the request code typically still runs normally (e.g. there may be new data in the database from this request) so it seems like there should be a record of it in the access log. Other web servers (nginx is the one I tested it on just now) still log to the access log when the client times out, nginx uses the status code 499 for this case.

A simple way to demonstrate -  in a request route in a wsgi app like flask, put in `time.sleep(5)`.  Then hit the route, say it's running locally on port 8000, with `curl --max-time 2 localhost:8000/`. The access log won't show any record of the request.

ikonst:
When a client disconnects, `resp.write(item)` would result in a `BrokenPipeError` which prevents the access from being logged.

Fixes #1060.

### DIFF
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -107,14 +107,16 @@ class AsyncWorker(base.Worker):
             if self.is_already_handled(respiter):
                 return False
             try:
-                if isinstance(respiter, environ['wsgi.file_wrapper']):
-                    resp.write_file(respiter)
-                else:
-                    for item in respiter:
-                        resp.write(item)
-                resp.close()
-                request_time = datetime.now() - request_start
-                self.log.access(resp, req, environ, request_time)
+                try:
+                    if isinstance(respiter, environ['wsgi.file_wrapper']):
+                        resp.write_file(respiter)
+                    else:
+                        for item in respiter:
+                            resp.write(item)
+                    resp.close()
+                finally:
+                    request_time = datetime.now() - request_start
+                    self.log.access(resp, req, environ, request_time)
             finally:
                 if hasattr(respiter, "close"):
                     respiter.close()
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -319,15 +319,17 @@ class ThreadWorker(base.Worker):
 
             respiter = self.wsgi(environ, resp.start_response)
             try:
-                if isinstance(respiter, environ['wsgi.file_wrapper']):
-                    resp.write_file(respiter)
-                else:
-                    for item in respiter:
-                        resp.write(item)
+                try:
+                    if isinstance(respiter, environ['wsgi.file_wrapper']):
+                        resp.write_file(respiter)
+                    else:
+                        for item in respiter:
+                            resp.write(item)
 
-                resp.close()
-                request_time = datetime.now() - request_start
-                self.log.access(resp, req, environ, request_time)
+                    resp.close()
+                finally:
+                    request_time = datetime.now() - request_start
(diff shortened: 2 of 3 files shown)


ikonst:
ptal @benoitc

ikonst:
@benoitc can take a look?

ikonst:
@benoitc can you review this?

ikonst:
@benoitc I'm seeing that gunicorn development renewed. Can you take a look at this?

benoitc:
@ikonitas yes. I should have fixed the CI today. So we can repass on it and merge :) thanks for the head-up :)

ikonst:
Do I need to do anything to kick off CI?

ikonst:
@benoitc gentle ping

ikonst:
@benoitc gentle ping

ikonst:
👋 @benoitc

ikonst:
@benoitc care to take a look?

gunicorn/workers/base_async.py
```diff
@@ -107,14 +107,16 @@ def handle_request(self, listener_name, req, sock, addr):
...
-                self.log.access(resp, req, environ, request_time)
+                try:
+                    if isinstance(respiter, environ['wsgi.file_wrapper']):
+                        resp.write_file(respiter)
+                    else:
+                        for item in respiter:
+                            resp.write(item)
+                    resp.close()
+                finally:
+                    request_time = datetime.now() - request_start
```
benoitc, line 118: why not put this in the `finally` below ? I don't see a need for wrapping it more than it is there

gunicorn/workers/gthread.py
```diff
@@ -319,15 +319,17 @@ def handle_request(self, req, conn):
...
-                self.log.access(resp, req, environ, request_time)
+                try:
+                    if isinstance(respiter, environ['wsgi.file_wrapper']):
+                        resp.write_file(respiter)
+                    else:
+                        for item in respiter:
+                            resp.write(item)
+
+                    resp.close()
+                finally:
```
benoitc, line 330: same question :)

gunicorn/workers/sync.py
```diff
@@ -176,14 +176,16 @@ def handle_request(self, listener, req, client, addr):
...
-                request_time = datetime.now() - request_start
-                self.log.access(resp, req, environ, request_time)
+                try:
+                    if isinstance(respiter, environ['wsgi.file_wrapper']):
+                        resp.write_file(respiter)
+                    else:
+                        for item in respiter:
+                            resp.write(item)
+                    resp.close()
+                finally:
```
benoitc, line 186: same question :)

benoitc (commented):
I like the idea but IMO this extra wrapping need to be removed. If you want to make the change let me know. If this can be done this week that would be awesome :)

benoitc (commented):
I like the idea but IMO this extra wrapping need to be removed. If you want to make the change let me know. If this can be done this week that would be awesome :)

gunicorn/workers/base_async.py
ikonst: I'm not so concerned about `request_time = datetime.now() - request_start` but I wasn't sure if `Logger.access` may fail, and I didn't want to introduce a new failure case where due to Logger failure, we then fail to call `respiter.close()`.

Looking at `Logger.access`, it suppresses `Exception` so it should be _safe enough_ to merge.

PR #2405, opened 2020-08-20
Warn in the case of bad systemd configuration

kenoss:
The timing of [calling sd_notify](https://github.com/benoitc/gunicorn/blob/20.0.4/gunicorn/arbiter.py#L161) is after [daemonizing](https://github.com/benoitc/gunicorn/blob/20.0.4/gunicorn/app/base.py#L217).  Hense, launching will fail eternally if user specify `Type = notify` and `daemon = True`.

User can face this pattern if user already have existing `gunicorn.py` setting file and try to use systemd.  I added warning to make user easilly figure out.

### DIFF
--- a/gunicorn/app/base.py
+++ b/gunicorn/app/base.py
@@ -219,7 +219,8 @@ class Application(BaseApplication):
 
         if self.cfg.daemon:
             if os.environ.get('NOTIFY_SOCKET'):
-                msg = "Warning: you shouldn't specify `daemon = True` when launching by systemd with `Type = notify`"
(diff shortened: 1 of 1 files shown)


kenoss:
Ah, too long line... Could you let me know how we should solve this in this project or general python?  I'm not so fluent in python.

kenoss:
I guess I fixed.  Let me know if other method is better.

kenoss:
@benoitc Could you have a look?

benoitc:
thanks!

Issue #1472, opened 2017-02-27
Gunicorn conf file, environment variables syntax


leonardbinet:
Hello,
In the documentation, it is written that there is no special syntax for config, but I don't understand how to use the env parameter to pass multiple values. 

Should I use a dictionary to pass multiple values?

For instance I would like to pass environment variables in gunicorn processes, my `gunicorn_config.py` file is:
```
bind = "127.0.0.1:8887"
log_file = "path/to/logs/gunicorn_logs.txt"
workers = 3

# Environment variables
env = {
    "DJANGO_SECRET_KEY": "something",
    "OTHER_SECRET": "something_else"
    }
```

But it doesn't work. Is my syntax right?
Thanks in advance

leonardbinet:
Ok, in found it:
```
# Environment variables
raw_env = [
    "DJANGO_SECRET_KEY=something",
    "OTHER_SECRET=something_else"
    ]
```

benoitc:
ok, closing the issue since you found it :)

alexprengere:
Hello, I was stuck on this exact issue today, because I used the [example](https://github.com/benoitc/gunicorn/blob/master/examples/example_config.py) which is not up to date with the [latest docs](https://gunicorn-docs.readthedocs.io/en/18.0/configure.html). Thanks to this issue, I could find `raw_env` though 😉

berkerpeksag:
@alexprengere I've just added an example to `examples/example_config.py` in https://github.com/benoitc/gunicorn/commit/01a1c7ca9a76d14a707b2946192ca7278247c2ce.

PR #2413, opened 2020-09-02
Show usage of raw_env in docs

marcanuy:
It was documented the usage of the cli parameter `env` but in the config file it should be `raw_env`.
Related issue #1472


benoitc:
I think we should keep this behaviour hidden. Closing th eissue but thanks anyway for the patch.

Issue #2216, opened 2019-12-11
Gunicorn worker shared class


abhishekduttagupta:
Hello,

I would like to know how we can share class/context across different gunicorn?
That means suppose I have a class which is getting created and updated by worker 1, and 2nd worker is trying to fetch it first and then update it?

What is the way possible to achieve this in gunicorn?

jamadden:
Beyond preloading the application, which shares a certain amount of memory and class data that is *copied* by each worker, gunicorn doesn't have any mechanism for sharing data between workers (and preloading does nothing about updating data). That is, gunicorn doesn't provide any IPC facilities. You'll need to pick and implement a method appropriate to your application.

 Some people use a shared database like Redis or MySQL/PostgreSQL, often with subscribers/listeners. Others use libraries like 0MQ to establish sockets between workers and communicate that way.

jeffin143:
why don't we inland support that, I mean ship an api, Which has some way to communicate across workers

Or at least let's mention that in docs, Like Data sharing across workers aren't there, I wasted a whole week just debugging why My code didn't break in dev mode and broke using unicorn

Skimmed through the docs but couldn't find and apt ans, Thanks to this issue 

Re wrote the code using Redis Django

Issue #1938, opened 2018-12-21
Shared object on master process?


lhotakj:
Hi all,
let me follow up #1692. Sure, you can use `redis` or socket based communicator for sharing objects between process, but I am wondering if this could be somehow done using server hooks methods? Could `Multiprocessing.Process` be used for it? Is there any build-in way how to achieve that, please? 
I believe that it would be much easier than relying on third-party / custom solution. 
Thank you, 
Jaroslav

lhotakj:
HI Benoit, 
thanks for your reply. I am aware of the danger of having an object shared in memory. Let me explain my intention - was thinking how to store let's say user session data, which becomes a problem when you spawn more than one worker. 

Actually, I am solving it now by starting a simple socket server [Kasi]*(https://github.com/lhotakj/Kasi) before `gunicorn` -- I could have use eg. `redis` but I preferred something small which doesn't require installation. However I still think it could be achieved internally by `gunicorn` like another process, exactly as you proposed. 
\* it's not completely ready, missing readme, tests, install, but the server/client work well.

Thanks :)

EeyoreLee:
hi @benoitc , I have the same problem. In contrast to this, the object what I wanna shared is a DeepLearning model. it will cost more gpu memory. when I setup 4 workers, there are 4 times gpu memory cost. But the app not only the one route. So I dont need so many model duplicates. If I can put model in the parent process, it will much better. It means I can setup 4 workers and 2 model duplicates. I can go load balancing myself. And I also know this is dangerous cause I must add a lock for it. But it is really necessary for me cause I also need to update my model via an http api. It will cause only the worker who gets the request to actually update. Due to the special nature of this object, I have no way to put it in something like Redis. Iam so sad. :(

EeyoreLee:
> hi @benoitc , I have the same problem. In contrast to this, the object what I wanna shared is a DeepLearning model. it will cost more gpu memory. when I setup 4 workers, there are 4 times gpu memory cost. But the app not only the one route. So I dont need so many model duplicates. If I can put model in the parent process, it will much better. It means I can setup 4 workers and 2 model duplicates. I can go load balancing myself. And I also know this is dangerous cause I must add a lock for it. But it is really necessary for me cause I also need to update my model via an http api. It will cause only the worker who gets the request to actually update. Due to the special nature of this object, I have no way to put it in something like Redis. Iam so sad. :(

[...]

PR #2414, opened 2020-09-02
Add extensible worker arguments

four43:
So this is mainly to solicit feedback. It's certainly not complete, but I figured a code example and a place to grow this idea was helpful.

**Problem:** I need to share some state across worker processes. In this case write to a file across multiple processes that don't clobber each other. That file then needs to be shipped off after X time/size and the mess of race conditions that surround that using multiple writers with no clear leader.

**Solution:**
Fork off a helper from the master process that can listen to a shared multiprocess queue and be the one source of truth for that. The issue is sharing this queue with the worker processes.

For a quick example (with this PR applied) I could do something like:

server.py
```
def app(environ, start_response, app_queue, **kwargs):

    # Add stuff to our shared queue
    app_queue.put(f"Req from: {environ['HTTP_USER_AGENT']}")

    data = b'Send data!\n'
    status = '200 OK'
    response_headers = [
        ('Content-type', 'text/plain'),
    ]
    return iter([data])
```

config.py
```
import multiprocessing

app_queue = multiprocessing.JoinableQueue()
helper_proc: multiprocessing.Process = None

def helper(server, app_queue):
    while True:
        try:
            msg = app_queue.get(block=True, timeout=3)
            server.log.info(f"WORKER MSG: ${msg}")
            app_queue.task_done()
        except queue.Empty:
            server.log.info("Worker waiting for messages...")


def pre_fork(server, worker):
    server.log.info("Pre-fork spawned (pid: %s)", worker.pid)
    # New part, modify these args that get passed to the handler.
    worker.cfg.handler_kwargs['app_queue'] = app_queue


def when_ready(server):
    global helper_proc
    server.log.info("Server is ready. Spawning workers and helper")
    helper_proc = multiprocessing.Process(target=helper, args=(server, app_queue,))
    helper_proc.start()


def on_exit(server):
    server.log.info("on_exit event")
    app_queue.join()
    helper_proc.kill()  # Everything is done processing if the messages were all ack-ed
```

I'm guessing other worker classes I'm less familiar with don't handle this as well? This seems like a great extension that allows a lot of flexibility though.

I think this could help #2216 #1938

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -50,6 +50,7 @@ class Config(object):
         self.usage = usage
         self.prog = prog or os.path.basename(sys.argv[0])
         self.env_orig = os.environ.copy()
+        self.worker_kwargs = {}
 
     def __str__(self):
         lines = []
--- a/gunicorn/workers/sync.py
+++ b/gunicorn/workers/sync.py
@@ -174,7 +174,7 @@ class SyncWorker(base.Worker):
             if self.nr >= self.max_requests:
                 self.log.info("Autorestarting worker after current request.")
                 self.alive = False
-            respiter = self.wsgi(environ, resp.start_response)
+            respiter = self.wsgi(environ, resp.start_response, **self.cfg.handler_kwargs)
             try:
                 if isinstance(respiter, environ['wsgi.file_wrapper']):
                     resp.write_file(respiter)


benoitc:
the `--env` can be used for that. This change would break the current expectation. I am closing it as I think there is a need for a better solution to share data across process and provide a real alternative to multiprocessing that takes advantage of the system.

PR #2418, opened 2020-09-10
fix: repair django wsgi running error

duanhongyi:
When running the WSGI of Django, you need to wrap it with WSGIContainer. Otherwise, the following error will be caused:

```
ERROR:tornado.application:Uncaught exception
  File "/usr/local/lib/python3.7/site-packages/tornado/http1connection.py", line 273, in _read_message
    delegate.finish()
  File "/usr/local/lib/python3.7/site-packages/tornado/httpserver.py", line 280, in finish
    self.request_callback(self.request)
TypeError: __call__() missing 1 required positional argument: 'start_response'
```

I think that in tornado6 and above, as long as it is not a WSGIContainer instance, we should wrap it into WSGIContainer

### DIFF
--- a/gunicorn/workers/gtornado.py
+++ b/gunicorn/workers/gtornado.py
@@ -110,6 +110,8 @@ class TornadoWorker(Worker):
             if not isinstance(app, tornado.web.Application) or \
             isinstance(app, tornado.wsgi.WSGIApplication):
                 app = WSGIContainer(app)
+        elif not isinstance(app, WSGIContainer):
+            app = WSGIContainer(app)
 
         # Monkey-patching HTTPConnection.finish to count the
         # number of requests being handled by Tornado. This


waghanza:
Hi @duanhongyi,

I was not aware that you can run a django app with tornado worker, I use to take https://github.com/mopemope/meinheld.

Which version of `tornado` are you using ?

Seems that removing the version condition is working for both versions **5** and **6**) ℹ️ this `PR` break using tornado6 with django, but not 5.1

Regards,

duanhongyi:
@waghanza 

When I mentioned this RP, I was worried about breaking the compatibility, so I reserved the judgment. I think the version judgment can be removed from the source code. 


Because WSGIContainer existed in tornado nine years ago ^_^

waghanza:
sure. I've tested, removing the conditions works, but I have to test also with `django` ❤️

waghanza:
@duanhongyi I've made https://github.com/benoitc/gunicorn/pull/2550 to ask for this condition removal (this condition breaks our tornado @benoitc)

See a working example that use the above PR
https://the-benchmarker.semaphoreci.com/workflows/1630d7dc-f14d-4d56-ac0e-ddf21be8ef08?pipeline_id=bb457178-c247-4d47-967a-bc49ee41afa1

alviezhang:
This PR breaks tornado applications execution which versions >= 6.0. Here is an related issue #2630, and you can reproduce this issue by the official tornado application example
```python
import tornado.ioloop
import tornado.web

class MainHandler(tornado.web.RequestHandler):
    def get(self):
        self.write("Hello, world")

def make_app():
    return tornado.web.Application([
        (r"/", MainHandler),
    ])


app = make_app()


if __name__ == "__main__":
    app.listen(8888)
    tornado.ioloop.IOLoop.current().start()

```

and command
```shell
gunicorn app:app -k tornado
```

And I also tested this PR reversion of code too, which also failed with tornado >= 6.0. 
This means since this PR merged and released, the tornado worker breaks till now.

duanhongyi:
> This PR breaks tornado applications execution which versions >= 6.0. Here is an related issue #2630, and you can reproduce this issue by the official tornado application example
> 
> ```python
> import tornado.ioloop
> import tornado.web
> 
> class MainHandler(tornado.web.RequestHandler):
>     def get(self):
>         self.write("Hello, world")
> 
> def make_app():
>     return tornado.web.Application([
>         (r"/", MainHandler),
>     ])
> 
> 
> app = make_app()
> 
> 
> if __name__ == "__main__":
>     app.listen(8888)
>     tornado.ioloop.IOLoop.current().start()
> ```
> 
> and command
> 
> ```shell
> gunicorn app:app -k tornado
> ```
> 
> And I also tested this PR original reversion of code `83319f7` also failed with tornado >= 6.0,
> which means since this PR merged and released, the tornado worker breaks till now.

#2633

PR #2427, opened 2020-09-23
Remove reference to six.py

moshekaplan:
six.py was removed in https://github.com/benoitc/gunicorn/pull/2083 , there's no need to exclude it from tests.


berkerpeksag (approved):
LGTM!

PR #2431, opened 2020-10-05
Added a line to the documentation to further explain the use of logconfig_dict.

Luttik:
The usage of logconifg_dict can be confusing, not only since there is not a clear example on the site now, but also since the documentation of [python's dictConfig](https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig) is minimal. And since you need to know about the specific loggers that gunicorn uses to configure them properly.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -304,6 +304,8 @@ older file configuration format.
 
 Format: https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig
 
+For more context you can look at the default configuration dictionary for logging, which can be found at ``gunicorn.glogging.CONFIG_DEFAULTS```.
+
 .. versionadded:: 19.8
 
 .. _syslog-addr:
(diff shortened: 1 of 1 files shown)


docs/source/settings.rst
```diff
@@ -304,6 +304,8 @@ older file configuration format.
 
 Format: https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig
 
+For more context you can look at the default configuration dictionary for logging, which can be found at ``gunicorn.glogging.CONFIG_DEFAULTS``.
```
tilgovi, line 307: This is great. I wonder if there's a way we could actually embed `CONFIG_DEFAULTS` here as the default value in the documentation, or if we could link to it.

Luttik: Good question. We can of course link to [logging](https://github.com/benoitc/gunicorn/blob/master/gunicorn/glogging.py) but I don't know if there is any way to auto-link to a variable in the file.

Luttik: It is possible to link directly to a linenumber though. But a commit could then break the location. And locking to a commit hash is also not preferable I believe...

tilgovi (approved):

PR #2438, opened 2020-10-13
Improve settings documentation

adamchainz:
I was reading the documentation to configure another application successfully deployed with Gunicorn, and thought the documentation here could be improved a bit for clarity.

* Use code highlighting for the names, to indicate they are the names to use in the Python configuration file.
* Spell out the CLI flags and defaults with "Command line" and "Default" prefixes so it's clear what these are
* Consistently use Python syntax for the defaults, so the types may be more easily understood.
* Split multiple CLI flags with "or" rather than a comma, for clarity and easier copy-paste.
* Improve some descriptions, such as for ``raw_env`` and a header for "Server Hooks".

### DIFF
--- a/docs/gunicorn_ext.py
+++ b/docs/gunicorn_ext.py
@@ -50,27 +50,29 @@ def format_settings(app):
 def fmt_setting(s):
     if callable(s.default):
         val = inspect.getsource(s.default)
-        val = "\n".join("    %s" % l for l in val.splitlines())
-        val = " ::\n\n" + val
+        val = "\n".join("    %s" % line for line in val.splitlines())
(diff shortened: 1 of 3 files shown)


berkerpeksag (commented):
Thank you for the PR. This file is generated by https://github.com/benoitc/gunicorn/blob/master/docs/gunicorn_ext.py. Could you please update that file instead?

adamchainz:
Done! I dropped by section header for "Server Hooks" since the autogenerator doesn't currently support them.

adamchainz:
Is there anything I can do to help get this reviewed?

berkerpeksag (approved):
LGTM, thank you!

PR #2439, opened 2020-10-14
Update nginx.conf

ipal0:
added location for static files


tilgovi:
The current example is just an example and it uses `try_files` so that the static root can be shared with the application path. I'm going to close this, but if I've misunderstood please feel free to comment and I can reconsider! Thank you!

PR #2446, opened 2020-10-30
Fix signature of the "fall through" InotifyReloader

jkugler:
The signature of `__init__` on the "fall through" InotifyReloader
was missing the extra_files paramater, so specifying the inotify
reload engine on the command line when one did not have inotify
installed would, instead of a nice message about needed inotify
installed would result in the following traceback:

```
[2020-10-30 00:55:43 +0000] [7] [ERROR] Exception in worker process
  File "/usr/lib/python3/dist-packages/gunicorn/arbiter.py", line 583, in spawn_worker
    worker.init_process()
  File "/usr/lib/python3/dist-packages/gunicorn/workers/base.py", line 132, in init_process
    self.reloader = reloader_cls(extra_files=self.cfg.reload_extra_files,
TypeError: __init__() got an unexpected keyword argument 'extra_files'
```

I didn't see an easy way to writing a test for this, but would be
happy to take pointers.

### DIFF
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -118,7 +118,7 @@ if has_inotify:
 else:
 
     class InotifyReloader(object):
-        def __init__(self, callback=None):
+        def __init__(self, extra_files=None, callback=None):
             raise ImportError('You must have the inotify module installed to '
                               'use the inotify reloader')
 


frafra:
Is there any reason why this PR is stuck? Any feedback from @benoitc?

benoitc:
thanks for the PR

Issue #2454, opened 2020-11-10
murder_worker does not let enough time for sigabort


tardyp:
Hello,

We are debugging some cases where our worker are getting overloaded, and we hit the murder timeout.

In this code, we don't see an additional delay between SIGABRT and SIGKILL
https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L500

In practice, we never see the SIGABRT hook on our workers, so we cannot dump the thread stacks to see where we are at that point

De we miss another timeout mechanism?
Shall we patch arbiter to add another graceful timeout?

tilgovi:
The delay is only supposed to be one second here as a best effort to let the worker respond to the abort by logging or otherwise dumping some state.

I'm working on fixing the graceful shutdown for all the async worker types so that they can notify the arbiter not to kill them during graceful shutdown. Abort is not a graceful shutdown, though.

tardyp:
In our case, the goal was to let the worker dump stack traces for all its threads. So that we can know what is causing the large load.

I kinda insist on the specificity of this bug, and the need to be taken in account. Can we keep the bug open until your fix is in?

tardyp:
Also, you mention that the delay is supposed to be one second, while in my tests, there is *no delay*. 

This is because the following check will be true for the second check (after we first aborted the worker), as it was for the first check. last_update() won't be updated as we are in a case of worker lock.

```
    if time.time() - worker.tmp.last_update() <= self.timeout:
```

tilgovi:
There should be a delay because the arbiter only runs its main loop once per second unless it receives a https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L351

PR #2455, opened 2020-11-10
add a delay between SIGABRT and SIGKILL for worker timeouts

tardyp:
candidate fix for #2454

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -491,8 +491,11 @@ class Arbiter(object):
             return
         workers = list(self.WORKERS.items())
         for (pid, worker) in workers:
+            timeout = self.timeout
+            if worker.aborted:
+                timeout += self.cfg.graceful_timeout
             try:
-                if time.time() - worker.tmp.last_update() <= self.timeout:
+                if time.time() - worker.tmp.last_update() <= timeout:
                     continue
             except (OSError, ValueError):
                 continue


tilgovi:
Thank you for the patch, but I don't think this is the way I'd like to handle this issue. This function is meant for killing workers that are stuck. A proper implementation of graceful shutdown in the async workers should continue to notify the arbiter of their liveness during shutdown. Not all workers are doing this correctly right now. There is some discussion in https://github.com/benoitc/gunicorn/issues/2317 where I'm working on fixing this.

I'll close your PR, but if you think I've misunderstood I'll be happy to discuss further.

tardyp:
Hi @tilgovi, I indeed think there is a misunderstanding.

This fix is in reality for the worker stuck use case.
When this is the case, there is no delay between the SIKABRT and SIGKILL, because of the logic I am fixing.

In #2454, you say there is 1 second delay, please point me to the actual code that is introducing this delay, because I can't find it, and in my tests and logs I never see a one second delay between those two signals.

tilgovi:
@tardyp the arbiter process is supposed to sleep for one second if it does not receive any signals: https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L357

We could re-open this PR and modify it if you find that this is not reliable enough, but we should not wait the whole graceful shutdown period since abort is not graceful.

PR #2460, opened 2020-11-21
[Typo] Custom Application Section

SalahEddineBC:
There is a typo in custom configuration's page where commands to run a custom app were missing a parameter.


tilgovi:
Thank you!

PR #2471, opened 2020-12-09
Add link to reference docs in README

Leo-Ryu:
Hi there!

My name is Leo, a CS student at Brown building an automated platform for editable codebase documentation.

I generated documentation for this repository and placed a link in the README. Please feel free to try it out and let me know what you think!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -25,6 +25,10 @@ Documentation
 
 The documentation is hosted at https://docs.gunicorn.org.
 
+Reference Docs
+--------------
+https://docs.contour.so/benoitc/gunicorn/getting-started
+
 Installation
 ------------
 


jamadden:
Thank you for your contribution! I appreciate the effort spent here.

While completely automated documentation has its place, there are some issues with this that I think make it unlikely that this will get merged to be an official part of the project. (It could certainly be useful for personal learning, of course.)

First, it seems to be a simple flat listing of every symbol defined in the project, whether or not it has a docstring. Most of those symbols (classes and functions) are internal implementation details that are very deliberately *not* documented: they are not public, and no consumer outside this code base needs to know about them. A big chunk of the other symbols are from example programs, which are also typically not why someone would be looking for gunicorn API reference documentation. (But I could see that comprehensive symbols dumps could be useful to someone learning the gunicorn code base itself.)

Second, because of the flat nature of the listing, it's extremely hard to navigate and put into context. A big part of API documentation is information about how things fit together (cross references and hierarchy) and that seems to be missing.

Third, how would this ever get updated? (We already have a process that updates documentation; it's not clear to me that we need another. See below.)

Last, this project already has a documentation process based on Sphinx, automatically built by [readthedocs](https://readthedocs.org/projects/gunicorn-docs/builds/) and automatically published to https://docs.gunicorn.org/. With Sphinx's [autodoc extension](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) it's easily possible to create reference documentation in a similar format; the process is only semi-automated because it requires some level of curation from the documentation writers in choosing what modules, classes, functions, or other symbols get documented, but that's a good thing. Were gunicorn to publish API reference documentation, I suspect that would be the preferred way to do it, if only because it's extremely popular in the Python community (e.g., Sphinx is used by Python itself to produce [the standard library documentation](https://docs.python.org/3/library/index.html)).

Issue #2215, opened 2019-12-06
Silent OOM (Booting worker)


ilivans:
Here is a reproducible example of the problem mentioned in https://github.com/benoitc/gunicorn/issues/1663

`app_oom.py`:
```
#!flask/bin/python
import logging
import traceback

from flask import Flask, jsonify

app = Flask(__name__)
app.logger.addHandler(logging.StreamHandler())
app.logger.setLevel('DEBUG')


@app.route('/hello_oom', methods=['GET'])
def hello_oom():
    try:
        huge_list = []
        while True:
            huge_list.extend([1] * 100000)
        logging.info('Success')
        return jsonify({'output': 'Success'}), 201
    except:
        logging.error(traceback.format_exc())
        return jsonify({'error': traceback.format_exc(), 'error_code': 10}), 500  # Unknown error


if __name__ == '__main__':
    app.run(host='0.0.0.0', debug=False)
```

`Dockerfile.oom`:
```
FROM ubuntu:16.04
RUN apt update -y \
    && apt install -y gcc python3-pip python3-dev \
    && pip3 install flask==1.1.1 gunicorn==19.9.0
COPY app_oom.py app_oom.py
EXPOSE 8888
CMD ["gunicorn", "-b", "0.0.0.0:8888", "--timeout", "20000", "--workers", "1", "--log-level", "debug", "app_oom:app"]
```

`docker build -t app_oom . -f Dockerfile.oom`
`docker run -ti -p 8888:8888 -m 500m --name app_oom --rm app_oom`

When I make a request, in the logs there is no error message, but a message about new worker booting
```
[2019-12-06 14:48:56 +0000] [12] [INFO] Booting worker with pid: 12
[2019-12-06 14:48:56 +0000] [1] [DEBUG] 1 workers
[2019-12-06 14:48:58 +0000] [12] [DEBUG] GET /hello_oom
[2019-12-06 14:49:32 +0000] [14] [INFO] Booting worker with pid: 14
```

And the client gets no response (using `curl http://0.0.0.0:8888/hello_oom`):
```
curl: (52) Empty reply from server
```
@benoitc

monobaila:
I've just been hit by exactly the same issue with a django app deployed on kubernetes. In my case I set a low memory limit which caused OOM killer to be invoked once this was exceeded.

Do you see evidence of this if you run `dmesg` on the test machine?

[...]

PR #2475, opened 2020-12-16
Log a warning when a worker was terminated due to a signal

aberres:
This happens for example when being OOM killed.

See #2215

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -526,6 +526,12 @@ class Arbiter(object):
                     if exitcode == self.APP_LOAD_ERROR:
                         reason = "App failed to load."
                         raise HaltServer(reason, self.APP_LOAD_ERROR)
+                    if os.WIFSIGNALED(status):
+                        self.log.warning(
+                            "Worker with pid %s was terminated due to signal %s",
+                            wpid,
+                            os.WTERMSIG(status)
+                        )
 
                     worker = self.WORKERS.pop(wpid, None)
                     if not worker:


aberres:
@benoitc @tilgovi Similar to the just merged #2277 this is a change which could make people running Kubernetes workloads a tad happier. Otherwise it is pretty hard to monitor OOM situations.

benoitc:
which signal are we talking?

aberres:
> which signal to who are we talking?

Kill.

When you have set memory limits set in Kubernetes it can easily happen that a worker is shot down without any kind of logging. Restarting work nicely, it is just that having a log eases the post mortem.

benoitc:
i would suggest to configure kubernetes to send a sigterm instead of a
sigkill signal instead. That way the worker will be terminated gracefully/

Also if you know your application will leak or use a lot of ram over the
time you should limit the number of connections a worker can accept until
it is killing itself.

aberres:
Yeah, that's an ongoing discussion of it should not send a SIGKILL immediately. But not likely to be changed as I understood the issues I was looking at.

More often than not when this situation arises it is likely a  configuration issue and the requested memory should be increased.

Thanks a lot for the merge!

tilgovi:
K8s does send TERM for normal shutdown, but OOM killer in Linux always sends KILL, I think.

CodeOnCoffee:
@benoitc Why was this revered in: https://github.com/benoitc/gunicorn/commit/76f8da24cbb992d168e01bda811452bcf3b8f5b3

We have workers being silently OOMKilled in Kubernetes and would like to setup an alert based on the logs

javabrett:
Due to reentrant logging calls when handling a signal during another logging call, as observed in https://github.com/benoitc/gunicorn/issues/2564.

PR #2476, opened 2020-12-16
Remove the --log-config-dict flag

### LINKED ISSUE #1909: log-config-dict cli option is broken, can't work as coded.
Despite https://github.com/benoitc/gunicorn/issues/1880 being closed, the `--log-config-dict` cli option is broken and can't work as coded because the CLI's `type` member isn't set, so the dict_validator tries to validate a string (what's coming from ARGV), unconverted, and that fails.
I tried hacking the code to use `type = json.loads` so that the string would be converted to a dict by ArgParse, but couldn't make that work either. (I think that *can* work, I was just screwing something up I'm guessing).

tilgovi:
There is no support for decoding any dictionary supplied on the command
line. The only way to supply a dictionary logging config is through the
configuration file.

Close #1909.

### DIFF
--- a/docs/source/news.rst
+++ b/docs/source/news.rst
@@ -2,6 +2,12 @@
 Changelog
 =========
 
+Unreleased
+==========
+
+- remove the `--log-dict-config` CLI flag because it never had a working format
+  (the `logconfig_dict` setting in configuration files continues to work)
+
 20.0.4 / 2019/11/26
 ===================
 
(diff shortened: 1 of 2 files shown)


jamadden (approved):
LGTM.

I'm curious why we don't add changelog entries as we go. What's the process for updating that? (If we're concerned about conflicts, I've had good luck with [towncrier](https://pypi.org/project/towncrier/) so far.)

tilgovi:
I'll add a changelog entry, thank you.

I would love it if we had automatic changelog generation. We should finalize the next release and then introduce something like that going forward.

PR #2480, opened 2020-12-18
Update Debian and Ubuntu distribution information

kofrezo:
- Recommend latest (supported) Debian and Ubuntu versions
- Update provided versions of gunicorn in Debian and Ubuntu

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -102,7 +102,7 @@ stable ("buster")
 The version of Gunicorn in the Debian_ "stable" distribution is 19.9.0
 (December 2020). You can install it using::
 
-    $ sudo apt-get install gunicorn
+    $ sudo apt-get install gunicorn3
 
 You can also use the most recent version 20.0.4 (December 2020) by using
 `Debian Backports`_. First, copy the following line to your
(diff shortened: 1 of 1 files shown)


kofrezo:
Hey :wave:,
It is not much of a contribution but I felt like I can help you a little bit with updating these parts while reading it :-)

tilgovi:
Thank you, this is great! There's a build failure for the docs linting, for some reason. Can you tell if that's spurious or real and whether there's anything to do?

tilgovi:
Add, line 122 and line 144, you need to extend the underline. Sadly, reStructuredText is picky like that.

kofrezo:
> Add, line 122 and line 144, you need to extend the underline. Sadly, reStructuredText is picky like that.

I see thanks for pointing me to the right lines. I adjusted them.

kofrezo:
I added another commit where I adjusted the package name to `gunicorn3`. While Debian release newer than Stretch have a virtual package gunicorn3 which points to gunicorn with Python3 support for Stretch (oldstable) there is still a difference (Python2 vs Python3 version).

I think this makes it more clear which version one installs no matter with release one is using.

tilgovi:
Very helpful. Thank you!

PR #2481, opened 2020-12-18
Typo: "killed at startup"

guettli:
I guess the "and" was a typo.


tilgovi:
I think "and" is not a typo here. I think the intention is to communicate that workers may stop and new workers may start.

We could say "It is normal workers top stop and for new workers to start" or something like this, if you'd like to try to make this clearer. Thank you!

tilgovi:
Ohhh, I just realized this should be "killed and start up" (two words). That might be the source of the confusion. Thank you, @guettli, I'll make a change.

tilgovi:
Fixed in 6fbe6a1.

PR #2492, opened 2021-01-07
Clarify the secure scheme behaviour, with examples

TMiguelT:
The way that `forwarded_allow_ips`, and the value of `secure_scheme_headers` interact is somewhat complex, and isn't documented in a way that is clear to me. For this reason, I have clarified this in a note, using a series of examples. I have also clarified that only one entry in the dictionary needs to be matched for secure mode to be enabled.


TMiguelT:
I've mostly based this documentation on this part of the code, but please let me know if it is in any way inaccurate: https://github.com/benoitc/gunicorn/blob/6fd97b95595c9b9034a476741a53d849fecfc729/gunicorn/http/message.py#L110-L118

gunicorn/config.py
```diff
@@ -1236,10 +1236,16 @@ class SecureSchemeHeader(Setting):
...
-        uses to indicate HTTPS requests. These tell Gunicorn to set
+        uses to indicate HTTPS requests. If the source IP is permitted by
+        ``forwarded-allow-ips`` (below), *and* at least one request header matches
+        a key-value pair listed in this dictionary, then Gunicorn will set
         ``wsgi.url_scheme`` to ``https``, so your application can tell that the
         request is secure.
 
+        If the other headers listed in this dictionary are not present in the request, they will be ignored,
+        but if the other headers are present and do not match the provided values, then
+        the request will fail to parse. See the note below for more detailed examples of this behaviour.
```
tilgovi, line 1247: This paragraph is a little bit confusing to me. Perhaps the first sentence is unnecessary and it would be better to focus on the failure case.

"If any of the headers are present and do no match their expected value, then the request will fail."

tilgovi (approved):
This is great. Thank you very much, this is indeed a little bit complicated.

I made one suggestion to try to make the explanation even simpler!

gunicorn/config.py
TMiguelT: It's important to note that any *one* key in  `secure_scheme_headers` must match, not *every* key. That's what I'm getting at in that first sentence.

tilgovi: Maybe "Requests only need to provide one header with its expected value. If requests provide more than one header, all headers must have their expected values or the request will fail."

Does something like this work?

TMiguelT: How about a more formalised:
> In order to enable the HTTPS behaviour:
> * At least one header in the request must be present as a key in the dictionary
> * All keys in the dictionary that are match the name of a request header must also exactly match by value

TMiguelT: I'm not actually certain this is better, but I haven't found any of our suggestions entirely satisfactory. Possibly because English just isn't perfect at expressing this precise condition.

tilgovi:
I re-read the proposed language and it seems clearer this time. Maybe I was sleepy last time.

Thanks so much for the helpful contribution.

PR #2499, opened 2021-01-20
Disable logs propagation

### LINKED ISSUE #2496: Gunicorn duplicates logs
I need to run gunicorn directly from code, because my app has its own CLI. Also, I want to make all log the same format. Before that I used loguru, but I had got the same problem, when I try to pass `logconfig_dict`. I thought the problem was in loguru and try to rewrite code to pure python and gunicorn and I have got the same bug. 

Python 3.8.5
gunicorn 20.0.4
Ubuntu 20.04.1 LTS (GNU/Linux 5.4.0-59-generic x86_64)

image

```
import gunicorn.app.base

GUNICORN_LOGGING_FORMAT = dict(
    formatters={
        "generic": {
            "format": "%(levelname)s %(funcName)s %(lineno)d %(message)s"
        }
    }
)


    data = b'Hello, World!\n'
    status = '200 OK'
    response_headers = [
        ('Content-type', 'text/plain'),
    ]
    return iter([data])


class GunicornApplication(gunicorn.app.base.BaseApplication):

    def __init__(self, app, options=None):
        self.options = options or {}
        self.application = app
        super().__init__()

    def load_config(self):
        config = {key: value for key, value in self.options.items()
                  if key in self.cfg.settings and value is not None}
        for key, value in config.items():
            self.cfg.set(key.lower(), value)

    def load(self):
        return self.application


options = {
    "loglevel": "info",
    "logconfig_dict": GUNICORN_LOGGING_FORMAT

}
GunicornApplication(app, options=options).run()
```

dbalabka:
...when logconfig_dict or logconfig provided. Fix #2496

### DIFF
--- a/gunicorn/glogging.py
+++ b/gunicorn/glogging.py
@@ -54,14 +54,14 @@ CONFIG_DEFAULTS = dict(
             "gunicorn.error": {
                 "level": "INFO",
                 "handlers": ["error_console"],
-                "propagate": True,
+                "propagate": False,
                 "qualname": "gunicorn.error"
             },
 
             "gunicorn.access": {
                 "level": "INFO",
                 "handlers": ["console"],
-                "propagate": True,
+                "propagate": False,
                 "qualname": "gunicorn.access"
             }
         },


tilgovi:
I responded over https://github.com/benoitc/gunicorn/issues/2496#issuecomment-767000598 as well. I think an argument can be made for either choice.

I'm tempted to keep this as it is, because it would be potentially disruptive to stop propagating logs that people expect to see right now.

dbalabka:
@tilgovi if Gunicorn utilities sem.versioning, we can bump major version as I've mentioned in comment:
https://github.com/benoitc/gunicorn/issues/2496#issuecomment-763646589

Still, logging isn't affecting Gunicorn functionality so, IMO it is not a BC break and config can be adjusted w/o concerns. 

Hope, that most of the users do not update software blindly and lock versions in their requirements.txt/pyproject.toml/pipenv.toml

tilgovi:
@benoitc what do you think? Minor breaking change to log defaults. Maybe now that you made a release this is a good time. We just need to update the changelog.

benoitc:
parent ticket has been closed

PR #2526, opened 2021-02-24
Fix a naming error for gthread worker

smilerlee:
The name for threaded worker is `gthread`, not `threads`.

This fix only affects a startup log message says `Using worker: xxx` when `sync` is used and `threads` greater than 1.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -106,7 +106,7 @@ class Config(object):
         # are we using a threaded worker?
         is_sync = uri.endswith('SyncWorker') or uri == 'sync'
         if is_sync and self.threads > 1:
-            return "threads"
+            return "gthread"
         return uri
 
     @property


benoitc:
The worker name is correct afaik. It is associated to the module like `ggevent` for the gevent worker : https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gthread.py

I think it's ok to use the correct name of the worker there. Thanks for the patch anyway.

smilerlee:
@benoitc I'm afraid you'll get `gthread` if you configured `worker_class` to `gthread`, not `threads` (or `thread` if you prefer, which is found nowhere) as you would expected.

I can see it would be better to strip off the `g`-prefix from the module name (`gevent` for `ggevent` module, `tornado` for `gtornado` module, etc), but unfortunately, `gthread` is what we have got for long.

benoitc:
I misread the change. I need more café... Sorry for that.

smilerlee:
Glad I’m helped :)

Issue #2196, opened 2019-11-23
application start time


benoitc:
Some applications may takes time to be loaded or initialised which may be an issue sometimes because the worker may be killed in between. One way to fix that would be notifying the arbiter when the application is loaded (which can be done via the shared pipe). 

To prevent the cases whre the application is too long to start, an application start time would be given. If the arbiter is not notified during that time it would kill the worker and report the timeout as usual. 

This would fix #1658, #1757 . 

Thoughts?

tilgovi:
I think we should do it. We still need to decide if the existing `--timeout` should act as a request timeout for non-sync workers. I think we first say only that it is _not_ an application boot timeout and make that a new option.

tilgovi:
We could even start with no timeout for application start. We could keep the current heartbeat behavior for `--timeout` but only start checking it once the worker "boots". That's a breaking change, but for most use I would think health checks and such will make sure that we do not need to have a timeout for booting the application.

benhiller:
Hello! We ran into an issue related to this with our gunicorn deployment. We set a smaller timeout for requests, but our workers sometimes take longer than that just to initialize themselves. Ideally we would improve our application to take less time to initialize, but that seemed like it could be a decent amount of work, and modifying gunicorn to avoid this case seemed relatively simple compared to that.

I made a small change to gunicorn that adds a `boot-timeout` setting. When it is set, the arbiter uses that to decide whether workers have timed out rather than the existing `timeout` setting if the workers haven't booted yet. I am not familiar with the gunicorn codebase, so it is definitely possible I made a glaring mistake when working on this, but if you would like to look at the patch I wrote to add this setting, it is here: https://github.com/benhiller/gunicorn/pull/1 As another caveat, I may also not be aware of the various ways in which gunicorn can be used, so it possible this change might not be compatible with other gunicorn configurations.

I wanted to share that to see if you would be interested in merging it as a pull request to gunicorn? The solution suggested above of not applying the timeout for workers until the boot would also definitely help with our problem, but I went with this approach instead since it seemed less disruptive to the existing behavior, and only changes the behavior for people who opt-in to this by setting a new gunicorn flag. 

Anyway, let me know what you think!

gueugaie:
Hi,

Such a feature may help me, I might consider writting it myself, but I admin I have limited knowledge of the code base, and not a great deal of time to see it through.

Nonetheless, the way I understand things, @benoitc seems to be OK with a basic design such that : 

[...]

PR #2545, opened 2021-03-21
Add boot_timeout setting

benhiller:
Summary: We have an issue with our use of gunicorn where the workers
frequently timeout during boot due to our 29s timeout. That is a
reasonable timeout once they start handling requests, but it doesn't
really make sense for booting, which can be slower without impacting 
request latency.

Unfortunately, gunicorn currently only lets us set a single timeout that
controls both the worker booting and responding to requests. This PR 
adds a new setting to guicorn, `--boot-timeout`, which would allow us 
to set a longer timeout for workers when they are booting.

To preserve backwards compatibility with existing gunicorn
configurations, I have `boot-timeout` default to 0. One difference between
`boot-timeout` and `timeout` is that I have the arbiter interpret a
`boot-timeout` value of 0 as meaning it should fallback to the `timeout`
setting. This should mean this can be committed without changing any
existing behavior, though it does mean there is a difference between
what a value of 0 means for `timeout` vs `boot-timeout`. Let me know what
you think the right trade-off here is.

I tried to adhere to the contribution guidelines here. I have not added
unit tests, but can look into that if this seems like a change you'd
want to merge. I also ran the command to update `settings.rst` but it
seemed to generate other changes as well to things I hadn't touched, so
I left that out for now, but can also add that in assuming the other changes
 are things which would be included once someone else ran that command.
I did run the existing test suite and confirmed they passed after this change.

Relevant Issue: #2196

Test Plan: Added `print` statements to `murder_workers` to print whether
or not the worker was considered booted, and what timeout was applied.
Used that along with a `print` statement in `post_worker_init` to confirm
that the worker was correctly considered booted after it was
initialized.

First, tried a config where `timeout` was set to a non-zero value but
`boot-timeout` was 0 (the default) and confirmed that we used the
`timeout` value before and after the worker was booted.

Next, tried a config where `timeout` was set to a non-zero value and
`boot-timeout` was also set to a non-zero value. Confirmed we used the
`boot-timeout` value during worker initialization, and `timeout`
afterwards.

Finally, tried a config where `timeout` was set to 0 and
`boot-timeout` was set to a non-zero value. Confirmed we used the
`boot-timeout` value during worker initialization, and did not apply a
timeout afterwards. (This seems like it would be an odd configuration, 
but just wanted to check it behaved as expected).

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -28,6 +28,7 @@ Ben Oswald <ben.oswald@root-space.de>
 Benjamin Gilbert <bgilbert@backtick.net>
 Benoit Chesneau <bchesneau@gmail.com>
 Berker Peksag <berker.peksag@gmail.com>
+bhiller <benhiller@gmail.com>
 bninja <andrew@poundpay.com>
 Bob Hagemann <bob+code@twilio.com>
 Bobby Beckmann <bobby@macs-MacBook-Pro.local>
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -100,6 +100,7 @@ class Arbiter(object):
         self.address = self.cfg.address
         self.num_workers = self.cfg.workers
         self.timeout = self.cfg.timeout
+        self.boot_timeout = self.cfg.boot_timeout
         self.proc_name = self.cfg.proc_name
 
         self.log.debug('Current configuration:\n{0}'.format(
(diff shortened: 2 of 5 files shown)


benhiller:
I updated this to use the ctime comparison method rather than the chmod-based approach for checking whether the worker has booted. I know that doesn't address all the feedback you had about this approach on the issue, but this seemed like a strict improvement over using chmod, and it was a small change, so I thought I would go ahead and make it!

benhiller:
Hi @benoitc, just wanted to check if you had a chance to look at this or had any other thoughts on what would make this mergeable. Thanks!

benoitc:
not applicable today. Thanks for the patch anyway.

PR #2558, opened 2021-04-08
Fix descriptions for Arbiter.murder_workers

laggardkernel:
Changed the description for `Arbiter.murder_workers()`. 

In my understanding, this function is used to cleanup workers timed out, which indicates the workers fall into a bad state. Not for cleaning up idle workers. "idle" sounds like the workers have no requests to process and just run in a loop waiting for coming requests. When reading the source code, the original description did confuse me a while.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -485,7 +485,7 @@ class Arbiter(object):
 
     def murder_workers(self):
         """\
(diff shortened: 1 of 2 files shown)


benoitc:
imo the proposal makes the description less accurate. closing the issue then. Thanks for the patch anyway

PR #2570, opened 2021-04-26
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10

### LINKED ISSUE #2569: threading.Thread.setDaemon has been deprecated in favor of setting daemon attribute directly in Python 3.10
Ref : python/cpython#25174

https://github.com/benoitc/gunicorn/blob/cf55d2cec277f220ebd605989ce78ad1bb553c46/gunicorn/reloader.py#L20

https://github.com/benoitc/gunicorn/blob/cf55d2cec277f220ebd605989ce78ad1bb553c46/gunicorn/reloader.py#L77

tirkarthi:
Fixes #2569


salty-horse:
Can this be merged?

PR #2573, opened 2021-04-29
Move init/wrap_socket to handle phase

pipeti:
Proposed to move the ssl wrap_socket to the handle phase so a
misbehaving SSL client only blocks a single thread.
Could happen when handshake is not happening on time or at all
and the client keeps the connection open. Multiple misbehaving
clients can still cause thread exhaustion through blocking and
so a worker timeout.
The sync worker is subject to the same problem
(causing a worker timeout), but there is no simple
improvement for that.

### DIFF
--- a/docs/source/2021-news.rst
+++ b/docs/source/2021-news.rst
@@ -10,8 +10,8 @@ Changelog - 2021
 ===================
 
 - document WEB_CONCURRENCY is set by, at least, Heroku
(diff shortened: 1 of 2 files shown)


benoitc:
Closing due to inactivity. This PR has been open for over 3.5 years without activity. If this change is still needed, please feel free to open a fresh PR against current master. Thank you for your contribution!

Issue #2582, opened 2021-05-09
eventlet v0.30.2 breaks gunicorn with ImportError


goodspark:
I'm using gunicorn and the geventlet worker class. I recently upgraded to eventlet from 0.29.1 to 0.31.0 to address a security vulnerability (https://github.com/eventlet/eventlet/security/advisories/GHSA-9p9m-jm8w-94p2).

Now when my app starts, gunicorn fails:

```
[Traceback (most recent call last):
  File "/app/.heroku/python/lib/python3.8/site-packages/gunicorn/util.py", line 99, in load_class
    mod = importlib.import_module('.'.join(components))
  File "/app/.heroku/python/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/app/.heroku/python/lib/python3.8/site-packages/gunicorn/workers/geventlet.py", line 23, in <module>
    from eventlet.wsgi import ALREADY_HANDLED as EVENTLET_ALREADY_HANDLED
ImportError: cannot import name 'ALREADY_HANDLED' from 'eventlet.wsgi' (/app/.heroku/python/lib/python3.8/site-packages/eventlet/wsgi.py)
```

Looks like this was broken in eventlet 0.30.3 (https://github.com/eventlet/eventlet/commit/df0bc00c3b758b0a632929a7ade2125d0a80c08a#diff-f5505c9146e0d3140cc0c6685771ff0a4f0cbadcb95e8115ce2501744007a9f4L72-L84).

Edit: Looks like https://github.com/benoitc/gunicorn/pull/2581/files already solves this!

benoitc:
fixed in latest master.

PR #2583, opened 2021-05-10
Fix compatibility with latest eventlet

Boring-Mind:
Fix issue with import error while using latest eventlet (>=0.30.3).

Changes made in eventlet: https://github.com/benoitc/gunicorn/pull/2581

### DIFF
--- a/gunicorn/workers/geventlet.py
+++ b/gunicorn/workers/geventlet.py
@@ -17,7 +17,13 @@ else:
 
 from eventlet import hubs, greenthread
 from eventlet.greenio import GreenSocket
-from eventlet.wsgi import ALREADY_HANDLED as EVENTLET_ALREADY_HANDLED
+
+try:
+    from eventlet.wsgi import ALREADY_HANDLED as EVENTLET_ALREADY_HANDLED
+except ImportError:
+    # Since eventlet 0.30.3
+    from eventlet.wsgi import WSGI_LOCAL as EVENTLET_ALREADY_HANDLED
+
 import greenlet
 
 from gunicorn.workers.base_async import AsyncWorker


jamadden:
Thanks for the PR, intended to address #2582.

This is an alternative to the approach taken in #2581, which is meant to work with eventlet both before and after the change.

Boring-Mind:
I'm also not sure that this code works but it gives a basic idea of how it would work.

ozamani9:
while this incompatiabilty with eventlet version be released soon.  we need to upgrade eventlet to 0.31.0 to resolve the vulnerability.   very eager to find out when it will be available.

thanks
omid

benoitc:
> Thanks for the PR, intended to address #2582.
> 
> This is an alternative to the approach taken in #2581, which is meant to work with eventlet both before and after the change. Based on the code there, I'm not sure that the code here is correct (or vice versa).

what do you mean ? This changes allows to handle the changes at load time instead of running a check every time so I would go for it. The only issue  to have with this changes is the need to clarify in the documentation that updating a dependency  like eventlet is like updating gunicorn and requires to send an USR2 signal.  Do you see any blocker vs the other proposal?

benoitc:
> while this incompatiabilty with eventlet version be released soon. we need to upgrade eventlet to 0.31.0 to resolve the vulnerability. very eager to find out when it will be available.
> 
> thanks
> omid

did you test the patch? Any feedback about it would speed a release of course.

temoto:
@Boring-Mind sorry, ALREADY_HANDLED from previous version and WSGI_LOCAL work in completely different way. It wasn't just a rename of variable.

Boring-Mind:
@temoto no problems, I was happy to participate in the development and help to resolve that issue in any way.

mgorny:
Thanks from Gentoo for the patch!

benoitc:
> Thanks from Gentoo for the patch!

which patch?

mgorny:
> > Thanks from Gentoo for the patch!
> 
> which patch?

I meant this pull request.

PR #2592, opened 2021-06-07
Fix description of the post_request hook signature

PeterJCLaw:
I'm guessing that the previous description either pre-dates the current signature or was a copy/paste from pre_request. The new description was established from the signature example and inspection of the calling code.


laggardkernel:
The post_request hook function could be defined with less arguments. `worker`, `req` are must, others are optional.  Check `test_config.test_post_request()` and `config.validate_post_request()` for detail.

@benoitc  Any idea why the CI failed? It happened in my PRs as well.

PeterJCLaw:
Aha, thanks, I'd missed that. I've updated the suggested change.

PeterJCLaw:
Hrm, rebasing on `master` doesn't seem to have fixed CI either.

@benoitc is there anything I can do to help here?

benoitc:
Applied this fix to gunicorn/config.py in commit 380f0978. Thank you for the contribution!

PR #2607, opened 2021-07-08
Fix pylint errors

vigneshwaran-nv-10329:


gunicorn/glogging.py
```diff
@@ -204,9 +204,9 @@ def setup(self, cfg):
             for stream in sys.stdout, sys.stderr:
                 stream.flush()
 
-            self.logfile = open(cfg.errorlog, 'a+')
-            os.dup2(self.logfile.fileno(), sys.stdout.fileno())
-            os.dup2(self.logfile.fileno(), sys.stderr.fileno())
+            with open(cfg.errorlog, 'a+') as self.logfile:
+                os.dup2(self.logfile.fileno(), sys.stdout.fileno())
+                os.dup2(self.logfile.fileno(), sys.stderr.fileno())
```
isidentical, line 209: The problem with this is that, it will set the `self.logfile` to a closed file (since after this context is finalized, it will close the file). So there is no point of setting it to an attribute, since none of the users of that attribute will able to benefit from it.

benoitc: i don't see the point of using a with statement there.  please keep it that way.

gunicorn/workers/gthread.py
```diff
@@ -49,7 +49,8 @@ def init(self):
         if self.parser is None:
             # wrap the socket if needed
             if self.cfg.is_ssl:
-                self.sock = ssl.wrap_socket(self.sock, server_side=True,
+                context = ssl.create_default_context()
```
benoitc, line 52: this should be a separate commit and not related to pylint really as it is using new code.

benoitc:
not applicable anymore.

PR #2539, opened 2021-03-13
util.py: improve import time

hauntsaninja:
I know in general people hesitate to improve import time by deferring imports, but I claim this is worth it! `import pkg_resources` is really, really slow, often measured in seconds (e.g., see https://github.com/pypa/setuptools/issues/510 / the existence of importlib.metadata in the standard library in Python 3.8+).

Import times might not matter too much to gunicorn, but e.g. this impacts the import time of aiohttp, which is used for many things other than servers.

Resolves #2613

PR #2614, opened 2021-07-22
optimization: speed-up sole importing up to 10X

### LINKED ISSUE #2613: Importing gunicorn.config takes a lot of time
The import-time of certain packages are quite important for CLI applications, since even `0.1` seconds on each invocation adds up to something. While I was trying to figure out why importing `aiohttp` takes so long I noticed that the majority of the time spent on importing `gunicorn.config` which imported `gunicorn.utils` which then resulted with the import of `pkg_resources` (which is notoriously known for bad import times when the environment have too many packages);
```
 $ python -Ximporttime -c 'import gunicorn.config' 2>&1 | tail
import time:       424 |        424 |       sysconfig
import time:    215305 |     253223 |     pkg_resources
import time:       173 |        173 |     gunicorn.errors
import time:       113 |        113 |     gunicorn.workers
import time:       242 |        242 |     setproctitle
import time:       301 |     265796 |   gunicorn.util
import time:       229 |        229 |       inotify
import time:        18 |        246 |     inotify.adapters
import time:       244 |        490 |   gunicorn.reloader
import time:      3049 |     293565 | gunicorn.config
```

There is a [very long issue](https://github.com/pypa/setuptools/issues/510) (100+ comments) on the setuptools tracker about why this is slow but the important thing is that, gunicorn doesn't necessarily need to import it every time. There is only a single function which uses this to select entrypoints, so it might be worth considering to inline this import;
https://github.com/benoitc/gunicorn/blob/1299ea9e967a61ae2edebe191082fd169b864c64/gunicorn/util.py#L57-L104

```
 $ python -Ximporttime -c 'import gunicorn.config' 2>&1 | tail
import time:        38 |         38 |       atexit
import time:      1853 |       2203 |     logging
import time:       145 |        145 |     gunicorn.errors
import time:       108 |        108 |     gunicorn.workers
import time:       239 |        239 |     setproctitle
import time:       315 |      12995 |   gunicorn.util
import time:       235 |        235 |       inotify
import time:        18 |        252 |     inotify.adapters
import time:       213 |        465 |   gunicorn.reloader
import time:      2981 |      39848 | gunicorn.config
```

With the inline, on my environment (I have too many packages, but still) it went from `0.3` seconds to `0.039848` seconds. It also speeds up `aiohttp` considerably. 

master;
```
 $ python -Ximporttime -c 'import aiohttp' 2>&1 | tail
import time:       314 |     397838 | aiohttp
```

my branch;
```
 $ python -Ximporttime -c 'import aiohttp' 2>&1 | tail
import time:       292 |     144031 | aiohttp
```

isidentical:
Resolves #2613

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -180,3 +180,4 @@ Xie Shi <xieshi@douban.com>
 Yue Du <ifduyue@gmail.com>
 zakdances <zakdances@gmail.com>
 Emile Fugulin <emilefugulin@hotmail.com>
+Batuhan Taskaya <batuhan@python.org>
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -22,8 +22,6 @@ import time
 import traceback
 import warnings
 
-import pkg_resources
-
 from gunicorn.errors import AppImportError
 from gunicorn.workers import SUPPORTED_WORKERS
 import urllib.parse
@@ -56,6 +54,8 @@ except ImportError:
 
 def load_class(uri, default="gunicorn.workers.sync.SyncWorker",
                section="gunicorn.workers"):
+    import pkg_resources
+
     if inspect.isclass(uri):
         return uri
     if uri.startswith("egg:"):


benoitc:
why closing this PR? The question is not if the release time is long or not. Is this PR OK? Is this solving an issue? Is it helpful? That's the right questions. 

I am re-opening the PR as I think it may be useful. There is no need to import more than what's really needed. It  is also make upgrade of processes easier.

benoitc:
TODO:

- [ ] test with other workers and others addons

isidentical:
I am not really familiar with the test process with other workers/addons, though if you could guide me to the commands to do so I can run benchmarks but from what I can say this patch does not change any behavior.

m-novikov:
I think it is better to wrap `pkg_resources` in the function because it's only needed in the specific circumstances. See #2616 description.

hauntsaninja:
I made the exact same PR a couple months ago https://github.com/benoitc/gunicorn/pull/2539 :D 

> Is this solving an issue? Is it helpful?

As mentioned in my PR:

> I know in general people hesitate to improve import time by deferring imports, but I claim this is worth it! import pkg_resources is really, really slow (e.g., see pypa/setuptools#510 / the existence of importlib.metadata in the standard library in Python 3.8+).
Import times might not matter too much to gunicorn, but e.g. this impacts the import time of aiohttp, which is used for many things other than servers.

For instance, I have a cloud CLI tool that uses aiohttp (https://github.com/hauntsaninja/boostedblob). If `gunicorn` is present in a user's environment, because of the `pkg_resources` import, this slows down every CLI command by a couple seconds.

isidentical:
> I made the exact same PR a couple months ago #2539 :D

Oh, completely missed that one. Closing this as duplicate! 

> For instance, I have a cloud CLI tool that uses aiohttp (https://github.com/hauntsaninja/boostedblob). If gunicorn is present in a user's environment, because of the pkg_resources import, this slows down every CLI command by a couple seconds. pkg_resources is really exceptionally slow to import, so even if you're hesitant to play the game of deferring imports, I claim in this case it's worth making an exception.

That is the same issue we experience on another CLI tool ([dvc](https://github.com/iterative/dvc)), which lead me to send this fix since the additional cost is extremely significant.

PR #2615, opened 2021-07-22
ci: fix CI failures due to grp import on windows

isidentical:
Example failing build: https://ci.appveyor.com/project/benoitc/gunicorn/builds/40090320/job/5vy7s88qiu0dlv0n

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -7,7 +7,6 @@
 
 import argparse
 import copy
-import grp
 import inspect
 import os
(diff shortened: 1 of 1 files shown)


isidentical:
Apparently there are a lot of other stuff missing due to incomplete windows support, so closing this since it doesn't seem like a trivial thing to do the fix the CI. I still am not sure though why there is no linux CI for this project / and even with an incomplete windows support why it only tests on windows.

PR #2616, opened 2021-07-23
Defer import of pkg_resources until it is actually required

m-novikov:
Import performance of pkg_resources is slow due to path enumeration
done during the import.

See the issue https://github.com/pypa/setuptools/issues/510

To avoid unnecessary work we only discover entrypoints once we
know that worker uri specified with "egg:" prefix

For my particular case import takes ~1s with this patch it's 60ms

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -22,7 +22,6 @@ import time
 import traceback
 import warnings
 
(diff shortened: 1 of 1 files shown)


m-novikov:
Didn't see #2614

PR #2620, opened 2021-08-06
Improve access-log-format documentation section

talkerbox:
Added more clearly info about user name identifier (u). That will save a bit of time for developers due log configuration, like this guy https://stackoverflow.com/questions/51660511/gunicorn-access-logs-shows-empty-user


benoitc:
thanks!

PR #2628, opened 2021-08-17
fix: log access log when a client closed the connection

### LINKED ISSUE #1060: Access Log line missing when client times out

ieralt:
When a client closed the connection, resp.write(item) would raised a BrokenPipeError which prevents the access from being logged.

Fixes #1060.

### DIFF
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -115,9 +115,9 @@ class AsyncWorker(base.Worker):
                     for item in respiter:
                         resp.write(item)
                 resp.close()
+            finally:
                 request_time = datetime.now() - request_start
                 self.log.access(resp, req, environ, request_time)
-            finally:
                 if hasattr(respiter, "close"):
                     respiter.close()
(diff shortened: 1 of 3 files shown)


ikonst:
Seems very similar to https://github.com/benoitc/gunicorn/pull/2392. In my change, I've tried to preserve the existing behavior, where an exception raised in the logger would not prevent `close()` execution.

Issue #2630, opened 2021-08-18
gunicorn run tornado app failed


ljcjc:
gunicorn version 20.1.0 and worker class is tornado
tornado version 6.0.0
app class  is tornado.web.Application

gunicorn run this app and rais error below
File "/.venv/lib/python3.8/site-packages/tornado/wsgi.py", line 113, in __call__
    app_response = self.wsgi_application(
TypeError: __call__() takes 2 positional arguments but 3 were given"

I found that the causes of this problem in gunicorn/workers/gtornado.py line113 and 114
`
elif not isinstance(app, WSGIContainer):
       app = WSGIContainer(app)
`

alviezhang:
Reproduce this bug too, tested version 6.0 and version 6.1.

This line breaks code execution.
https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gtornado.py#L113

duanhongyi:
This RP is really a problem. I think he can be like this:
```
        if tornado.version_info[0] < 6:
            if not isinstance(app, tornado.web.Application) or \
                    isinstance(app, tornado.wsgi.WSGIApplication):
                app = WSGIContainer(app)
        elif not isinstance(app, WSGIContainer) and \
                not isinstance(app, tornado.web.Application):
            app = WSGIContainer(app)
```

duanhongyi:
I'll submit an RP to fix this later.

PR #2633, opened 2021-08-25
fix:  gunicorn run tornado app failed

duanhongyi:
#2630 
#2418

### DIFF
--- a/gunicorn/workers/gtornado.py
+++ b/gunicorn/workers/gtornado.py
@@ -110,7 +110,8 @@ class TornadoWorker(Worker):
             if not isinstance(app, tornado.web.Application) or \
             isinstance(app, tornado.wsgi.WSGIApplication):
                 app = WSGIContainer(app)
-        elif not isinstance(app, WSGIContainer):
+        elif not isinstance(app, WSGIContainer) and \
+        not isinstance(app, tornado.web.Application):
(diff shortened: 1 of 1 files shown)


alviezhang:
This `tornado` worker is intent to support tornado web framework, I don't know if it is appropriate to support Django within this module.

And there are no tests in `tests/workers` directory too, we might have another case in order to fix `django` problem and eventually breaks `tornado`.

benoitc:
do you have any log reproducing the issue?

duanhongyi:
Because the application of the tornado web framework does not meet the WSGI specification. And tornado server can run it directly. So we don't need to wrap tornado web application with WSGIContainer.


The error is as follows:

```
ERROR:tornado.application:Uncaught exception
  File "/root/tmp/tornado/lib/python3.8/site-packages/tornado/http1connection.py", line 273, in _read_message
    delegate.finish()
  File "/root/tmp/tornado/lib/python3.8/site-packages/tornado/httpserver.py", line 280, in finish
    self.request_callback(self.request)
  File "/root/tmp/tornado/lib/python3.8/site-packages/tornado/wsgi.py", line 113, in __call__
    app_response = self.wsgi_application(
TypeError: __call__() takes 2 positional arguments but 3 were given
```

PR #2635, opened 2021-08-27
Update wsgiapp.py

tal-zvon:
Removed unnecessary import.

I was wondering what it did, so I [asked about it](https://stackoverflow.com/q/68947408/6423456).
It appears to be unnecessary, and is only there because it was needed before a refactor moved the function into the same module as the class.

I tried removing the import, and nothing broke when I ran gunicorn.


benoitc:
this can be used when loading module manually. Imo let's keep it there.

PR #2636, opened 2021-08-27
Step 1 of 2 in enabling gunicorn to run on Windows

bjones1:
Per #524, this is the first step in enabling gunicorn to run on Windows.  The missing piece: getting a non-blocking `os.pipe` running on Windows (see notes below). My implementation decisions:
- In `arbiter.py`, only add signals supported by the current platform. (Windows doesn't support many of the Unix signals).
- In `config.py`, disallow setting `user` or `group` for Windows only.
- In `sock.py`, don't support Unix sockets on Windows. (While Windows 10 now supports `AF_UNIX`, [Python doesn't](https://bugs.python.org/issue33408)).
- In `utils.py`, use `os.set_inheritable` instead of direct calls to fcntl. The [underlying C code on Unix](https://github.com/python/cpython/blob/main/Python/fileutils.c#L1406) calls fcntl, so this should have no impact on Unix.

Based on the [contribution guidelines](https://github.com/benoitc/gunicorn/blob/master/CONTRIBUTING.md), I have:

- Named this branch `524-windows-1-of-2`, considering Windows support a feature rather than a bugfix.
- There are (not yet) relevant unit tests, since this doesn't (yet) run on Windows, and creation of a Windows test suite probably deserves some discussion.
- Wrote (IMHO) clean code.
- Provided a clear description and referenced #524.
- No documentation updates apply: it doesn't fully support Windows yet, and there's no mention I can find of Windows support in the docs to change.
- I ran tests using Ubuntu using Python 3.8, since the Appveyor tests fail (why do we have Appveyor CI for gunicorn?) and the Travis tests are disabled (???). In the master branch, all the tests in `tests/test_valid_request.py` failed on the master branch; these changes didn't produce any additional failures.
- Added my name to THANKS.

However, I haven't discussed the design. I'm a bit confused here -- the link to the mailing list took me to the [community](https://gunicorn.org/#community) page, where clicking on the mailing list link brings me to the [project page](https://github.com/benoitc/gunicorn/projects/3). I see the [card for Windows support](https://github.com/benoitc/gunicorn/projects/3#card-1863696), but this links to issue #524. Is that issue the location for the discussion?

# Non-blocking `os.pipe` for Windows
This is the most difficult part of the code. Here are my notes on what needs to be done:

The [Python implementation of os.pipe](https://github.com/python/cpython/blob/main/Modules/posixmodule.c#L10105) calls [CreatePipe](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-createpipe). Unfortunately, `CreatePipe` [only provides blocking I/O](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-createpipe#remarks).

However, [named pipes do allow non-blocking I/O](https://docs.microsoft.com/en-us/windows/win32/ipc/synchronous-and-overlapped-input-and-output). Windows-specific Python code using the win32 API should therefore provide a replacement for the current Unix-only approach. Several code snippets ([1](https://github.com/PeiSeng/Named_Pipe_Communication), [2](https://stackoverflow.com/questions/50040067/how-to-create-a-non-blocking-named-pipe-from-python)) might provide guidance on writing these routines.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -33,6 +33,7 @@ Bob Hagemann <bob+code@twilio.com>
 Bobby Beckmann <bobby@macs-MacBook-Pro.local>
 Brett Randall <javabrett@gmail.com>
 Brian Rosner <brosner@gmail.com>
+Bryan A. Jones <bjones@ece.msstate.edu>
 Bruno Bigras <bigras.bruno@gmail.com>
 Caleb Brown <git@calebbrown.id.au>
 Chris Adams <chris@improbable.org>
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -41,8 +41,11 @@ class Arbiter(object):
 
     # I love dynamic languages
     SIG_QUEUE = []
-    SIGNALS = [getattr(signal, "SIG%s" % x)
-               for x in "HUP QUIT INT TERM TTIN TTOU USR1 USR2 WINCH".split()]
+    SIGNALS = []
+    for x in "HUP QUIT INT TERM TTIN TTOU USR1 USR2 WINCH".split():
+        signal = getattr(signal, "SIG%s" % x, None)
+        if signal:
+            SIGNALS.append(signal)
     SIG_NAMES = dict(
         (getattr(signal, name), name[3:].lower()) for name in dir(signal)
         if name[:3] == "SIG" and name[3] != "_"
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -7,10 +7,8 @@
 
 import argparse
 import copy
-import grp
 import inspect
 import os
-import pwd
 import re
 import shlex
 import ssl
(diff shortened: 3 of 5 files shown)


benoitc:
thanks for the change. I will have a deeper look on it over the weeke-end. os.pipe is really used for signaling . We can abstract signaling in a separate module to make it easier and replace by a compatibility module on windows. Part lf the plan is also to replace it soon by another way to handle it to bring more messaging capabilties.

We don’t need to have all the features on windows. if  unix sockets can’t be done on it for example we should just raise an issue.

benoitc:
i will replace travis CI asap.

bjones1:
> `os.pipe` is really used for signaling . We can abstract signaling in a separate module to make it easier and replace by a compatibility module on windows. Part lf the plan is also to replace it soon by another way to handle it to bring more messaging capabilties.

@benoitc, thanks for the quick feedback! Finding a more abstract signaling approach makes a lot of sense -- I'll look for your feedback either here on in #524 re: inter-thread vs. inter-process. If you could provide a bit of feedback on your future plans, that will help guide me in searching for a Windows-friendly implementation. I'm happy to message/chat if you'd like to think about this in real time.

> We don’t need to have all the features on windows. if unix sockets can’t be done on it for example we should just raise an «unsupported platform » error.

Great! I'll look at adding that exception to this PR.

bjones1:
> i will replace travis CI asap.

Thanks!

PR #2643, opened 2021-08-31
Platform-independent inter-process communication

bjones1:
This PR follows #2636 in the process of enabling gunicorn to run on Windows. This draft PR provides an opportunity for others to view my progress. Expect force pushes/rebases/etc. while this is a draft.

Progress: all pipe-based IPC is now encapsulated in a class.
Plans: move this class to a [multi-platform approach](https://docs.python.org/3/library/multiprocessing.html#multiprocessing-listeners-clients).

Problems: `os.fork` is the core of gunicorn's multiprocessing approach, and should never be changed for Unix. For Windows only, we need to look for a nice way to start a new process then send all relevant data copied by `os.fork` to that new process. I propose completing this change then placing that work in a separate PR.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -33,6 +33,7 @@ Bob Hagemann <bob+code@twilio.com>
 Bobby Beckmann <bobby@macs-MacBook-Pro.local>
 Brett Randall <javabrett@gmail.com>
 Brian Rosner <brosner@gmail.com>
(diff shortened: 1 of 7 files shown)


sarthakpati:
@bjones1 thanks for all the work here. Do you happen to have any news on this?

bjones1:
Unfortunately, I ran out of time to pursue this -- I abandoned this work.

PR #2653, opened 2021-09-15
Added optional socket timeout for sync and gthread workers.

### LINKED ISSUE #2652: Feature request: add optional socket timeout
Currently it is possible for misbehaving client to block workers for quite some time, from 30 seconds to eternity.

Having a separate option to set timeout for socket read/write would minimize the impact of these, while still allowing longer work to be executed after socket I/O.  For example option `--socket-timeout=5` could be implemented to limit the time spent with socket I/O, while keeping `--timeout=30` as separate timeout for the worker to process the complete request.

Following table summarizes current behavior:

|   | sync | gthread | eventlet | gevent | tornado |
|----------------------|--------|------------|------------|-----------|-----------|
| client connects but does not send HTTP request | 30s | infinity | 2s | 2s | infinity |
| client connects but does not proceed with TLS handshake | 30s | infinity | 2s | 2s | infinity |
| client sends request with keep-alive, then stays idle | 0s | 2s | 2s | 2s| infinity |

* 30 seconds is from default `--timeout`
* 2 seconds is from default `--keepalive`
* 0 seconds means immediately disconnected (according to docs, sync intentionally ignores keep-alive)

tsaarni:
Fixes #2652

Signed-off-by: Tero Saarni <tero.saarni@est.tech>

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -1324,6 +1324,25 @@ load.
 
 Must be a positive integer. Generally set in the 64-2048 range.
 
+.. _socket-timeout:
+
+``socket_timeout``
+~~~~~~~~~~~~~~~~~~
+
+**Command line:** ``--socket-timeout INT``
+
+**Default:** ``0``
+
+Timeout for reading from the socket associated with the client.
+
+Value is a positive number of seconds or 0. Setting it to 0 has the
+effect of infinite timeouts by disabling socket timeouts entirely.
+Worker level timeout still applies.
+
+.. note::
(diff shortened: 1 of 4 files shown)


tsaarni:
Alternative approach would be to re-use `--keepalive` for `socket.settimeout()` on sync and gthread.  I guess that would kind of align with the implementation in eventlet and gevent (see https://github.com/benoitc/gunicorn/issues/2652#issue-996948464)

benoitc:
see related ticket dor xommenta. Thanka for the patch anyway!

PR #2656, opened 2021-09-20
fix chdir documentation typo

magicpotion:
just a typo fix in documentation

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -976,9 +976,11 @@ Set the ``SO_REUSEPORT`` flag on the listening socket.
 
 **Command line:** ``--chdir``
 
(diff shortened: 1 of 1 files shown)


javabrett (approved):
Thanks for the fix.

Noting that this doc fragment regularly regresses as the default definition is via `getcwd()`:

https://github.com/benoitc/gunicorn/blob/60d0474a6f5604597180f435a6a03b016783885b/gunicorn/config.py#L1051-L1056

... which gets eval-ed when docs are auto-generated:

https://github.com/benoitc/gunicorn/blob/60d0474a6f5604597180f435a6a03b016783885b/docs/gunicorn_ext.py#L50-L54

Not sure what can be done about that though.

PR #2665, opened 2021-10-09
integral? that took me a while to figure out

jettero:
integral: "In mathematics, an integral assigns numbers to functions in a way that describes displacement, area, volume, and other concepts that arise by combining infinitesimal data."

integer: an int type

### DIFF
--- a/gunicorn/http/body.py
+++ b/gunicorn/http/body.py
@@ -18,7 +18,7 @@ class ChunkedReader(object):
 
     def read(self, size):
         if not isinstance(size, int):
-            raise TypeError("size must be an integral type")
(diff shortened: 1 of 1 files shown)


benoitc:
simply saying there was a typo is enough ;) 😅 thanks for the fix!

jettero:
> simply saying there was a typo is enough ;) sweat_smile thanks for the fix!

Not necessarily... the [thought bug](https://docs.python.org/3/library/numbers.html#numbers.Integral) also exists in python core also. It's just as wrong there, but I think it's there to stay.

benoitc:
thanks!

Issue #2564, opened 2021-04-21
RuntimeError due to a warning being logged on SIGHUP


belawaeckerlig:
Hi,

gunicorn might crash due to a warning being logged in case of a SIGHUP. I think this was introduced by this commit: b695b497b9b8e7351808848ab9e342219601742a

Here is the Traceback of the error:
```
Traceback (most recent call last):
  File "/usr/lib/python3.9/logging/__init__.py", line 1082, in emit
    stream.write(msg + self.terminator)
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.9/logging/__init__.py", line 1082, in emit
    stream.write(msg + self.terminator)
  File "/usr/local/lib/python3.9/dist-packages/gunicorn/arbiter.py", line 242, in handle_chld
    self.reap_workers()
  File "/usr/local/lib/python3.9/dist-packages/gunicorn/arbiter.py", line 530, in reap_workers
    self.log.warning(
  File "/usr/local/lib/python3.9/dist-packages/gunicorn/glogging.py", line 261, in warning
    self.error_log.warning(msg, *args, **kwargs)
  File "/usr/lib/python3.9/logging/__init__.py", line 1454, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "/usr/lib/python3.9/logging/__init__.py", line 1585, in _log
    self.handle(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 1595, in handle
    self.callHandlers(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 1657, in callHandlers
    hdlr.handle(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 948, in handle
    self.emit(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 1087, in emit
    self.handleError(record)
  File "/usr/lib/python3.9/logging/__init__.py", line 1000, in handleError
    sys.stderr.write('--- Logging error ---\n')
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib/python3.9/logging/__init__.py", line 1082, in emit
    stream.write(msg + self.terminator)
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>

[...]

PR #2667, opened 2021-10-16
Fix RuntimeError caused by logging in the SIGCHLD signal handler

BoasKluiving:
This PR is to fix an issue introduced by b695b49. This issue is also described in #2564. The problem is that we are logging something in the signal handler, which is not reentrant.

The issue occurs if Gunicorn happens to be logging something (for example [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L501)) and gets interrupted by SIGCHLD at that moment. This invokes the signal handler handle_chld()  ([here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L240])), which calls reap_workers().

reap_workers() however logs a warning [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L530-L534), causing the following RuntimeError:
`RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>`

Operations on BufferedWriter are unfortunately not reentrant, so we can only fix this by (a) not logging in the signal handler or (b) only setting a flag in signal handler and do the processing later.

This PR fixes the problem by removing the log message in the signal handler.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -508,6 +508,7 @@ class Arbiter(object):
         """\
         Reap workers to avoid zombie processes
         """
+        # Method is called by SIGCHLD signal handler; we cannot log anything
         try:
             while True:
                 wpid, status = os.waitpid(-1, os.WNOHANG)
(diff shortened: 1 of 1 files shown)


javabrett:
@BoasKluiving thanks for the PR.  b695b497b9b8e7351808848ab9e342219601742a was reverted by 76f8da24cbb992d168e01bda811452bcf3b8f5b3 having the same effect, so closing this PR.

PR #2677, opened 2021-11-02
GitHub Action to run tox

cclauss:
___GitHub Actions test results:___ https://github.com/cclauss/gunicorn/actions

Because Travis CI seems to be on vacation... https://travis-ci.org/github/benoitc/gunicorn

___TypeError: cannot set 'is_timeout' attribute of immutable type 'TimeoutError'___
```
============================= test session starts ==============================
platform linux -- Python 3.10.0, pytest-6.2.5, py-1.10.0, pluggy-1.0.0
cachedir: .tox/py/.pytest_cache
rootdir: /home/runner/work/gunicorn/gunicorn, configfile: setup.cfg, testpaths: tests/
plugins: cov-3.0.0
collected 224 items

[ ... ]
tests/workers/test_geventlet.py F                                        [ 99%]
tests/workers/test_ggevent.py .                                          [100%]

=================================== FAILURES ===================================
_________________________________ test_import __________________________________

    def test_import():
>       __import__('gunicorn.workers.geventlet')

tests/workers/test_geventlet.py:7: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
gunicorn/workers/geventlet.py:10: in <module>
    import eventlet
.tox/py/lib/python3.10/site-packages/eventlet/__init__.py:17: in <module>
    from eventlet import convenience
.tox/py/lib/python3.10/site-packages/eventlet/convenience.py:7: in <module>
    from eventlet.green import socket
.tox/py/lib/python3.10/site-packages/eventlet/green/socket.py:4: in <module>
    __import__('eventlet.green._socket_nodns')
.tox/py/lib/python3.10/site-packages/eventlet/green/_socket_nodns.py:11: in <module>
    from eventlet import greenio
.tox/py/lib/python3.10/site-packages/eventlet/greenio/__init__.py:3: in <module>
    from eventlet.greenio.base import *  # noqa
.tox/py/lib/python3.10/site-packages/eventlet/greenio/base.py:32: in <module>
    socket_timeout = eventlet.timeout.wrap_is_timeout(socket.timeout)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

base = <class 'TimeoutError'>

    def wrap_is_timeout(base):
[...]

PR #2679, opened 2021-11-03
Python's fcntl, grp, and pwd modules are Unix-only.

cclauss:
These Python Standard Library modules are Unix-only:
1. https://docs.python.org/3/library/fcntl.html
2. https://docs.python.org/3/library/grp.html

Make modifications require to have pytest pass on Windows.


cclauss:
Closing in favor of #2677

PR #2680, opened 2021-11-03
Use Appveyor Ubuntu, not Appveyor Windows

cclauss:
Windows is not ready for testing!!!
* Python's `fcntl`, `grp`, `pwd`, `os.geteuid()`, and `socket.AF_UNIX` are all ___Unix-only___.


benoitc:
Hi thanks for the patch but appveyor CI is to test deployement of gunicorn on windows. For linux , we still have to replace Travis by something else. not necessarily github actions though.

Issue #1493, opened 2017-04-03
Clarify what/how timeout and graceful_timeout work


tuukkamustonen:
(Sorry for the monologue here: simple things got complicated and I ended up digging through the stack. Hopefully what I've documented is helpful for the reader, however.)

As I've understood, by default:

- After `30` seconds (configurable with `timeout`) of request processing, gunicorn master process sends `SIGTERM` to the worker process, to initiate a graceful restart.
- If worker does not shutdown during another `30` seconds (configurable with `graceful_timeout`), master process sends `SIGKILL`. Seems like this signal is also sent when the worker _does_ gracefully shutdown during the `graceful_timeout` period (https://github.com/benoitc/gunicorn/commit/d1a09732256fa8db900a1fe75a71466cf2645ef9).

The questions:

- Are the signals correct?
- What actually happens when gunicorn (sync) worker receives these signals? How does it tell WSGI app that the signal was caught and something should happen (ok I assume it just "passes it on")?
- How does, for example, Flask handle `SIGTERM` signal - in practice, what happens during request processing? Does it just set a flag for the WSGI application (on werkzeug level) that it should shutdown after the request processing is complete? Or does `SIGTERM` already somehow affect ongoing request processing - kill IO connections or something to speed up request processing...?

On `SIGKILL`, I guess request processing is just forcefully aborted.

I could file a tiny PR to improve docs about this, if I get understanding how things actually work.

tuukkamustonen:
Hmm, I think https://github.com/benoitc/gunicorn/issues/1236#issuecomment-254059927 confirms my assumptions about `SIGTERM` simply setting worker to shutdown after request processing is complete (and setting worker not to accept any new connections).

tuukkamustonen:
Seems like how I interpreted `timeout` and `graceful_timeout` is wrong. Both periods actually refer to time at the start of request processing. So, by default, because both settings are set to `30` seconds, there is no graceful restart enabled. If I do something like `--graceful-timeout 15 --timeout 30` that should mean graceful restart is initiated at 15 seconds and worker is force killed at 30 seconds if request did not complete before that.

However, it seems like if response is returned between `graceful_timeout` and `timeout`, then the worker is not restarted after all? Shouldn't it:

I tested by `app.py`:

    import time
    from flask import Flask

    app = Flask(__name__)
    
    @app.route('/foo')
    def foo():
        time.sleep(3)
        return 'ok'

[...]

PR #2686, opened 2021-11-12
docs: gthread is a sync worker

jpmckinney:
https://github.com/benoitc/gunicorn/issues/1493#issuecomment-321461614

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -38,12 +38,6 @@ applications are programmed.
 closed after response has been sent (even if you manually add ``Keep-Alive``
 or ``Connection: keep-alive`` header in your application).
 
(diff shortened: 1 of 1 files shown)


benoitc:
gthread is not an asyncio worker but is an async worker. That should proabbly fiexed by adding it as a separate async worker. Though I am not sure why we document it. We should rather always use it using the threads setting.

benoitc:
the gthreath work is not an asyncio worker but still an async worker though; i know it was a long time but maybe you can make the change to move that information to Async workers? I will keep it open .

jpmckinney:
I've created a subsection of Async workers for Gthread workers.

PR #2687, opened 2021-11-12
Fix sendfile behaviour for open files with non-zero offset

### LINKED ISSUE #2669: WSGI sendfile does not respect file current offset
Hi,

on Linux, I need to stream a regular file from the disk starting from a specific offset.

According to Django doc, when you use a FileReponse you can seek your file to a specific offset and only the remaining part of the file will be streamed from this offset (just have to take care of the Content-Length).

It's working with the internal Django development server but while using gunicorn the file is streamed from the offset 0.

* OS: Debian Buster
* Python: python 3.7.3
* Virtual environment packages:
   ```
   asgiref==3.4.1
   Django==3.2.8
   gunicorn==20.1.0
   pkg-resources==0.0.0
   pytz==2021.3
   sqlparse==0.4.2
   typing-extensions==3.10.0.2
   ```

Here is my view from by "buggy" Django project:

```
import os

from pathlib import Path

from django.http import FileResponse

SRC = Path(__file__).parent / 'src'

if not SRC.is_file():
    with SRC.open('wt') as sample:
        sample.write(''.join(map(str, range(10))))
        sample.write('\n')

def get(request):
    try:
        offset = int(request.GET.get('offset') or 0)
    except ValueError:
        offset = 0

    src = SRC.open('rb')
    src.seek(offset)
    reply = FileResponse(src)
    reply['Content-Length'] = str(SRC.stat().st_size - offset)
    return reply
````

Result with `manage.py runserver`:

```
$ curl http://127.0.0.1:8000/?offset=5
56789
$
```

Result with `gunicorn buggy.wsgi -b 127.0.0.1:9000`:

```
curl http://127.0.0.1:9000/?offset=5
012345$
```

I resolve my case by changing in `gunicorn.wsgi.Response.sendfile`:

```
375c375
<         self.sock.sendfile(respiter.filelike, count=nbytes)
---
>         self.sock.sendfile(respiter.filelike, count=nbytes, offset=offset)
```

New result with `gunicorn buggy.wsgi -b 127.0.0.1:9000`:

```
curl http://127.0.0.1:9000/?offset=5
56789
$
```

Looking to file history, it seems the bug has been introduced while switching from os.sendfile to socket.sendfile in commit 2d40e6daceb9735d27bb91d9c32743695de8e01c

Thanks,

PS: while it's not a problem in my case and Django, at line 380, why do you seek back the file descriptor to the original offset before leaving `sendfile` ?

vytas7:
Closes #2669

This issue was caused by Gunicorn not passing the _offset_ parameter to [`socket.sendfile()`](https://docs.python.org/3/library/socket.html#socket.socket.sendfile), which is in turned caused by a deeper issue of the latter being a sloppy wrapper around the [`sendfile`](https://linux.die.net/man/2/sendfile) syscall, which takes an offset pointer parameter, and the default `NULL` values indicates that the current offset should be preserved; however Python's default of 0 means to spool back to the first byte of the file.

@benoitc I don't know what is the best way of adding a test for this -- the majority of current tests run by `tox` seem to be quite simple unit tests, whereas in this case it would probably make more sense to write a system test by spinning up a real Gunicorn process. If you pointed to a good way of adding such a test, I could add it. I didn't find any references to `sendfile` in the current test suite.

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -372,7 +372,7 @@ class Response(object):
             chunk_size = "%X\r\n" % nbytes
             self.sock.sendall(chunk_size.encode('utf-8'))
 
-        self.sock.sendfile(respiter.filelike, count=nbytes)
+        self.sock.sendfile(respiter.filelike, offset=offset, count=nbytes)
 
         if self.is_chunked():
             self.sock.sendall(b"\r\n")


benoitc:
thanks for the PR. for the test we will figure soon as it's moving outside of github.

PR #2689, opened 2021-11-15
Allow to bind to IPv6 address specified by name, not address

gozdal:


gunicorn/sock.py
```diff
@@ -176,30 +176,50 @@ def create_sockets(conf, log, fds=None):
...
+                sock_class = TCP6Socket
+            elif sock_family == socket.AF_UNIX:
+                sock_class = UnixSocket
+            else:
+                log.warning("Ignoring unknown socket family: %s", str(sock_family))
+                continue
+            sock = None
+            for i in range(5):
+                try:
+                    sock = sock_class(sock_addr, conf, log)
```
tilgovi, line 202: This is `sock_type` elsewhere in this file, and before this change. Maybe we keep the name? That it's a constructor is immaterial here, it could just as well be a factory function.

gunicorn/sock.py
```diff
@@ -176,30 +176,50 @@ def create_sockets(conf, log, fds=None):
...
+                        time.sleep(1)
+                else:
+                    break
 
-        if sock is None:
-            log.error("Can't connect to %s", str(addr))
-            sys.exit(1)
+            if sock is not None:
+                listeners.append(sock)
+                some_sock_succeeded = True
```
tilgovi, line 218: Should we bind to all addresses returned from `getaddrinfo`, or just the first one that succeeds?

gunicorn/sock.py
```diff
@@ -176,30 +176,50 @@ def create_sockets(conf, log, fds=None):
...
 
-        listeners.append(sock)
+    if not some_sock_succeeded:
```
tilgovi, line 220: I'd like to see `some_sock_succeeded` moved inside the outermost loop, and an exit if we fail to bind any of the specified bind addresses to at least one of their infos. I think the indentation of this check and the `some_sock_succeeded = False` should be done for each specified bind address. That would be consistent with the previous behavior.

tilgovi (changes requested):

gunicorn/sock.py
gozdal: I think we should bind to them all, try at least. IIUC `getaddrinfo` returns the result in unspecified order. It means that one call may return `[IPv6 address, IPv4 address]` and the next `[IPv4 address, IPv6 address]`. If binding only to the first one we'd be introducing non-determinism between runs.

gunicorn/sock.py
gozdal: > That would be consistent with the previous behavior.

Indeed, though I wonder if it's the best behavior - if a hostname resolves to both IPv4 and IPv6 or the user passed multiple addresses, failure to bind to any of them results in `sys.exit(1)`. 

I guess this could be left in the name of backwards compatibility but doing `sys.exit(1)` if any bind failed seemed pretty harsh to me.

gozdal:
@tilgovi can I do something to help getting this merged?

gozdal:
@tilgovi @benoitc 
can I do something to help getting this merged?

ddzialak:
Rebased and resolved conflicts

gozdal:
> `UnicodeError: encoding with 'idna' codec failed`

How can you get `UnicodeError` in this code?

gunicorn/sock.py
```diff
@@ -177,30 +177,49 @@ def create_sockets(conf, log, fds=None):
...
+                try:
+                    sock = sock_type(sock_addr, conf, log)
+                except OSError as e:
+                    if e.args[0] == errno.EADDRINUSE:
+                        log.error("Connection in use: %s", str(sock_addr))
+                    if e.args[0] == errno.EADDRNOTAVAIL:
+                        log.error("Invalid address: %s", str(sock_addr))
+                    if i < 5:
+                        msg = "connection to {addr} failed: {error}"
+                        log.error(msg.format(addr=str(sock_addr), error=str(e)))
```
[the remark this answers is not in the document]
gozdal: Good catch, it wasn't fully reverted but debug was missing.

gozdal:
> That is without the last commit, no? With that the continue´d errors will not go to sys.exit - putting the sync worker into a busy loop.

Indeed. You know this code much better than me ;) `continue` was a mistake there, changed it into `addrinfos = []`

PR #2690, opened 2021-11-15
add debug logging to gthread

forrestrice:


gunicorn/__init__.py
```diff
@@ -3,7 +3,7 @@
 # This file is part of gunicorn released under the MIT license.
 # See the NOTICE for more information.
 
-version_info = (20, 1, 0)
+version_info = (20, 1, 0, "dev1")
```
benoitc, line 6: can you remove this change?  this not needed until it's in master.

gunicorn/http/body.py
```diff
@@ -56,8 +57,14 @@ def parse_trailers(self, unreader, data):
 
     def parse_chunked(self, unreader):
         (size, rest) = self.parse_chunk_size(unreader)
+        while_count_size_gt_zero = 0
         while size > 0:
+            print(f"{threading.currentThread().getName()}: size > 0. parse_chunked count={while_count_size_gt_zero}", flush=True)
+            while_count_size_gt_zero += 1
+            while_count_size_gt_len_rest = 0
             while size > len(rest):
+                print(f"{threading.currentThread().getName()}: size > len(rest). parse_chunked count={while_count_size_gt_len_rest}", flush=True)
```
benoitc, line 66: can your replace printf by  the debug function from the log handler? This will be more useful.

benoitc (changes requested):

benoitc:
this seems too much verbose at production level . I would suggest to revisit it by adding some metrics instead.

PR #2691, opened 2021-11-17
Update documentation for `worker-connections` setting

egrim:
The thread worker type is affected by the `worker_connections` setting so this updates the documentation to reflect that (see [here](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gthread.py#L198)).

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -1415,7 +1415,7 @@ This setting only affects the Gthread worker type.
 
 The maximum number of simultaneous clients.
 
-This setting only affects the Eventlet and Gevent worker types.
+This setting only affects the Eventlet, Gevent, and Thread worker types.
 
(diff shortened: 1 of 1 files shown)


benoitc:
I am not sure about that change. Indeed the supported way to use it is to use the `--threads` option. What do others think about it?

egrim:
_First a quick aside since I've got you all looking at this PR: **thank you so much** for this lovely project and for all the time and effort you all put into it._

As for the PR: I've suggested this change because the documentation for `worker_connections` makes the rather specific claim that:
> This setting only affects the Eventlet and Gevent worker types.

This statement is not accurate (i.e., changing the `worker_connections` setting will affect the thread worker type) so I'd humbly suggest the documentation needs some kind of change.

PR #2712, opened 2022-01-06
pass pylint

Excalartur:


gunicorn/util.py
```diff
@@ -547,12 +547,12 @@ def seed():
         random.seed('%s.%s' % (time.time(), os.getpid()))
 
 
-def check_is_writeable(path):
+def check_is_writable(path):
     try:
-        f = open(path, 'a')
+        with open(path, 'a', encoding="utf-8") as f:
```
benoitc, line 552: is encoding really needed?

benoitc: @Excalartur ?

Excalartur: The linter fix addresses the pylint warning W1514 (unspecified-encoding)
 as explained in PEP 597: https://peps.python.org/pep-0597/
If you believe it's not critical, we can add this warning to the ignore list. What do you think?

benoitc: @Excalartur Maybe a little too cautious but I'm not sure it own't introduce some unknown breakage. So let's ignore it for now. Can you make the change?

Excalartur: I suspect that I have messed up with the rebase 😅
Can this PR be saved or is it doomed?

benoitc: we can keep it there I think it will work (unsure :)

Excalartur: Created a new PR #2996  :)

PR #2718, opened 2022-01-13
Update wsgi.py

theawgur:
Fix issue when gunicorn throw exception when it using as application server for pgadmin4 v6.4. Issue present because on some url script name not present in gequesting path

Description of issue
--------------------------

On runing gunicorn for pgadmin with:
```
gunicorn -b 0.0.0.0:8080 --log-syslog --chdir "$pgadmin_app" "pgadmin_wsgi:application"
```
When tring to get access to site error is rising:

```
         File "/opt/pgadmin/.e/lib/python3.8/site-packages/gunicorn/workers/sync.py", line 136, in handle
           self.handle_request(listener, req, client, addr)
         File "/opt/pgadmin/.e/lib/python3.8/site-packages/gunicorn/workers/sync.py", line 169, in handle_request
           resp, environ = wsgi.create(req, client, addr,
         File "/opt/pgadmin/.e/lib/python3.8/site-packages/gunicorn/http/wsgi.py", line 183, in create
           path_info = path_info.split(script_name, 1)[1]
       IndexError: list index out of range
```
Variables values:
* `path_info`: `'/'`
* `script_name`: `'/pgadmin4'`
   
Purpose of the change
------------------------------

Leave `path_info` as-is if it is not contain `script_name`

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -180,7 +180,11 @@ def create(req, sock, client, server, cfg):
     # set the path and script name
     path_info = req.path
     if script_name:
-        path_info = path_info.split(script_name, 1)[1]
+        try:
+            path_info = path_info.split(script_name, 1)[1]
+        except IndexError:
+            pass     # If script_name not part of path_info leave path_info alone
+        
     environ['PATH_INFO'] = util.unquote_to_wsgi_str(path_info)
     environ['SCRIPT_NAME'] = script_name
 


javabrett:
What is environment `SCRIPT_NAME` being set to here (presumably non-empty string), and why doesn't that appear in the pathinfo?

theawgur:
SCRIPT_NAME variable wasn't set. All actions I made: runing command `gunicorn -b 0.0.0.0:8080 --log-syslog --chdir "$pgadmin_app" "pgadmin_wsgi:application"` in `venv` enviroment.
[This is](https://github.com/benoitc/gunicorn/files/7907647/lounch.sh.zip) lounch script for service.

javabrett:
https://github.com/benoitc/gunicorn/blob/60d0474a6f5604597180f435a6a03b016783885b/gunicorn/http/wsgi.py#L181-L185

Do you have any logs, or know otherwise what `req.path` would be in your error case, and why it wouldn't start with `SCRIPT_NAME`?

theawgur:
In my initial message, I have given variable values. In my case `path_info` was `/` and `script_name` was `/pgadmin4`. I got this values by adding `print()` in patched piece of code before fixing.

URL not starting with `SCRIPT_NAME` because pgadmin serves resources (images, css, etc) by self, and can get as URL `/` only.

If I define `SCRIPT_NAME` as environment variable and set some not `/` to it, exception appear again.

javabrett:
It's not clear to me yet what kind of request this is that is failing, and why it is expected to succeed.

> If I define SCRIPT_NAME as environment variable and set some not / to it, exception appear again.

If you have set:

```
os.environ['SCRIPT_NAME'] = '/pgadmin4'
```

... or equivalent, then requests to e.g. http://server:8080/pgadmin4/login should work, but what requests don't include `/pgadmin4`?  You mention images:

> URL not starting with SCRIPT_NAME because pgadmin serves resources (images, css, etc) by self, and can get as URL / only.

... but the `req.path` in your test was `/` - why should http://server:8080/ work in this case?

When pgadmin4 is configured correctly with `SCRIPT_NAME`, image paths served by it should work fine, e.g. `/pgadmin4/static/img/login.svg?ver=60400`

theawgur:
PgAdmin4 work fine without url prefix. In my case, there is some hostname assigned to pgadmin. Why I need to add some prefix for it?

`gunicorn` can't serve root urls, I understand you correctly?

javabrett:
You are loading the WSGI application from:

https://github.com/postgres/pgadmin4/blob/41bcce09ce08115903567afe106b89eea8f3a352/web/pgAdmin4.wsgi

... instead of e.g. from:

https://github.com/postgres/pgadmin4/blob/41bcce09ce08115903567afe106b89eea8f3a352/web/pgAdmin4.py

The former contains:

https://github.com/postgres/pgadmin4/blob/41bcce09ce08115903567afe106b89eea8f3a352/web/pgAdmin4.wsgi#L16

... which causes your application to load at `/pgadmin4` and not `/`.

You'll need to either change that, or use an alternate application-loading script, or override it via `config_local.py`.

theawgur:
Fix my enviroment.

PR #2727, opened 2022-01-22
always open REDIRECT_TO first.

Geass-LL:
If we close fd 0 first, then open it later. this fd 0 will not be inherited by child process.
This seems to be a bug of python. os.dup2() has a parameter inheritable which defaults to
True, this can make sure our child process have fd 0.

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -489,9 +489,12 @@ def daemonize(enable_stdio_inheritance=False):
 
             if fd_null != 0:
                 os.dup2(fd_null, 0)
+                os.close(fd_null)
+            elif fd_null == 0:
+                os.set_inheritable(fd_null)
 
-            os.dup2(fd_null, 1)
-            os.dup2(fd_null, 2)
+            os.dup2(0, 1)
+            os.dup2(0, 2)
 
         else:
             fd_null = os.open(REDIRECT_TO, os.O_RDWR)
@@ -538,6 +541,7 @@ def daemonize(enable_stdio_inheritance=False):
 
             redirect(sys.stdout, 1)
             redirect(sys.stderr, 2)
+            os.close(fd_null)
 
 
 def seed():


Geass-LL:
Try run this scripts:

```python
import os
import subprocess
import time

os.close(0)
os.open("/dev/null", O_RDWR)
subprocess.run(["sleep", "12345"])
time.sleep(12345)
```
Check the fd sleep child process in /proc/{PID of sleep 12345}/fd, there is no fd 0.

javabrett:
@Geass-LL thanks for raising the issue.

IIUC, the proposed changes:
- Opens the fd to `/dev/null` _before_ closing fds 0,1,2, instead of after.
- Copies the fd returned for `/dev/null`, even when the fd is `0`.

What platform/setup are you running where this makes the difference - or how can we reproduce child processes with no inherited/open fd `0`?

Geass-LL:
> Opens the fd to /dev/null before closing fds 0,1,2, instead of after.
> Copies the fd returned for /dev/null, even when the fd is **0.**

Yes. And the opened fd shouldn't be 0, because it is still used by stdin.

> What platform/setup are you running where this makes the difference - or how can we reproduce child processes with no inherited/open fd 0?

Linux. You can reproduce this by running this script:

```python
import os
import subprocess
import time

os.close(0)
os.open("/dev/null", O_RDWR)
subprocess.run(["sleep", "12345"])
time.sleep(12345)
```

Use `ps aux | grep sleep` to get the PID of "sleep 12345" and run `cat /proc/{PID of sleep}/fd`. There will be no fd 0.

Geass-LL:
You can also reproduce this by flask.

```python
# save as flaskt.py
from flask import flask
import subprocess
import time
app = Flask(__name__)

@app.route('/')
def index():
    ret, out = subprocess.getstatusoutput("sleep 12345")
    print(ret)
    print(out)
    return '<h1>Hello World</h1>'

if __name__ == '__main__':
    app.run()
```
Run this script by `python3 /root/miniconda3/envs/flask/bin/gunicorn --bind 127.0.0.1:8111 --daemon flaskt:app -t 86400`. The "sleep" subprocess will be forked without fd 0 inherited. Adding "-R" option can make fd 0 inherited. But according to the comment in "gunicorn/util.py: daemonize", I think the stdin is designed to have the same behavior.

javabrett:
Thanks @Geass-LL .

Fascinating - this bit of test code highlights the problem with the existing code:

`cat test.py`:

```
import os
fd_null = os.open(os.devnull, os.O_RDWR)
print(str(fd_null))
os.closerange(0, 1)
fd_null = os.open(os.devnull, os.O_RDWR)
print(str(fd_null))
```

`python test.py`:

```
3
0
```

In other words, if you open `/dev/null` before closing `stdin`, it will be assigned some non-zero fd (3 in the above case, since I don't have any fds open other than 0/1/2).  Once `stdin` is closed, `open` will use fd `0` for it.  This is not `stdin` of course, it's `/dev/null`, which we want to assign to `stdin`'s normal fd.  The reuse of fd=0 causes the `dup2` to be bypassed by the `if` statement.

It appears this is Posix behaviour, since per https://linux.die.net/man/3/open :

> The open() function shall return a file descriptor for the named file that is the lowest file descriptor not currently open for that process.

... which is fd 0, since we just closed it.

Geass-LL:
> Once stdin is closed, open will use fd 0 for it. This is not stdin of course, it's /dev/null, which we want to assign to stdin's normal fd. The reuse of fd=0 causes the dup2 to be bypassed by the if statement.

Thanks for the review. Exactly. And this `fd 0` will never be inherited. Even if it is dumped to itself. i.e. `os.dup2(0, 0)`, it will not be inherited. Because python will check the input values, and if the two fds are the same, it will just return. See:

https://github.com/python/cpython/blob/d75a51bea3c2442f81d38ff850b81b8b7f3330f0/Modules/posixmodule.c#L9366

and 

https://github.com/python/cpython/blob/d75a51bea3c2442f81d38ff850b81b8b7f3330f0/Python/dup2.c#L30

javabrett (approved):
This change LGTM and appears required to fix the problem of non-existent and non-inheritable fd 0 for stdin when running daemonized.

I've updated my behaviour-test code above which explains what occurs with the existing code, which in-summary:

- When `fd_null` is opened after `closerange` is called, it is assigned fd 0, which is now free
- This is double-blocked from being copied to fd 0 - the `if` statement blocks it, and also `dup2` has a noop check when the fds are the same
- Because `dup2` is not applied, the fd is not promoted to inheritable

The effect of this change is: `/dev/null` originally gets fd != 0 (e.g. 3), the removal of the `if` statement is not actually required but is harmless, and `dup2` has force.  As a result, `/dev/null` for fd 0 becomes inheritable.

I don't see how tests could be written for this, since the call to `os.fork()` makes this much harder ... would have to be a sysint test.

Could probably use a second review from a maintainer /cc @tilgovi , @benoitc .

javabrett:
See also https://bugs.python.org/issue46454 (closed, not a bug).

javabrett:
For the above Flask app `sleep` subprocess test, results:

`master`:

```
# gunicorn process
Python  42041 user    0u   CHR                3,2      0t0                 310 /dev/null
Python  42041 user    1u   CHR                3,2      0t0                 310 /dev/null
Python  42041 user    2u   CHR                3,2      0t0                 310 /dev/null

# sleep process
sleep   42058 user    1   PIPE 0xbd2b8a914ea7f5b3    16384                     ->0xa45d0dbdc8f4b0a1
sleep   42058 user    2   PIPE 0xbd2b8a914ea7f5b3    16384                     ->0xa45d0dbdc8f4b0a1
```

PR:

```
# gunicorn process
Python  41964 user    0u   CHR                3,2      0t0                 310 /dev/null
Python  41964 user    1u   CHR                3,2      0t0                 310 /dev/null
Python  41964 user    2u   CHR                3,2      0t0                 310 /dev/null
Python  41964 user    3u   CHR                3,2      0t0                 310 /dev/null

# sleep process
sleep   42006 user    0u   CHR                3,2      0t5                 310 /dev/null
sleep   42006 user    1   PIPE 0x1a039fe371eac7d0    16384                     ->0x522d8b430ca23872
sleep   42006 user    2   PIPE 0x1a039fe371eac7d0    16384                     ->0x522d8b430ca23872
```

Note the side-effect of an additional fd open in gunicorn: fd=3 (/dev/null) which was copied to fd 0.  Do we need to close it?

benoitc:
i am not convinced there is an issue in that old code. What is the issue you’re trying to fix?

javabrett:
> What is the issue you’re trying to fix?

- A Flask (standalone) application creates a `subprocess`: the subprocess inherits stdin/stdout/stderr 0/1/2 fds
- The same Flask application running with `gunicorn ... --daemon`: the subprocess is missing stdin fd 0.  Since Gunicorn is daemonized, we expect fd 0 to be present and a reference to `/dev/null` in both the Gunicorn worker and the subprocess.

A consequence on [PEP 446](https://www.python.org/dev/peps/pep-0446/) is that the fd created for `/dev/null` to replace 0/1/2 is created as non-inheritable.  This wouldn't normally be an issue as `dup2()` creates inheritable fds on-copy, and this is working for stdout and stderr 1/2, but for stdin it is skipped by the `if` statement, and even if it wasn't, `dup2()` would be a noop.  It is perhaps unexpected that `os.open()` creates the `/dev/null` fd as fd=0, but it does when called after closing stdin (posix).

The alternative to the change in this PR would be to call Py3.4+ `os.set_inheritable(fd_null)`.

Geass-LL:
> Note the side-effect of an additional fd open in gunicorn: fd=3 (/dev/null) which was copied to fd 0. Do we need to close it?

Yes, and maybe both `fd_null` in the `else` statement. @javabrett

javabrett:
@Geass-LL can you confirm that you have a use-case where your application needs to create a `subprocess`, and it is a problem if `stdin` fd `0` is missing, but it is OK if it is present and `/dev/null`?

javabrett:
> Yes, and maybe also fd_null in the else statement.

Hmmm, good point - the `-R`/`--enable-stdio-inheritance` else looked OK, since it already opens `/dev/null` prior to `os.close(0)` ... and for precisely that reason, it already retains the extra (non-inheitable) fd=3 for `/dev/null`:

```
Python  44853 user    0u   CHR                3,2      0t0                 310 /dev/null
Python  44853 user    1w   REG                1,4        0            79038001 out.txt
Python  44853 user    2w   REG                1,4      289            79038002 err.txt
Python  44853 user    3u   CHR                3,2      0t0                 310 /dev/null
```

..., meaning as-is without/with redirect would be consistent, _with_ the extra fd.

But it probably makes sense to remove that, I don't think it should be there.

I'm leaning slightly towards the open-after-close (in both branches), getting fd `0`, then `os.set_inheritable(fd_null)` solution now, which just undoes the unintended result of PEP 446.

Geass-LL:
> @Geass-LL can you confirm that you have a use-case where your application needs to create a `subprocess`, and it is a problem if `stdin` fd `0` is missing, but it is OK if it is present and `/dev/null`?

Sure.  We meet this problem when we use flask to test rpm (with ndb as backend database).

```python
subprocess.getstatusoutput("rpm -qa")
```

During the query, rpm will first open its database in `/var/lib/rpm/Packages.db`, and it will get a `fd 0`.  Then rpm will check if  this fd is valid. Since `fd 0` is reserved for stdin, rpm treats 0 as invalid and fails.

I have also sent a PR to rpm. See: <https://github.com/rpm-software-management/rpm/pull/1891>

I believe there must be other scenes that assume 0 as stdin, and want `fd 3 or larger` returned from parent process.

tilgovi:
> I'm leaning slightly towards the open-after-close (in both branches), getting fd 0, then os.set_inheritable(fd_null) solution now, which just undoes the unintended result of PEP 446.

I agree with this completely. I would find that code the most explicit.

Geass-LL:
New version force pushed as @javabrett suggests.

Since I'm not familiar with gunicorn, I don't know how to deal with `fd < 0`. May be @tilgovi has a better solution.

javabrett:
@Geass-LL apologies for some duplicate effort here - I'd been drafting an alternative for discussion, see #2731 - can you review and test that, both with and without -R?  Looking for consistent and correct fds 0/1/2 but not 3.

javabrett (changes requested):
This will be replaced by #2731 when reviewed/agreed.

javabrett:
Replaced by #2731 - thanks @Geass-LL for identifying this and for the original patch.

PR #2731, opened 2022-01-27
Ensure fd 0 stdin </dev/null is always inheritable

javabrett:
Alternative to #2727 , see discussion there.

When `gunicorn --daemon` daemonizes the process, prior to this change it was
noted that in the general case (without `-R` / `--enable-stdio-inheritance`), when fd 0
was replaced with /dev/null, the dup2 copy is skipped, and per PEP 446
"Make newly created file descriptors non-inheritable", the result was a stdio
fd </dev/null which was non-inheritable.  As a result, any launched subprocess
did not have an open 0/stdin fd, which can cause problems in some applications.

This change retains the behaviour of opening /dev/null with fd 0, but adds a call
to os.set_inheritable(..) to ensure the fd is inheritable.

The -R branch had different logic but has now been standardised with the general
case.  It was previously opening /dev/null as fd 3 and the dup2() copy made it
inheritable as fd 0.  This branch now applies the same logic: open as fd 0
(i.e. after close(0)), then set_inheritable.  As a result, an extra fd 3 </dev/null
previously left open is no longer left open.

Test Flask application good for manual testing on any platform with `dd` or any other program that will read stdin:

`flaskt.py`

```
import os
import sys

# save as flaskt.py
from flask import Flask
import subprocess
import time
app = Flask(__name__)

@app.route('/')
def index():
    ret, out = subprocess.getstatusoutput("dd of=/dev/null")
    return '<h1>fd 0 is Inheritable: {}, ret: {}, out: {}</h1>'.format(str(os.get_inheritable(0)), ret, out)

if __name__ == '__main__':
    app.run()
```

Results from http://localhost:8080 ... `gunicorn -b 0.0.0.0:8000 --daemon --error-logfile err.txt flaskt:app`

master:
> fd 0 is Inheritable: False, ret: 1, out: dd: stdin: Bad file descriptor

PR:
> fd 0 is Inheritable: True, ret: 0, out: 0+0 records in 0+0 records out 0 bytes transferred in 0.000017 secs (0 bytes/sec)

### DIFF
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -486,7 +486,10 @@ def daemonize(enable_stdio_inheritance=False):
             closerange(0, 3)
 
             fd_null = os.open(REDIRECT_TO, os.O_RDWR)
+            # PEP 446, make fd for /dev/null inheritable
+            os.set_inheritable(fd_null, True)
 
+            # expect fd_null to be always 0 here, but in-case not ...
             if fd_null != 0:
                 os.dup2(fd_null, 0)
 
@@ -494,13 +497,17 @@ def daemonize(enable_stdio_inheritance=False):
             os.dup2(fd_null, 2)
 
         else:
-            fd_null = os.open(REDIRECT_TO, os.O_RDWR)
-
             # Always redirect stdin to /dev/null as we would
             # never expect to need to read interactive input.
 
+            os.close(0)
+
+            fd_null = os.open(REDIRECT_TO, os.O_RDWR)
+            # PEP 446, make fd for /dev/null inheritable
+            os.set_inheritable(fd_null, True)
+
+            # expect fd_null to be always 0 here, but in-case not ...
             if fd_null != 0:
-                os.close(0)
                 os.dup2(fd_null, 0)
 
             # If stdout and stderr are still connected to
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -497,17 +497,13 @@ def daemonize(enable_stdio_inheritance=False):
             os.dup2(fd_null, 2)
 
         else:
+            fd_null = os.open(REDIRECT_TO, os.O_RDWR)
+
             # Always redirect stdin to /dev/null as we would
             # never expect to need to read interactive input.
 
-            os.close(0)
-
-            fd_null = os.open(REDIRECT_TO, os.O_RDWR)
-            # PEP 446, make fd for /dev/null inheritable
-            os.set_inheritable(fd_null, True)
-
-            # expect fd_null to be always 0 here, but in-case not ...
             if fd_null != 0:
+                os.close(0)
                 os.dup2(fd_null, 0)
 
             # If stdout and stderr are still connected to
--- a/gunicorn/util.py
+++ b/gunicorn/util.py
@@ -472,6 +472,19 @@ def daemonize(enable_stdio_inheritance=False):
 
         os.umask(0o22)
 
+        # Always redirect stdin to /dev/null as we would
+        # never expect to need to read interactive input.
+
+        os.close(0)
+
+        fd_null = os.open(REDIRECT_TO, os.O_RDWR)
+        # PEP 446, make fd for /dev/null inheritable
+        os.set_inheritable(fd_null, True)
+
+        # expect fd_null to be always 0 here, but in-case not ...
+        if fd_null != 0:
+            os.dup2(fd_null, 0)
+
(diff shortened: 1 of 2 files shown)


Geass-LL:
Works for me. And the PR LGTM. Thanks.
@javabrett

gunicorn/util.py
```diff
@@ -486,21 +486,28 @@ def daemonize(enable_stdio_inheritance=False):
...
             os.dup2(fd_null, 1)
             os.dup2(fd_null, 2)
 
         else:
-            fd_null = os.open(REDIRECT_TO, os.O_RDWR)
-
             # Always redirect stdin to /dev/null as we would
             # never expect to need to read interactive input.
 
+            os.close(0)
```
benoitc, line 503: why closing it there?

tilgovi: This is making a best effort to create this new dev null as fd 0, and not creating an fd 3 that needs to be duplicated to 0. I think this is just trying to be more direct.

tilgovi: It's also nice symmetry. In one case, we close 0-3, then make a new 0-3. In the other, we close only 0, and make a new 0.

If we wanted to, we could make the code a little shorter by closing and recreating fd 0 unconditionally. Then the if branches are just about what to do with fd 1 and 2.

tilgovi: I think that might a actually be helpful for the reader to see that we never keep stdin open. Even though the option is to enable stdio inheritance, it really only allows stdout and stderr inheritance.

javabrett: Right, this is mostly for consistency with the fd-s created in both `not enable_stdio_inheritance` and `else`.  Since the aim is always to `</dev/null`, it's most direct to explicitly `close(0)` first in both cases, then `/dev/null` will (presumably always) be created as `fd=0`, which will skip the `dup2()` via the `if`, and the `dup2()` is a noop in that case anyway, which was the original cause of the non-inheritable fd raised in the issue.

Currently when `enable_stdio_inheritance`, `/dev/null` is created as `fd=3`, which is non-inheritable and not closed, so the worker ends-up with an extra `fd=3=/dev/null`, but any `subprocess` will only have `0/1/2`.

Since in both cases we always want to at least `close(0)` then create an inheritable replacement, it makes sense to me to do all those things explicitly, and the `dup2` isn't required, and is retained only for some weird situation if `close(0)` then `os.open()` created a non-zero fd, which shouldn't happen.

I can look at another commit to refactor a little more and make the stdin redirect common.

The other option of course is to not touch the `enable_stdio_inheritance` branch, since it currently fortuitously doen't have the non-inheritable stdin fd - that would mean ignoring the slight current difference, where with `enable_stdio_inheritance` the worker has an additional open, non-inheritable `fd=3=/edv/null`, which is mostly inconsequential.

javabrett:
I pushed another commit that pulls-up the common stdin `</dev/null` code and makes it clear that stdin is always redirected from `/dev/null`; the common code uses the `close(0)` first and therefore open as fd 0 approach, so we won't get fd 3 and need to copy.

fd 0 is made inheritable explicitly.  If for some weird reason we get a non-zero fd after `close(0)`, that case is still caught so still OK, but we don't bother to close any non-zero fd should that happen.

gunicorn/util.py
benoitc: well it's a common practice to no close until you really need it. I would keep it that way and focus the commit on what really need to be fixed.

benoitc: here is also related old commit : https://github.com/benoitc/gunicorn/commit/29aefcc1cfc32097481aea6213a0df09962c7817

I would fix what latest Python has broken and have a separate patch/discussion about the rest.

javabrett: Thanks @benoitc for the review.

I've pushed a commit that reverts the refactoring change, and another that reverts the part of the original change to make the `-R` branch operate the same as the main branch i.e. `os.close(0)` then `os.open(...)` -> 0.  Running with `-R` will retain an additional `3 -> /dev/null`, non-inheritable, which is being relied-on (3 != 0) in order to trigger make-inheritable as part of `dup2(...).  As you say, these could be considered in a future PR if desired.

Retained is `os.set_inheritable(fd_null, True)` in the non `-R` branch, and some clarifying comments only.  This directly undoes the effect of PEP 446 in Python 3.4, which occurs due to the direct replacement of fd 0, bypassing `dup2(...)`.

Issue #2453, opened 2020-11-09
Gunicorn 20.0.4 Print Server Log on STDERR


Nitrosito:
Hi
I used Gunicorn 20.0.4 conterized and running on GKE on Google, the problem is:

Server Logs Print into /dev/stderr , and Google Stackdriver (or google monitoring) interpret as ERROR LOG (logic)

My parameters to run Gunicorn:
```
exec gunicorn -w 3 -b 0.0.0.0:3000 --enable-stdio-inheritance --log-level debug --timeout 1000 --threads 3  --access-logfile - --error-logfile - handler:app
```
access-logfile with "-" for print access.log into stdout
error-logifle with "-" for print error.log into stderr

Why server logs print on STDERR instead of STDOUT?
for example, if i set gunicorn log level to info.. all logs print into stderr.. same for other log levels on gunicorn

Nitrosito:
I'm very sorry, my fault, for putting "capture-output" ... I did not understand how it works

redirect stdout and stderr to "stderr" log

I close the incident, thank you

Nitrosito:
I see, by default, only 2 handler.. one for error and one for other log on gunicorn:

```
CONFIG_DEFAULTS = dict(
        version=1,
        disable_existing_loggers=False,

        loggers={
            "root": {"level": "INFO", "handlers": ["console"]},
            "gunicorn.error": {
                "level": "INFO",
                "handlers": ["error_console"],
                "propagate": True,
                "qualname": "gunicorn.error"
            },

            "gunicorn.access": {
                "level": "INFO",
                "handlers": ["console"],
                "propagate": True,
                "qualname": "gunicorn.access"
            }
        },
        handlers={
            "console": {
                "class": "logging.StreamHandler",
                "formatter": "generic",
                "stream": "sys.stdout"
            },
            "error_console": {
                "class": "logging.StreamHandler",
                "formatter": "generic",
                "stream": "sys.stderr"
            },
        },
        formatters={
            "generic": {
                "format": "%(asctime)s [%(process)d] [%(levelname)s] %(message)s",
                "datefmt": "[%Y-%m-%d %H:%M:%S %z]",
                "class": "logging.Formatter"
            }
        }
)
```

Source: https://github.com/mjjbell/gunicorn/blob/master/gunicorn/glogging.py

I create a logging.ini 
```
[loggers]
keys=root, gunicorn.error, gunicorn.access

[handlers]
keys=console, error_console

[formatters]
keys=generic

[...]

PR #2752, opened 2022-02-24
INFO and DEBUG logs will be logged through `gunicorn.access`

mindflayer:
This is a brief explanation of the problem (quoting from https://github.com/benoitc/gunicorn/issues/2453#issuecomment-1049235886):
"""
If I am not mistaken, the problem seems quite clear to me: apart from access logs, everything else is logged through `self.error_log` (`gunicorn.error`) which outputs to `stderr`.
"""

Fix for #2453.

Let me know if any additional refactor is needed. For instance, I would rename `self.access_log` as `self.info_log`, keeping the old one as alias for retro-compatibility.


benoitc:
why would you expect errors logged to stdout instead of stderr?

mindflayer:
That's exactly what is about. The problem is exactly when gunicorn logs to stderr INFO logs.

philn-delphia:
> why would you expect error or debug logged to stdout instead of stderr? I think only info logs should be eventually logged to stdout. Also you can pretty easily redirect to stdout from the command line. Can you give more context?

Monitoring tools (like stackdriver for GKE) interpret stderr as errors, while stdout should be used for any other output. This is (for me) primarily an issue with containerized applications, but I imagine other tools might also care that stderr is used appropriately.

racinmat:
If it's problem to have it as a default, this feature could be opt-in, but it would be great to have this possibility, it would make the usage in GCP much easier.

buraczyn:
Any update on this? Seeing the same thing in my application and would love to be able to have it print correctly

mananpreetsingh:
any update on this?

PR #2753, opened 2022-02-25
Clarify forwarded_allow_ips setting, fix spelling

trendels:
I find the documentation of the `forwarded_allow_ips` setting a bit hard to understand, so I propose this small change (together with some spelling fixes): https://github.com/benoitc/gunicorn/commit/8af94b52f9e1fb15dd55cbf4ecbac476499b8491

I also rebuilt the documentation, as suggested by the contributing guide. It turns out a bunch of other settings had their docs updated in the meantime and were also included. I hope that's OK, if not I can open another PR with just the first commit.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -310,8 +310,6 @@ file format.
 ``logconfig_dict``
 ~~~~~~~~~~~~~~~~~~
 
-**Command line:** ``--log-config-dict``
-
 **Default:** ``{}``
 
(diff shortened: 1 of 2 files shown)


benoitc:
Closing as the documentation has been migrated from RST to MkDocs format and the substantive improvements from this PR have been incorporated during that migration. The detailed explanations about the interplay between forwarded_allow_ips and secure_scheme_headers are now in the current docs. Thank you for your contribution!

PR #2756, opened 2022-03-11
Efficient response processing

gongul:
If the returned response contains a lot of data, it makes many unnecessary calls to resp.write.


benoitc:
Thanks for the patch but this handling of the response body is wanted. Response is streamed and should not be returned at once. This allows the application to not get all the response at once in memory. I will close it for now, but feel free to comment it if needed.

gongul:
@benoitc Does the application mean a program that executes gunicorn? What do you think about processing by determining the chunk size?

PR #2758, opened 2022-03-13
Prevent unnecessary setuid call

dwarfmaster:
Prevent `setuid` call if the uid is already the right one. In my use case, that allows me starting `gunicorn` with `systemd` straight with the right user, without the permission to do `setuid` syscalls, for some more security.


benoitc:
thanks!

PR #2780, opened 2022-04-13
Drop obsolete code from `SSLVersion` setting

### LINKED ISSUE #2549: SSL version setting code is slightly messy
`default` and `desc` attributes of `SSLVersion` are redeclared.
It looks like b01fe34e5633768052e70d88e7e8db84368f93e3 was supposed to update the setting, but old code was not deleted.
https://github.com/benoitc/gunicorn/blob/b83448bfb42184f43adf155f86bcd2a570a1139d/gunicorn/config.py#L2036-L2080

Thanks,
Illia

illia-v:
b01fe34e5633768052e70d88e7e8db84368f93e3 updated the setting, but the old code was left in place.

Fixes #2549.

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -2052,32 +2052,6 @@ class SSLVersion(Setting):
        ``ssl.PROTOCOL_TLS`` when Python >= 3.6 .
 
     """
-    default = ssl.PROTOCOL_SSLv23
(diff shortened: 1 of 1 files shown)


benoitc:
no longer needed. Thanks for the PR anyway :)

PR #2808, opened 2022-05-28
Disabled reap_workers() on SIGCHLD without workers

### LINKED ISSUE #2807: Raising "Worker failed to boot." on on_exit hook
### Environment
```console
$ uname -srom
Linux 5.15.41-1-MANJARO x86_64 GNU/Linux
```
```console
$ python --version
Python 3.10.4
```
```console
$ gunicorn --version
gunicorn (version 20.1.0)
```

### What is happenning
I added `on_exit` hook:
```python3
import subprocess

def on_exit(server):
    for _ in range(10):
        subprocess.run(["/bin/sh", "./justexit.sh", "3"], capture_output=True)
```
`justexit.sh` is really simple:
```sh
#!/usr/bin/env sh
if [ -z "$1" ]; then
    exit 0
fi
exit $1
```

And on shutdown got this error:
```
gunicorn.errors.HaltServer: <HaltServer 'Worker failed to boot.' 3>
```
Full example can be found [here](https://github.com/TitaniumHocker/gunicorn-subprocess-bug).

### Why it happens
`on_exit` hook is running in the same process, so if it creates child process, like in my example, server will receive `SIGCHLD` signal on child process exit. Server [will call](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L242) `reap_workers` method while handling this signal. In this method server [gets exit code](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L513) of child process and if it is equal to **3** server [will interpret](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L523) this as `Worker failed to boot` error.

TitaniumHocker:
Closes #2807

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -239,7 +239,8 @@ class Arbiter(object):
 
     def handle_chld(self, sig, frame):
         "SIGCHLD handling"
-        self.reap_workers()
+        if self.WORKERS:
+            self.reap_workers()
         self.wakeup()
 
     def handle_hup(self):


benoitc:
thanks for the patch. I understand the optimisations. Though we need to keep the reap worker call for this too lines: 

https://github.com/benoitc/gunicorn/pull/2808/files#diff-2cb7ac1e139f4df283747ee8c3cd0f26bfaa8aaf4c5891097f8c0e382d1a618aL516-L517

which catch the case we restart ourself. Maybe it can be moved there https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L518 though. Feel free to send me a new PR with thiis change.

Issue #1562, opened 2017-08-02
`--reload` option not working


ahmedsaoudi:
Upon using Gunicorn for my Django app, I tried to enable the `--reload` option to make my work easier, avoiding manual restarting for every change.

Even though I tried the option on the command line (`gunicron --reload --bind ...`), I normally run Gunicorn as a systemd service. Here the config file: 

```
[Unit]
Description='gunicorn daemon'
After=network.target

[Service]
User=me
Group=www-data
WorkingDirectory=path/to/djangoapp
ExecStart=/path/to/Envs/venv/bin/gunicorn --access-logfile - --error-logfile /path/to/gunicorn-error.log --workers 3 --reload --bind unix:/path/to/djangoapp/appname.sock appname.wsgi:application
ExecReload=/bin/kill -s HUP $MAINPID
ExecStop=/bin/kill -s TERM $MAINPID

[Install]
WantedBy=multi-user.target
```
As you can see, I'm using Gunicorn from a Virtualenv, as the rest of the app; but I don't know if that's related.

And apparently, I'm not alone; somebody else has the same issue: https://unix.stackexchange.com/questions/375292/how-to-restart-systemd-service-after-files-update-under-gunicorn-server

benoitc:
is there any log that can be useful?

Code0x58:
Does it just never work? The way the reloading works might not be quite as simple as you expect, see [the code](https://github.com/benoitc/gunicorn/blob/19.7.1/gunicorn/reloader.py):
* if [inotify](https://pypi.python.org/pypi/inotify) isn't installed (or you use `--reload-engine=poll`) then gunicorn will only look for changes in your loaded python modules (including those added after each worker starts), and files added using `--reload-extra-file`
  * if you don't use a .py file extension and your file name ends with py/pyo/pyc, then it won't be watched

* otherwise the directories of python modules which were loaded when the watcher started, and of the files provided with `--reload-extra-file` will be watched for changes using inotify
   * files added to directories which were created after the watcher was created won't be watched
   * subdirectories of those watched are not automatically watched
   * if your editor uses memory mapping to modify the file, then it the changes won't be detected

  Modules that only imported in functions or inside something like `if __name__ == '__main__':`, unless they are loaded as a result of your gunicorn configuration file, or WSGI entry point.
  You could use [`inotifywait`](https://linux.die.net/man/1/inotifywait) to see if the issue is to do with patchy watching, or inotify limitations. 

[...]

PR #2820, opened 2022-06-29
Fix auto reloader

### LINKED ISSUE #2339: reload does not work for uvicorn.workers.UvicornWorker
this is my commandline`
gunicorn -k uvicorn.workers.UvicornWorker --reload-engine=inotify   --workers=1 --log-level debug  --access-logfile - --error-logfile - --disable-redirect-access-to-syslog --reload  main:app`

note that i have tested with both master branch of gunicorn as well as last stable release.

This is the code im using
```

from fastapi import FastAPI

app = FastAPI()


@app.get("/")
async def read_root():
    # raise 1
    return {"Hello": "World"}


@app.get("/items/{item_id}")
async def read_item(item_id: int, q: str = None):

    return {"item_id 2": item_id, "q": q}

```

reload does not pickup when i make changes to the app. im using vscode as an editor

### LINKED ISSUE #2814: Not reloading Python modules when inside k3s+docker
Hi folks,

I'm not sure if this is a bug or if I'm doing something wrong, but I'm building a project where I need to run my app inside Kubernetes pods on my own computer, so I use k3s for that. However, I've set it up to reload my gunicorn instance running inside my pod, using `poll` as the reload engine (gave up using inotify, wasn't working reliably inside the pod), and, even though gunicorn says it detected file changes and is reloading the workers, they don't really seem to be reloaded, as I keep getting the same contents over and over.

To test this inside my app, which is basically a FastAPI-based API, I created an endpoint that provides me a certain JSON payload. Every time I change this payload, I see gunicorn saying it's reloading the modules, but if I hit the endpoint again I get the old payload back.

Maybe Gunicorn is caching something that I don't know? Do you guys see anything wrong in my approach?

Thanks!

bigsbug:
Auto-reloading with the Uvicron worker after detecting new changes produces the following issues: 
Fixes #2339
Fixes #2814

Also, I guess there are some issues with the auto-reload, such as:
#1562

There is a problem with using the sys.exit() in a forked process (child processes or workers) instead of using the os._exit() that is mentioned in [python documentation](https://docs.python.org/3/library/os.html#os._exit).

This modification solved the problem and passed all tests.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -15,6 +15,7 @@ Alex Gaynor <alex.gaynor@gmail.com>
 Alex Robbins <alexander.j.robbins@gmail.com>
 Alexandre Zani <alexandre.zani@gmail.com>
 Alexis Le-Quoc <alq@datadoghq.com>
+Ali Sadeghi Far <booxgitboox@gmail.com>
 Anand Chitipothu <anandology@gmail.com>
 Andreas Stührk <andy-python@hammerhartes.de>
 Andrew Burdo <zeezooz@gmail.com>
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -581,7 +581,7 @@ class Arbiter(object):
             self.log.info("Booting worker with pid: %s", worker.pid)
             self.cfg.post_fork(self, worker)
             worker.init_process()
-            sys.exit(0)
+            os._exit(0)
         except SystemExit:
             raise
         except AppImportError as e:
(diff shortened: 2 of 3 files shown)


bigsbug:
Hi, @brosner do you can review this PR and merge it?

benoitc:
Changes looks good. I will check. I would rather prefer that python fix this. Has an issue been open upstream?

bigsbug:
yes there has an open issue for it

bigsbug:
Hi, @brosner, I checked this problem with different versions of python and the result are the same !!. and I also checked the documentation of several versions of python and all of them mentioned this situation.
I don't think it is an issue with python, and in the python upstream I can't find any related issue.
so what do you think?

tilgovi:
I don't think I understand this PR. From what I read, it's frequently the case that a program may not want to close flush stdio or call exit handler from a child process, but does Gunicorn? I don't think we register any `atexit` handlers, but developers might do so in server hooks. Would they expect those to be called when a worker exits? I don't know. This would definitely be a change in behavior.

Do we have an explanation for why it solves the linked issues? I think that's what I'm missing. I'm not understanding the value, only the risk.

benoitc:

PR #2827, opened 2022-07-13
add attribute remote_addr to req object

zffocussss:
Hi benotic: 
    as I come from uvicorn, and we are running gunicorn + django behind nginx,so I think it is more convenient for people to get the remote address with request.remote_addr than the raw REMOTE_ADDR.I hope you consider my suggestions and approve it.
   PS: I use the pre_request to add some request headers based on the remote address.Uvicorn also have a request object,I can get the client ip with request.client.host.

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -31,6 +31,7 @@ class Message(object):
         self.cfg = cfg
         self.unreader = unreader
         self.peer_addr = peer_addr
+        self.remote_addr = peer_addr
         self.version = None
         self.headers = []
         self.trailers = []
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -31,7 +31,7 @@ class Message(object):
         self.cfg = cfg
         self.unreader = unreader
         self.peer_addr = peer_addr
-        self.remote_addr = peer_addr
+        self.remote_addr = peer_addr[0]
         self.version = None
         self.headers = []
         self.trailers = []


benoitc:
DO you have any example of usage for it?

benoitc:
@zffocussss ?

zffocussss:
> @zffocussss ?

sorry to not reply to you in time.as I forgot it probably.In my opinion,it is just a shortcut which is useful for people which want get the real client ip directly.

jayaddison:
Should e3336cfa9fe6dba623276155f48208c6d118d937 have been merged instead of c9d01c801967596f66636cfc0c47660339d32631 here?

zffocussss:
> Should [e3336cf](https://github.com/benoitc/gunicorn/commit/e3336cfa9fe6dba623276155f48208c6d118d937) have been merged instead of [c9d01c8](https://github.com/benoitc/gunicorn/commit/c9d01c801967596f66636cfc0c47660339d32631) here?

commit https://github.com/benoitc/gunicorn/commit/c9d01c801967596f66636cfc0c47660339d32631 is enough,as I do not touch any codes but this tiny snippet

jayaddison:
Ok, thanks @zffocussss.  c9d01c801967596f66636cfc0c47660339d32631 seems incomplete though -- it does not include unit tests.  Could you check?

jayaddison:
@benoitc maybe worth reverting this (or merging #2896) this until the feature can be completed?

TomiBelan:
Hi. Unfortunately, this PR probably broke the ability to bind to unix sockets. See #2904.

TomiBelan:
Bug #2904 is fixed in master. The broken commit c9d01c801967596f66636cfc0c47660339d32631 (one half of this PR) was reverted in 30baeebf337850a7ce3cb8099ffa248cef5831ef (PR #2927) by @benoitc and @samypr100.

I wonder if they saw my comment, or they found the problem independently...

For what it's worth, 69bf1d0cf44abffeeb469d4c36e1ff978b6e5769 (the second half of this PR) is still in master -- on its own, useless, but harmless.

PR #2834, opened 2022-07-27
refactor: prefer contextlib.suppress to ImportError: pass

kkirsche:
Use [contextlib](https://docs.python.org/3/library/contextlib.html)'s [suppress method](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) to silence a specific error, instead of passing in an exception handler.

The context manager slightly shortens the code and significantly clarifies the author's intention to ignore the specific errors. The standard library feature was introduced following a [discussion](https://bugs.python.org/issue15806), where the consensus was that

> A key benefit here is in the priming effect for readers... The with statement form makes it clear before you start reading the code that certain exceptions won't propagate.

This feature was added in Python 3.4, as Gunicorn supports 3.5+, this is safe for use.

### DIFF
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -4,6 +4,7 @@
 # See the NOTICE for more information.
 # pylint: disable=no-else-continue
 
+import contextlib
 import os
 import os.path
 import re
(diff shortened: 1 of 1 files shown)


benoitc:
Thanks for the patch!  However i disagree with the statement :)  The with force me to know what does that contextlib where looking at the exception and see where I ignore it forces me to understand the why.  It ease refactoring or review IMHO to keep that way.

PR #2836, opened 2022-07-27
refactor: remove python 2 new-style classes as only python 3.5+ is supported now

kkirsche:
In Python 2, there were different ways to create classes: the old and new style. `gunicorn` uses the new-style of class creation, which inherits from `object`. This difference is described here: 

https://docs.python.org/2/reference/datamodel.html#new-style-and-classic-classes
https://www.python.org/download/releases/2.2.3/descrintro/

As `gunicorn` supports Python 3.5+ at this time, the use of new-style classes can be removed as all classes in Python 3 are new-style classes by default.

### DIFF
--- a/examples/frameworks/cherryapp.py
+++ b/examples/frameworks/cherryapp.py
@@ -1,7 +1,7 @@
 import cherrypy
 
 
-class Root(object):
+class Root:
     @cherrypy.expose
     def index(self):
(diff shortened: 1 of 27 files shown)


benoitc:
Closing as the equivalent cleanup was already merged in #3189. Thank you for your contribution!

PR #2853, opened 2022-08-22
Added type hinting for pre-request and post-request documentation

miohtama:
This will increase the discoverability of the code and makes starting write handlers easier.


benoitc:
no type hitting is sipported in gunicorn. We prefer code clarity and readability. Thanks for the patch!

PR #2870, opened 2022-09-29
Miscellaneous lint fixups

jayaddison:
Addresses some warnings, errors and recommendations reported by `pylint` and `pycodestyle`.

In particular there are three categories of `pylint` reports that are _not_ addressed within this pull request:

- The recommendation to provide an explicit `encoding` argument to the `open` function (rationale: this could introduce behavioural changes)
- The notification that the `wrap_socket` method is deprecated (this may be addressed by #2649)
- The suggestion to convert string formatting to use Python 3.6+ f-strings (this requires a base Python requirement change, and also affects many touchpoints throughout the codebase)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,29 +1,24 @@
 name: lint
 on: [push, pull_request]
+permissions:
+  contents: read # to fetch code (actions/checkout)
 jobs:
-  tox-lint:
+  lint:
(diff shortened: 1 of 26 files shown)


jayaddison:
It looks like continuous integration checks are passing again in the mainline branch, so closing this (feel free to recover the commits and continue from them).

PR #2871, opened 2022-09-29
Use f-strings for string variable interpolation across the codebase

jayaddison:
- Bumps the minimum Python version requirement to v3.7
- Updates string/variable interpolation to use f-strings as recommended by `pylint`


gunicorn/glogging.py
```diff
@@ -313,18 +311,18 @@ def atoms(self, resp, req, environ, request_time):
         if hasattr(req_headers, "items"):
             req_headers = req_headers.items()
 
-        atoms.update({"{%s}i" % k.lower(): v for k, v in req_headers})
+        atoms.update({f"{{{k.lower()}}}i": v for k, v in req_headers})
```
jayaddison, line 314: This (and two similar touchpoints within the next 20 lines) is one of the places where I wonder whether f-strings are an improvement or not.  The escaping logic is not completely intuitive.

gunicorn/app/base.py
```diff
@@ -35,7 +35,7 @@ def do_load_config(self):
             self.load_default_config()
             self.load_config()
         except Exception as e:
-            print("\nError: %s" % str(e), file=sys.stderr)
+            print(f"\nError: {e}\n", file=sys.stderr)
```
jayaddison, line 38: Note: additional newline added here; I'll undo that since it's a behaviour change.

README.rst
```diff
@@ -32,7 +32,7 @@ The documentation is hosted at https://docs.gunicorn.org.
 Installation
 ------------
 
-Gunicorn requires **Python 3.x >= 3.5**.
+Gunicorn requires **Python 3.x >= 3.7**.
```
jayaddison, line 35: Note: Python 3.7 isn't strictly required for formatted string literals; they were [introduced in Python 3.6](https://docs.python.org/3.6/whatsnew/3.6.html).

That confused me for a moment: why the version requirement bump to 3.7 here?

It's been a while since I opened this, but I think the rationale was that [Python 3.6 is no-longer-developed](https://peps.python.org/pep-0494/#schedule-last-security-only-release), that it wouldn't make sense to increase a _minimum_ version number to support that.

But I think I've changed my mind.  Some systems could still have valid reasons to use Py3.6, and as long as this change is compatible with those, then that seems like a better lower-bound.  I'll push another commit for that soon.

(they should probably upgrade, but that's up to them, really)

gunicorn/config.py
```diff
@@ -57,13 +57,13 @@ def __str__(self):
         for k in sorted(self.settings):
             v = self.settings[k].value
             if callable(v):
-                v = "<{}()>".format(v.__qualname__)
-            lines.append("{k:{kmax}} = {v}".format(k=k, v=v, kmax=kmax))
+                v = f"<{v.__qualname__}()>"
+            lines.append(f"{k:{kmax}} = {v}")
```
jayaddison, line 61: Missing escaping here, I think (so probably some additional test coverage required)

gunicorn/instrument/statsd.py
```diff
@@ -99,21 +99,21 @@ def access(self, resp, req, environ, request_time):
             status = int(status.split(None, 1)[0])
         self.histogram("gunicorn.request.duration", duration_in_ms)
         self.increment("gunicorn.requests", 1)
-        self.increment("gunicorn.request.status.%d" % status, 1)
+        self.increment(f"gunicorn.request.status.{status}", 1)
```
jayaddison, line 102: Even though the `status` variable _should_ have an integer type here, let's maintain the existing integer presentation type for it during string formatting.

gunicorn/config.py
jayaddison: It turns out that this was not an escaping issue after all.

- Legacy [Python string formatting](https://docs.python.org/3.6/library/string.html?highlight=str#format-examples) using the string `%` operator _does_ retain curly braces
  - For example, `"{%s}" % "hi"` evaluates to `"{hi}"`
- `str.format` string formatting uses curly-braces to delimit positional and/or named arguments
  - For example, `"{}".format("hi")` evaluates to `"hi"`

The code referenced by this comment was mistaken for legacy-string formatting (which would require escaping of curly braces) but is in fact `str.format` based (so no escaping of curly braces is required).

Even so: the use of `kmax` as a fill operator seemed not-entirely-intuitive (although very neat), so some test coverage for that behaviour has been added in 719a4b1b5023161187bc73c695cf26d7cfaaa469 (with demonstration that it does catch errors in 9f0f9a2917e0aadb01cd92308ce456be58f67bed -- subsequently reverted by 7bc470853c7338ade6c9f84eeed5c4d80c96d56c).

jayaddison:
Not completely convinced that this is valuable; closing.  Feel free to recover the pull request and continue it.

PR #2872, opened 2022-09-29
Remove Travis CI configuration

jayaddison:
As far as I can tell, the Travis CI configuration in this repository is no longer required or maintained.  This changeset removes it; this was discovered while searching around the codebase for Python version references as part of #2871.


benoitc:
thanks!

PR #2879, opened 2022-10-15
Fix HTTP status code for Request Header Fields Too Large

### LINKED ISSUE #2878: Wrong HTTP status code on Request Header Fields Too Large
When the request header gets too large (e.g. a lot of cookies), gunicorn returns HTTP 400 instead of HTTP 431, which can mask the real problem through proxies. Gunicorn should return the right HTTP error code.

Teko012:
This change makes sure that a proper HTTP status code is returned for Request Header Fields Too Large, which fixes #2878.


benoitc:
thanks!

PR #2885, opened 2022-10-26
align license with OSI template

lsmith77:
This will ensure that GitHub will properly detect the license.

```
licensee detect .
License:        MIT
Matched files:  LICENSE
LICENSE:
  Content hash:  4c2c763d64bbc7ef2e58b0ec6d06d90cee9755c9
  Attribution:   Copyright 2009-2018 (c) Benoît Chesneau <benoitc@e-engura.org>, 2009-2015 (c) Paul J. Davis <paul.joseph.davis@gmail.com>
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       MIT
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,12 +62,6 @@ Contributing
 See `our complete contributor's guide <CONTRIBUTING.md>`_ for more details.
 
 
-License
--------
-
-Gunicorn is released under the MIT License. See the LICENSE_ file for more
-details.
-
 .. _Unicorn: https://bogomips.org/unicorn/
 .. _`#gunicorn`: https://web.libera.chat/?channels=#gunicorn
 .. _`Libera.chat`: https://libera.chat/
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -62,12 +62,6 @@ Contributing
 See `our complete contributor's guide <CONTRIBUTING.md>`_ for more details.
 
 
-License
```
apollo13, line 65: I think it wouldn't hurt to keep this.

benoitc:
from where is coming this license detection? Seems like a patch should be sent to github possibly instead of removing readability

README.rst
benoitc: I agree. Let just change the LICENSE file possibly. Can you provide this change?

lsmith77:
> from where is coming this license detection? Seems like a patch should be sent to github possibly instead of removing readability. Is this possible?

you can run https://github.com/licensee/licensee locally

the tool gets confused if there are multiple references to licenses.

README.rst
benoitc: bump?

benoitc:
i am not sure I like this change. This makes it barely unreadable. What I f I have multiple copyrights date. Imo this githb code should be fixed instead of asking to projects to make their list of copyrights unreadable.

tilgovi:
Stalled. Closing.

PR #2887, opened 2022-10-27
Fix sendfile.py example

python273:
because of `@validator` decorator, the fn returns `wsgiref.validate.IteratorWrapper` instead of `gunicorn.http.wsgi.FileWrapper` which breaks sendfile

https://github.com/benoitc/gunicorn/blob/69c508ac6e4b301045d3ce21acce8b416415d4c5/gunicorn/workers/sync.py#L181-L182

### DIFF
--- a/examples/sendfile.py
+++ b/examples/sendfile.py
@@ -9,7 +9,7 @@ import os
 from wsgiref.validate import validator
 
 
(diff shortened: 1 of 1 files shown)


benoitc:
thanks!

PR #2891, opened 2022-11-04
Update docs to acknowledge Gthread worker_connections

pb-dod:
The Gthread worker does use `worker_connections`: https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gthread.py#L198

However, the docs currently say it doesn't.

This PR updates the docs to reflect the current code.


pb-dod:
Oops, there's already an open PR for this: https://github.com/benoitc/gunicorn/pull/2691

PR #2896, opened 2022-11-23
Fix `remote_addr` error

bcail:
Tests are currently failing with
`TypeError: 'NoneType' object is not subscriptable`

This commit fixes that by setting `remote_addr` equal to `peer_addr`.


apollo13:
I'd go even further an simply revert the original commit.

bcail:
Sure, I can revert the original commit if that would be merged. Just let me know.

benoitc:
similar change already applied

PR #2912, opened 2023-01-02
Added support for proxy protocol v2.

apollo13:


gunicorn/http/message.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, cfg, unreader, peer_addr):
         self.cfg = cfg
         self.unreader = unreader
         self.peer_addr = peer_addr
-        self.remote_addr = peer_addr[0]
+        # self.remote_addr = peer_addr[0]
```
apollo13, line 34: Without this the tests don't work at all, see https://github.com/benoitc/gunicorn/pull/2896 -- should be remove before merge.

gunicorn/http/message.py
```diff
@@ -2,19 +2,19 @@
...
 
 from gunicorn.http.body import ChunkedReader, LengthReader, EOFReader, Body
 from gunicorn.http.errors import (
     InvalidHeader, InvalidHeaderName, NoMoreData,
     InvalidRequestLine, InvalidRequestMethod, InvalidHTTPVersion,
     LimitRequestLine, LimitRequestHeaders,
 )
-from gunicorn.http.errors import InvalidProxyLine, ForbiddenProxyRequest
-from gunicorn.http.errors import InvalidSchemeHeaders
+from gunicorn.http.errors import InvalidProxyHeader, InvalidProxyLine
```
benoitc, line 16: why this change?

apollo13:
Introduced by accident, will fix up later today.

gunicorn/http/message.py
apollo13: Hi Benoit, looking over this twice I am not sure which change you mean or what the problem with it is? Do you mean why I introduced `InvalidProxyHeader` or why I changed the order?

gunicorn/http/message.py
```diff
@@ -180,97 +180,103 @@ def __init__(self, cfg, unreader, peer_addr, req_number=1):
...
+    def proxy_protocol(self, unreader, buffer):
         """\
         Detect, check and parse proxy protocol.
 
         :raises: ForbiddenProxyRequest, InvalidProxyLine.
         :return: True for proxy protocol line else False
         """
         if not self.cfg.proxy_protocol:
-            return False
+            return
```
sirkonst, line 259: A docstring above still says "... else False". 

And why did you decide change False to None?

gunicorn/http/message.py
```diff
@@ -324,6 +330,47 @@ def parse_proxy_protocol(self, line):
...
+        if ver_cmd & 0xF0 != 0x20:
+            raise InvalidProxyHeader("invalid version %r" % data)
+        if ver_cmd & 0xF not in {0x00, 0x01}:
+            raise InvalidProxyHeader("unsupported command %r" % data)
+
+        body = self.read_bytes(unreader, buffer, length)
+        if ver_cmd & 0xF == 0x00:
+            return  # LOCAL command, do not change source addr
+
+        if fam == 0x11:  # TCPv4
```
sirkonst, line 345: How about to make constants with meaningful names for `0x11` and so on instead comments? Or [enum.IntFlag](https://docs.python.org/3/library/enum.html#enum.IntFlag)?

gunicorn/http/message.py
```diff
@@ -324,6 +330,47 @@ def parse_proxy_protocol(self, line):
...
+            fmt = "!16s16sHH"
+            ip_class = ipaddress.IPv6Address
+        else:
+            raise InvalidProxyHeader("unsupported protocol %r" % body)
+
+        try:
+            s_addr, d_addr, s_port, d_port = struct.unpack(fmt, body)
+        except struct.error as e:
+            raise InvalidProxyHeader("cannot unpack %r: %s" % (body, e))
+        s_addr = str(ip_class(s_addr))
```
sirkonst, line 360: Why is it needed? - converting `str -> ipaddress -> str`. Why just not use `str` representation as is?

If it is needed for validation then where is the error handling?

gunicorn/http/message.py
benoitc: @apollo13 ?

gunicorn/http/message.py
benoitc: @apollo13 ?

gunicorn/http/message.py
benoitc: i was thinking to the import removed, but I need to re-read the code. I guess they were not used anymore?

apollo13:
Hi @sirkonst and @benoitc -- sorry for the late reply; I have lost track here. I'll update and address the comments today. In the meantime it would be great if you could review #2913, due to the bugs in `Unreader` (at least if I understood the purpose of it correctly) I was not able to utilize it in this PR properly.

gunicorn/http/message.py
apollo13: It is no longer needed. Previously the code checked the return value based on this and then changed behavior. Now the proxy protocol parsing is fully "transparent" and leaves the streams in a usable state for follow up. Will adjust the comment in a follow up commit.

gunicorn/http/message.py
apollo13: Sounds like a plan, will see what I can come up with.

gunicorn/http/message.py
apollo13: `s_addr` is never really a string but either an 4-byte integer (ipv4 case) or a 16-byte integer (ipv6 case). Those are "narrow" binary representations of the addresses and not directly usable. So this code is turning the address into something human readable.

As for validation errors: How much do we want to trust the trusted proxies to send the correct data? :D

apollo13:
I think I have addressed all comments, as far as I can see the linting errors are coming from other files. Let me know what you think.

benoitc:
sorry , didn't intent to close this . I will take care of this change in coming release

Issue #2913, opened 2023-01-03
Potential bug in `Unreader.unread`


apollo13:
Hi there,

while working on #2912 I think I found a potential bug in the `Unreader` implementation. See the following testdiff that currently fails:

```diff
diff --git a/tests/test_http.py b/tests/test_http.py
index 33481266..32dfdba6 100644
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -123,9 +123,17 @@ def test_unreader_read_when_size_is_none():
 
 
 def test_unreader_unread():
-    unreader = Unreader()
+    unreader = IterUnreader(())
+    unreader.unread(b'hi there')
+    assert b'hi there' == unreader.read()
+
     unreader.unread(b'hi there')
-    assert b'hi there' in unreader.read()
+    hi = unreader.read(2)
+    assert b'hi' == hi
+    unreader.unread(hi)
+    result = unreader.read()
+    print(result)
+    assert b'hi there' == result
 
 
 def test_unreader_read_zero_size():
```

The `result` after the final `unread` is `b' therehi'` instead of the expected (imo?) `b'hi there'`. This is because unreads append to the buffer which is only the correct thing to do if the buffer is empty, which assumes that the whole buffer has been read. As demonstrated this does not have to be the case. You can probably argue that this bug can't occur because `Unreader` will only be used correctly but imo this is a footgun waiting to happen, what do you think?

If we agree that this is a bug I can work on a fix.

benoitc:
Closing due to lack of activity. If you can confirm this is still an issue and have a proposed fix, please submit a PR.

PR #2915, opened 2023-01-06
bugfix // Unread should place data at the beginning of the buffer

mapogolions:
Please see added unit test that reproduces the issue.


mapogolions:
Related issue https://github.com/benoitc/gunicorn/issues/2913

benoitc:
Applied this fix in commit 0032c797. All tests pass. Thank you for the contribution!

PR #2916, opened 2023-01-13
Rectification: prefer the term 'main' for the co-ordinating process name

jayaddison:
Seemed worthwhile after noticing some redundancy in a comment during review of #2871 (addressed by 37b5599c7d0eaa21fd867a2ca2e29bf70a473f51).

Also includes a tangential edit to remove the word `evil` from the codebase (please discontinue use of `gunicorn` if you previously relied on this).

### DIFF
--- a/docs/source/design.rst
+++ b/docs/source/design.rst
@@ -11,18 +11,18 @@ Server Model
 ============
 
 Gunicorn is based on the pre-fork worker model. This means that there is a
-central master process that manages a set of worker processes. The master
-never knows anything about individual clients. All requests and responses are
-handled completely by worker processes.
+central main process that manages a set of worker processes. The main never
+knows anything about individual clients. All requests and responses are handled
+completely by worker processes.
 
 Master
 ------
 
-The master process is a simple loop that listens for various process signals
-and reacts accordingly. It manages the list of running workers by listening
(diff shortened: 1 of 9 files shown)


jayaddison:
That's OK: it does have (in fact multiple) meanings within programming and other industries too (the production of musical records, for example).

The reason for suggestions like this is that there may be alternative names that are equally-clear and that may not have as many negative associations for other people.

Not making the change is understandable because it could cause breakage -- process list or log filtering problems, for example -- potentially affecting lots of application sites, given the popularity of `gunicorn`.

In practice, my guess is that such mechanisms tend to be used by more sophisticated application sites that should have the financial resources to allocate staff and time to detection and resolution of such problems.

My hope would be that the community of users and developers around `gunicorn` would be understanding and supportive even if that happened -- or ideally would discover and spot potential problems before they could cause errors.

jayaddison:
> In practice, my guess is that such mechanisms tend to be used by more sophisticated application sites that should have the financial resources to allocate staff and time to detection and resolution of such problems.

(it's also possible that this is completely wrong, and that many underfunded institutions might rely on code behaviours without having the associated staffing to update and repair problems.  that would probably imply a need to investigate why those measures had to be introduced, and, if possible, improving operability of the code -- in the open -- to gradually filter out the requirement for those workarounds in underresourced environments without reducing (or indeed preferably even improving) their quality of service)

Issue #2917, opened 2023-01-16
Gunicorn gthread deadlock


JorisOnGithub:
### Problem
When using gunicorn with the gthread worker, and using multiple WebView2 browsers, our gunicorn instance ends up in a deadlock.
The WebView2 browser (and maybe other browsers as well, but I have not tested those) can make speculative TCP connections: TCP connections that the browser thinks it will need as it expects to be making multiple HTTP requests to the server in quick succession. Those speculative TCP connections can be created without any data being sent on them.

The Gunicorn gthread worker currently makes the assumption that every tcp connection opened by a client, will contain data. When a new connecion is accepted, the gunicorn worker thread does a blocking recv call on the new socket, this blocks the thread. In our case this recv call would stay blocking, as no data was written to this particular speculative TCP connection.

In general all worker threads can get blocked at the same time when many users connect/send requests to a gunicorn server at the same time with a browser that uses speculative TCP connections (or browsers that create TCP connections without sending data for other reasons). This causes any new requests to be stuck in the queue of the gthread worker and not be processed.

#### Real world
They way we initially reproduced the issue was by refreshing N WebView2 browser instances pointing to a gunicorn gthread instance at the same time, where N > the number of threads configured for gunicorn. The more windows, the more likely for gunicorn to enter a deadlock. In our testing the blocked worker threads can stay blocked for hours, as in our case the 'empty' TCP connections were not closed from the browser side. Reproducing the deadlock is not deterministic, refreshing more windows at the same time makes it more likely to hit the deadlock. In our testing, we were able to consistently reproduce it with 10 threads and roughly 12 or more windows.
When refreshing some of the windows (which closes the speculative TCP connections created by them), the threads that were blocked on those connections unblocked (because the socket they were blocked on closed) and continue serving the application without issues.

#### Minimal example
An easier way to reproduce this for maintainers is to simply run the example test app with `gunicorn --worker-class gthread --threads 2 test:app`. Then open the example app in edge at localhost:8000. Then try to open localhost:8000 in an incognito edge tab, or another browser (I tried firefox). The app won't load in those other tabs. This might depend on your system/edge settings (I reproduced it with all defaults).

Doing `ss -tn | grep 8000` on the machine running gunicorn shows the empty open TCP connections.
[...]

PR #2918, opened 2023-01-16
gthread: only read sockets when they are readable (#2917)

JorisOnGithub:
This PR solves issue #2917

It changes the behaviour of the gthread workers. Instead of always reading from tcp sockets opened by a client, the worker only does so when the socket is readable. The main thread does not immediately enqueue new connections anymore to the worker thread queue. Instead it adds new connections to the poller, these connections will be added to the worker thread queue when they are readable.

Please let me know if there are any questions or suggestions for this PR!

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -40,12 +40,15 @@ class TConn(object):
 
         self.timeout = None
         self.parser = None
+        self.initialized = False
 
         # set the socket to non blocking
         self.sock.setblocking(False)
 
     def init(self):
+        self.initialized = True
         self.sock.setblocking(True)
+
         if self.parser is None:
             # wrap the socket if needed
             if self.cfg.is_ssl:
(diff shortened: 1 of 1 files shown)


TBoshoven:
Verified as follows that this resolves the issue:
* Started gunicorn application with 4 gthread workers
* Open four TCP connections to the listening port
* Attempt to make a request

Before this change there was no answer. After this change there is.

tilgovi (commented):
This looks great. Thanks for helping to resolve this issue!

I have only one question. Should we add the connection to the keepalive queue with an initial timeout? In other words, should we treat the connection as initially idle, and close it after the keepalive timeout?

TBoshoven:
> should we treat the connection as initially idle, and close it after the keepalive timeout?

I think that's a reasonable way to approach this.

benoitc:
> This looks great. Thanks for helping to resolve this issue!
> 
> I have only one question. Should we add the connection to the keepalive queue with an initial timeout? In other words, should we treat the connection as initially idle, and close it after the keepalive timeout?

hrm yes I think that could be a nice addition as well.

benoitc:
merged it , i will add the timeout in a separate commit. Thanks a lot for this patch! Thanks @tilgovi for the review as well.

benoitc:
@JorisOnGithub @tilgovi https://github.com/benoitc/gunicorn/commit/e7781d2c2c669a87532ec6dd5d04490085ef7559

PR #2937, opened 2023-02-01
#1564 separate logs and metrics

vladimir-avinkin:
Proposed fix for #1564 

Separate the metrics from logs, fixing multiple issues including issues in other projects that depend on gunicorn (https://github.com/encode/uvicorn/issues/610)

Because of the split, some metrics were dropped because implementing them is no longer feasible not that they are separate (gunicorn.log*) and they weren't that useful in my experience.

Implement an abstract metrics system with two concrete implementations and the ability to pass custom ones.

Two implementations are previously used gunicorn's statsd implementation and also dogstatsd that uses a 3rd party library that is now included in gunicorn's extras


joshhoegen:
Hey @vladimir-avinkin... I guess this is just a sanity check...

 is this related to statsd metrics not being sent when using Gunicorn + Uvicorn? I am completely stumped, and I can't tell if [this thread](https://github.com/encode/uvicorn/issues/610#issuecomment-611987371) is specifically the official Gunicorn integration or sending StatsD histograms, increment, etc. 

I have DD working locally using only uvicorn to run, but once I run ginicorn with unicorn workers, metrics fail to report.

vladimir-avinkin:
@joshhoegen Yes, this is related, current gunicorn metrics implementation is tied to login subsystem, which uvicorn overrides thus disabling the metrics. The thread you've linked has explained it well.

danielebra:
hey @vladimir-avinkin, great work on laying the foundation for this separation! Do you know what is the current status of this is?

Jean-Daniel:
At this point, I think you can consider this project in maintenance mode, where only minor bug fix are applied, and switch to an other fork if you want new feature or important bug fix (like this one or #2938 which are show-stoppers but are pending review or progress from more than 1 year).

If you want to get this feature, you should create your own fork and apply the patches you need.

benoitc:
> At this point, I think you can consider this project in maintenance mode, where only minor bug fix are applied, and switch to an other fork if you want new feature or important bug fix (like this one or #2938 which are show-stoppers but are pending review or progress from more than 1 year).
> 
> If you want to get this feature, you should create your own fork and apply the patches you need.

this is untrue. 24.0.0 has been released. Sometimes you know free contributors have some other duties.

benoitc:
Closing due to inactivity. This is a large change that needs significant review and testing. If there's still interest, please consider opening a fresh PR against current master. Thank you for the effort\!

PR #2958, opened 2023-03-22
2747-Replace pkg_resources.parse_version with packaging.version.parse

taleinat:
This is a step towards removing the dependency on pkg_resources, which is part of setuptools, and thus makes setuptools a runtime dependency.

Note the addition of the [packaging](https://packaging.pypa.io/) library as a dependency.

See: #2747


sirkonst (approved):

gotmax23 (approved):

gotmax23:
I merged this into https://github.com/benoitc/gunicorn/pull/2963

PR #2960, opened 2023-03-25
Add prometheus

### LINKED ISSUE #2738: Support prometheus for access metrics
Gunicorn only supports statsd for metrics. please add prometheus too.

I think it can be done using a new logger for prometheus, and add `prometheus_listen_addr` and `prometheus_multiproc_dir` to settings with https://github.com/prometheus/client_python .

ATofighi:
This is a try to fix #2738 and add prometheus metrics for gunicorn

My Ideas/Checklist:
 - [x] Allow to use more than one logger and run all of them for logging. It can allow us to use both statsd, prometheus and custom logger.
 - [ ] Add prometheus to requirements - Should it be added as `extras_require`? for example, use `pip install gunicorn[prometheus]` to have prometheus metrics
 - [x] Write a chainLogger class to do above line.
 - [x] Write a custom logger for prometheus instrument
 - [x] Use prometheus in multiprocess mode
 - [ ] Somehow run prometheus metrics server as a special worker
   - I think it can run with worker_class in arbiter, but I don't know how to do it clean and maintainable.
 - [ ] Write unit tests
 - [ ] prometheus multiprocess path argument on collector doesn't work!! find another way :(

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -11,6 +11,8 @@ import sys
 import time
 import traceback
 
+from prometheus_client import multiprocess
+
 from gunicorn.errors import HaltServer, AppImportError
 from gunicorn.glogging import BaseLogger
 from gunicorn.pidfile import Pidfile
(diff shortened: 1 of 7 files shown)


ATofighi:
This implementation may also be a solution for #1564 or has conflicts with #2937

ATofighi:
I findout that prometheus_client force using environ variable for multiprocdir, I close this PR and think about it.

PR #2986, opened 2023-05-08
add missing double quote

chandanch:
Opening double quote was missed during the recent hence, showing the below error:

auth.split(b":", 1)[0].decode(UTF-8", "replace")
SyntaxError: unterminated string literal (detected at line 472)


benoitc:
thanks!

PR #2992, opened 2023-05-11
Deprecate ssl_version option

tsaarni:
This change defaults `SSLContext` to Python's `ssl.create_default_context()` and marks `ssl_version` option as deprecated. The option value will be ignored, and warning will be printed in stderr. The reason for deprecation is that `ssl_version` option was depending on old method of setting TLS min/max version. Rationale for ignoring the given value is that the method has not worked well anymore with modern Python versions. See discussion thread https://github.com/benoitc/gunicorn/pull/2649#discussion_r1190801170.

As an alternative, config file can be used to set the min/max values, for example:

```python
def ssl_context(conf, default_ssl_context_factory):
    import ssl
    context = default_ssl_context_factory()
    context.minimum_version = ssl.TLSVersion.TLSv1_3
    context.maximum_version = ssl.TLSVersion.TLSv1_3
    return context
```

Additionally, this PR fixes some documentation discrepancies by regenerating `settings.rst` from `config.py`. The PR also suggests small modification to `docs/gunicorn_ext.py` that makes it possible to get sane defaults generated to `settings.rst` for `user`, `group` and `chdir` options, which otherwise will reset to whatever was effective for the user who regenerated the documentation.

### DIFF
--- a/docs/gunicorn_ext.py
+++ b/docs/gunicorn_ext.py
@@ -48,7 +48,9 @@ def format_settings(app):
 
 
 def fmt_setting(s):
-    if callable(s.default):
+    if hasattr(s, "default_doc"):
+        val = s.default_doc
+    elif callable(s.default):
         val = inspect.getsource(s.default)
         val = "\n".join("    %s" % line for line in val.splitlines())
         val = "\n\n.. code-block:: python\n\n" + val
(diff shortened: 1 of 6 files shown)


benoitc:
thanks for the documentation fixes and new default. Merging it as is :)

Issue #3038, opened 2023-07-21
Connection reset during max-requests auto-restart with gthread


christianbundy:
We're observing intermittent HTTP 502s in production, which seem to be correlated with the "autorestarting worker after current request" log line, and are less frequent as we increase `max_requests`. I've reproduced this on 21.2.0 and 20.1.0, but I haven't [yet] performed a bisect to determine when it started.

I've produced a minimal reproduction case diverging from the example on gunicorn.org as little as possible, but please let me know if there are other changes you'd recommend:

## Application

```python3
def app(environ, start_response):
    data = b"Hello, World!\n"
    start_response("200 OK", [
        ("Content-Type", "text/plain"),
        ("Content-Length", str(len(data)))
    ])
    return iter([data])
```

## Configuration

```sh
gunicorn --max-requests 10 --log-level debug --worker-class gthread myapp:app
```

## Reproduction

```bash
#!/usr/bin/env bash

set -euo pipefail

# start gunicorn
gunicorn --max-requests 10 --log-level debug --worker-class gthread myapp:app &

# wait for gunicorn to be ready
sleep 5

# send 20 sequential HTTP requests with minimal output
for i in {1..20}; do
    curl -S -s -o /dev/null -w "REPRO[$i]: %{http_code}\n" "127.0.0.1:8000/foo/${i}" || true
done
```

## Quickstart

For convenience, I've packaged this into a single command that consistently reproduces the problem on my machine. If you have Docker installed, this should Just Work™️. 

```sh
cat <<END_OF_DOCKERFILE | docker build -t gunicorn-repro-max-requests-rst - && docker run --rm gunicorn-repro-max-requests-rst
FROM python:3.11

RUN pip install --disable-pip-version-check --no-cache gunicorn

RUN cat <<EOF > myapp.py
def app(environ, start_response):
    data = b"Hello, World!\n"
    start_response("200 OK", [
        ("Content-Type", "text/plain"),
        ("Content-Length", str(len(data)))
    ])
    return iter([data])
EOF

RUN cat <<EOF > repro.sh
#!/usr/bin/env bash

set -euo pipefail

# start gunicorn
gunicorn --max-requests 10 --log-level debug --worker-class gthread myapp:app &

# wait for gunicorn to be ready
sleep 5

# send 20 sequential HTTP requests with minimal output
for i in {1..20}; do
[...]

PR #3039, opened 2023-07-21
Prevent gthread connection reset on max-requests restart

christianbundy:
**Problem**

The `while self.alive` in `ThreadWorker.run()` doesn't protect against `ThreadWorker.accept()` being called, which can cause a connection reset when the gthread worker type is combined with the `max_requests` setting.

**Solution**

Add a guard clause to protect against this. It's very possible that there's a better way to solve this, but this is my first stab at the problem.

---

Fixes: #3038

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -118,6 +118,9 @@ class ThreadWorker(base.Worker):
         self._wrap_future(fs, conn)
 
     def accept(self, server, listener):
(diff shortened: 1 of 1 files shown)


mevinbabuc:
What are the chances of this getting merged ?

benoitc:
Closing this draft PR due to age and draft status. If this issue is still relevant, please feel free to open a fresh PR against current master. Thank you!

Issue #3056, opened 2023-08-17
Signal exit log has pid hardcoded in message


jeroenp:
The signal logging in arbiter.py has the pid and signal in the message format string. That makes it hard to group these messages together. 

Also, the preceding `if exitcode > 0` clause duplicates the log statement for the exit error that was already in the function a few lines above it. I'm assuming that the first one is valuable to keep because there are two "raises" in between.

PR #3057, opened 2023-08-17
De-duplicate and parameterize exit error logging.

jeroenp:
The signal log message had the PID and signal in the message format string, which makes each message unique. This is awkward in log processing with for example Sentry, where you want to see similar messages grouped together.

The exit message was emitted twice, from the "if exitcode != 0" and the "if exitcode > 0" clause. You would be seeing things like:

[2023-08-17 09:50:10 +0200] [66450] [ERROR] Worker (pid:66451) exited with code 1 [2023-08-17 09:50:10 +0200] [66450] [ERROR] Worker (pid:66451) exited with code 1.

See issue #3056

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -524,7 +524,8 @@ class Arbiter(object):
                     # infinite start/stop cycles.
                     exitcode = status >> 8
                     if exitcode != 0:
-                        self.log.error('Worker (pid:%s) exited with code %s', wpid, exitcode)
+                        self.log.error("Worker (pid:%s) exited with code %s",
+                                       wpid, exitcode)
                     if exitcode == self.WORKER_BOOT_ERROR:
                         reason = "Worker failed to boot."
                         raise HaltServer(reason, self.WORKER_BOOT_ERROR)
@@ -532,12 +533,7 @@ class Arbiter(object):
                         reason = "App failed to load."
                         raise HaltServer(reason, self.APP_LOAD_ERROR)
 
-                    if exitcode > 0:
-                        # If the exit code of the worker is greater than 0,
-                        # let the user know.
-                        self.log.error("Worker (pid:%s) exited with code %s.",
-                                       wpid, exitcode)
-                    elif status > 0:
+                    if exitcode == 0 and status > 0:
                         # If the exit code of the worker is 0 and the status
                         # is greater than 0, then it was most likely killed
                         # via a signal.
@@ -545,13 +541,12 @@ class Arbiter(object):
                             sig_name = signal.Signals(status).name
                         except ValueError:
                             sig_name = "code {}".format(status)
-                        msg = "Worker (pid:{}) was sent {}!".format(
-                            wpid, sig_name)
+                        msg = "Worker (pid:%s) was sent %s!"
 
                         # Additional hint for SIGKILL
                         if status == signal.SIGKILL:
                             msg += " Perhaps out of memory?"
-                        self.log.error(msg)
+                        self.log.error(msg, wpid, sig_name)
 
                     worker = self.WORKERS.pop(wpid, None)
                     if not worker:


benoitc:
I like this change. Are we abble to create a test for it?

jeroenp:
> I like this change. Are we abble to create a test for it?

Something like a logging class in dummy application in test_arbiter.py that captures log messages? I was assuming that arbiter.stop() hits these changes, but instead only reap_workers() hits this.

tilgovi (approved):
I like the change and think we should merge it with or without tests. It's pretty straightforward. That said, a simple test would be to just call `reap_workers` with `os.waitpid` patched on an arbiter constructed with a patched `log`.

But it seems excessive to me.

jeroenp:
There's also PR 3094. https://github.com/benoitc/gunicorn/pull/3094, which is related and prevents "operator confusion" so to say.

gunicorn/arbiter.py
```diff
@@ -522,37 +522,40 @@ def reap_workers(self):
...
+                    signum = os.WTERMSIG(status) \
+                        if os.WIFSIGNALED(status) else 0
+                    try:
+                        sig_name = signal.Signals(signum).name
+                    except ValueError:
+                        sig_name = str(signum)
+
                     if exitcode != 0:
-                        self.log.error('Worker (pid:%s) exited with code %s', wpid, exitcode)
+                        self.log.error("Child exited with code %s (pid: %s)",
```
benoitc, line 537: we should use "Worker"  instead of "Child" here. This is more meaningful in the context of Gunicorn arbitration.

gunicorn/arbiter.py
```diff
@@ -522,37 +522,40 @@ def reap_workers(self):
...
+                                       exitcode, wpid)
+                    elif signum in (0, signal.SIGTERM):
+                        # 0: as seen from macos/py311 from reload on workers
+                        # where workers from SIGTERM but waitpid/nohang gives
+                        # me 0.
+                        self.log.info("Child terminated (pid: %s)", wpid)
```
benoitc, line 543: same commmenta s above.

gunicorn/arbiter.py
```diff
@@ -522,37 +522,40 @@ def reap_workers(self):
                     # A worker was terminated. If the termination reason was
                     # that it could not boot, we'll shut it down to avoid
                     # infinite start/stop cycles.
-                    exitcode = status >> 8
+
+                    exitcode = os.WEXITSTATUS(status) \
```
benoitc, line 526: why not reusing the status ?

benoitc (changes requested):
please replace the word Child by Worker. Can you tell me why you don't want to use the `exitcode= status >> 8` ?

gunicorn/arbiter.py
```diff
@@ -522,37 +522,40 @@ def reap_workers(self):
...
-                        # via a signal.
-                        try:
-                            sig_name = signal.Signals(status).name
-                        except ValueError:
-                            sig_name = "code {}".format(status)
-                        msg = "Worker (pid:{}) was sent {}!".format(
-                            wpid, sig_name)
-
-                        # Additional hint for SIGKILL
-                        if status == signal.SIGKILL:
```
benoitc, line 552: we miss this extra hint  in the change. Could you add it?

gunicorn/arbiter.py
jeroenp: Hi Benoit, 

Sorry, I totally missed that I committed to this open PR! That was not my intention. I'll have a look what happened before and what I want from these changes. My intent was to lower the log level on the regular reload/sigterms. And the two raises in between the log statements were irksome to me. 

What I find confusing that both master and worker are logging the same event, with the same wording. In the logs it's not clear which process is which.

gunicorn/arbiter.py
jeroenp: I was trying to find out why I was/am seeing 0 on worker reload, where I was expecting to see 15 for SIGTERM. I was hoping that this might give me more, but clearly 0 stays 0.

gunicorn/arbiter.py
jeroenp: You miss the hint? I thought it was misleading because it never is a memory issue whenever we get to see these error logs.

PR #3060, opened 2023-09-01
Validate chunk sizes.

kenballus:
Because of quirks of `int()` chunk sizes like `'+0x0'` are currently erroneously accepted by the HTTP parser. This patch checks that chunk sizes are valid (match the grammar in the RFC) before using them.


kenballus:
Forgot that I incorporated this change into #3059. Closing.

Issue #2816, opened 2022-06-20
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'> #2667 


InfernalAzazel:
error:

2022-06-20 08:06:06 +0800] [7] [WARNING] Worker with pid 2855 was terminated due to signal 6
[2022-06-20 08:06:06 +0800] [2864] [INFO] Booting worker with pid: 2864
[2022-06-20 08:06:06 +0800] [7] [WARNING] Worker with pid 2856 was terminated due to signal 6
[2022-06-20 08:06:06 +0800] [7] [WARNING] Worker with pid 2861 was terminated due to signal 6
--- Logging error ---
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1103, in emit
    stream.write(msg + self.terminator)
RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1103, in emit
    stream.write(msg + self.terminator)
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 242, in handle_chld
    self.reap_workers()
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 530, in reap_workers
    self.log.warning(
  File "/usr/local/lib/python3.10/site-packages/gunicorn/glogging.py", line 261, in warning
    self.error_log.warning(msg, *args, **kwargs)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1489, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1624, in _log
    self.handle(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1634, in handle
    self.callHandlers(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1696, in callHandlers
    hdlr.handle(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 968, in handle
    self.emit(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1108, in emit
    self.handleError(record)
  File "/usr/local/lib/python3.10/logging/__init__.py", line 1021, in handleError
    sys.stderr.write('--- Logging error ---\n')
[...]

Issue #3050, opened 2023-08-07
Error instead of warning in logs when sending SIGTERM since v21


feliixx:
Hi @benoitc,

Thanks a lot for all the work you've done on Gunicorn!

We recently upgraded from version `20.1.0` to `21.2.0` and noticed in our logs when restarting our application that a `warning` has been replaced by an `error` when sending `SIGTERM` to a worker:

version 20.1.0:

```
systemd: Stopping App...
app: gunicorn WARNING Worker with pid 2542594 was terminated due to signal 15
app: gunicorn INFO Handling signal: term
app: gunicorn INFO Worker exiting (pid: 2542537)
app: gunicorn INFO Worker exiting (pid: 2542545)
app: gunicorn INFO Shutting down: Master
systemd: Stopped App.
systemd: Started App.
```

version 21.2.0:

```
systemd: Stopping App...
app: gunicorn ERROR Worker (pid:1340527) was sent SIGTERM!
app: gunicorn INFO Handling signal: term
app: gunicorn INFO Worker exiting (pid: 1340420)
app: gunicorn INFO Worker exiting (pid: 1340424)
app: gunicorn INFO Shutting down: Master
systemd: Stopped App.
systemd: Started App.
```

This is a bit annoying for us as we carefully monitor any error sent by our app, and this seems to be a false positive.

The commits introducing this change are 76f8da2 and 28df992, however it's not clear from the commit messages why the log level was changed. [The documentation about signals](https://docs.gunicorn.org/en/stable/signals.html) does say that sending `SIGTERM` is the proper way to do a graceful shutdown.

Is it possible to change the log level back to `warning`, at least for `SIGTERM`? We'll be happy to send a PR if you agree.

Venture200:
Hello, any update on this ?

feliixx:
For other people affected by this: we've managed to sidestep the issue by using `KillMode=mixed` in our systemd service config: 

```
[Service]
KillMode=mixed
...
```

We've been using this config in production for over a year with Gunicorn 21.2.0 and above without issues.

pauljeannot:
For my understanding, using `KillMode=mixed` means that while the master process receives SIGTERM for a graceful shutdown, the workers are immediately sent a SIGKILL, which prevents them from completing any pending tasks or writing logs (so no false positive in your monitoring). 

Am I right?

feliixx:
[...]

PR #3064, opened 2023-09-05
arbiter: don't log if handling SIGCHLD

hydrargyrum:
Logging when handling a signal is a bad practice
See https://docs.python.org/3/library/logging.html#thread-safety

Fixes https://github.com/benoitc/gunicorn/issues/2816

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -239,7 +239,7 @@ class Arbiter(object):
 
     def handle_chld(self, sig, frame):
         "SIGCHLD handling"
-        self.reap_workers()
+        self.reap_workers(signal_safe=False)
         self.wakeup()
 
     def handle_hup(self):
@@ -507,7 +507,7 @@ class Arbiter(object):
             else:
                 self.kill_worker(pid, signal.SIGKILL)
 
-    def reap_workers(self):
+    def reap_workers(self, signal_safe=True):
         """\
         Reap workers to avoid zombie processes
         """
@@ -532,12 +532,12 @@ class Arbiter(object):
                         reason = "App failed to load."
                         raise HaltServer(reason, self.APP_LOAD_ERROR)
 
-                    if exitcode > 0:
+                    if signal_safe and exitcode > 0:
                         # If the exit code of the worker is greater than 0,
                         # let the user know.
                         self.log.error("Worker (pid:%s) exited with code %s.",
                                        wpid, exitcode)
-                    elif status > 0:
+                    elif signal_safe and status > 0:
                         # If the exit code of the worker is 0 and the status
                         # is greater than 0, then it was most likely killed
                         # via a signal.


hydrargyrum:
@benoitc what is done on purpose? raising exceptions and breaking logging system

benoitc:
This code should be threadsafe. I suspect the error is happening before. We need first to understand why this code is re-entrant here.  

Anyuway I didn' aim to close, but only comment.  Let keep it open since it is actually solving the cause.

benoitc:
linking to https://github.com/benoitc/gunicorn/issues/3050 I think we should revisit this change that introducde more logging.  

@hydrargyrum if you happen to be around i'm on irc.

ja0x:
Wanted to shine a little bit of info I found since I have encountered this bug as well.  The issue isn't with thread-safety but rather with re-entrancy within the logging module and the underlying `BufferedWriter` (from IO module).

The issue occurs (usually on a slow disk) when the SIGCHLD handler interrupts the main process while it was in the process of writing its logging buffer.  The handler (which is the same thread) will try to write to that buffer but will be unable because it doesn't allow reentrant code, it will then raise an exception.  

If logger is set to write `stderr` to that same stream or file, it then sets off a cascading reentrant code error, eventually resulting in a recursion error.

It is a hard bug to reproduce, because the main thread has to interrupt itself exactly when it was writing the buffer.  

Ref:
https://docs.python.org/3/library/io.html#reentrancy,
https://github.com/python/cpython/blob/e24f9ae7035cfb4c11c4a5c2820121fc22463f8b/Modules/_io/bufferedio.c#L29

TomiBelan:
@ja0x is correct. Python programs MUST NOT print or log anything inside a signal handler. https://stackoverflow.com/q/45680378 is another link that explains the root cause of this problem.

This gunicorn bug has a long and sad history.

*Dec 2017:* Issue #1663 was reported: In some cases (for example OOM) workers die without any error logged, or even crashloop.
*Dec 2019:* Issue #2215 was reported: Essentially the same as #1663, but with a reproducible example, and focused on missing logging.
*Apr 2020:* PR #2315 (ee1e5c1928c71fa366dfac99c65f630e03aed8b7) was created by @tilgovi. It logs an **error** in the SIGCHLD handler if a worker **exits with a nonzero code**. And adds more logging to arbiter halt (that part is okay).
*Sep 2020:* PR #2419 (28df9926d7c650248d267eeafaa8b2c2cdd9af38, ca36d410fd86c82cb2c998a1dab5f7f967405fdd) was created by @mildebrandt trying to fix issue #1663. It logs an **error** in the SIGCHLD handler if a worker **exits with a nonzero code** or **is killed by a signal**.
*Dec 2020:* PR #2475 (b695b497b9b8e7351808848ab9e342219601742a) was created by @aberres trying to fix issue #2215, and immediately merged by @benoitc. It logs a **warning** in the SIGCHLD handler if a worker **is killed by a signal**.
*Apr 2021:* gunicorn 20.1.0 was released containing PR #2475. Soon, issue #2564 was reported: Users started experiencing crashes with `RuntimeError: reentrant call`.
*Oct 2021:* PR #2667 was created by @BoasKluiving to revert PR #2475 in order to fix issue #2564.
*Dec 2021:* PR #2475 was reverted by @benoitc in commit 76f8da24cbb992d168e01bda811452bcf3b8f5b3 in order to fix issue #2564.
*Jan 2022:* PR #2667 was closed by @javabrett because he noticed that PR #2475 was reverted in the meantime.
*Jun 2022:* Issue #2816 was reported by @InfernalAzazel. AFAICT, it's actually just an exact duplicate of issue #2564, because the fix in commit 76f8da24cbb992d168e01bda811452bcf3b8f5b3 wasn't released.
*Dec 2022:* PR #2908 was created by myself. It logs a **warning** if a worker **is killed by a signal**, but the log is written later in the main loop, not directly in the SIGCHLD handler. I did not receive any response in the PR.
*May 2023:* PR #2315 and PR #2419 were merged by @benoitc, reintroducing the `RuntimeError: reentrant call` bug. Issue #2564 was closed (incorrectly).
*Jul 2023:* gunicorn 21.0.0 was released containing PRs #2315 and #2419.
*Aug 2023:* Issue #3050 was reported by @feliixx, who wants the log to be a **warning**, not an **error**.
*Sep 2023:* PR #3064 (this PR) was created by @hydrargyrum, trying to fix issue #2816 by skipping logging in the SIGCHLD handler. (FWIW: you missed one self.log.error call, and nobody except `handle_chld` calls `reap_workers` anyway.)

If I may be frank, I feel very disappointed and frustrated by the gunicorn project. It is a cruel irony that my PR #2908 which attempts to handle this correctly was ignored, but PRs #2315 and #2419 were merged, causing a regression.

But I'll try to swallow my bitter feelings and offer constructive advice.

**My advice to @benoitc is to do this:**

1. `git revert -m 1 1fd1c82195` (revert both commits of #2419)
2. `git revert ee1e5c1928` (revert the bad half of #2315, keep the good half)
3. Review and (if I may be so bold) merge my PR #2908
4. Reject PR #3064 (this PR)

Doing that fixes #2564 properly (at least I think so), fixes #3050, doesn't regress #2215, and removes the need for PR #3064 (this PR).

Hope that helps.

tilgovi:
This is all*extremely* helpful and politely offered advice. I'll try to find time to push this along. I appreciate all your effort here and the exception patience with helping to sort this out.

Most of the maintainers are volunteers and we don't always find the sustained time to reconstruct all of this context, so gathering it in one place is immensely helpful.

aberres:
@TomiBelan Thank you for the write-up. Since my failed approach three years ago I am silently hoping to get the issue fixed for good :)

tilgovi:
I've rebased #2314, which moves reaping to the main thread. That should move the logging out of the signal handler. Any reviews or testing of that would be appreciated. I've assigned myself the related issues to follow up on any other changes afterwards.

piskvorky:
Also bitten by this. I debugged and googled my way to this PR – I have nothing to add to @ja0x and @TomiBelan excellent analysis, so just a +1. I experience the same issue with reentrant logging in `Arbiter.reap_workers()` during `SIGCHLD` https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L242

To make matters worse, in our case gunicorn is auto-launched and monitored from supervisord, running a Flask app. After the unhandled exception due to `RuntimeError: reentrant call inside <_io.BufferedWriter name='<stderr>'>`, the server port was not released correctly, and the newly restarted gunicorn process (by supervisor) keeps printing this forever:

```
[2024-01-25 06:50:00 +0000] [384109] [INFO] Starting gunicorn 21.2.0
[2024-01-25 06:50:00 +0000] [384109] [INFO] starting REST server
[2024-01-25 06:50:00 +0000] [384109] [ERROR] Connection in use: ('127.0.0.1', 8000)
[2024-01-25 06:50:00 +0000] [384109] [ERROR] Retrying in 1 second.
[2024-01-25 06:50:01 +0000] [384109] [ERROR] Connection in use: ('127.0.0.1', 8000)
[2024-01-25 06:50:01 +0000] [384109] [ERROR] Retrying in 1 second.
…
```

I.e. the server is bricked after the unhandled exception, somehow.

On a positive note, I've been using gunicorn for years and this is the first major issue. Kudos and huge thanks to all the maintainers!

randomir:
@TomiBelan, thanks for the excellent summary. I just started digging myself when I discovered it.

In my case it's gunicorn (sometimes) dying on `SIGHUP` during a log rotate.

To maintainers, please note the seriousness of this bug, even if it's intermittent. And the fact it was _re-introduced_ in the same version in which it was fixed (21.0.0) makes it open for 3+ years. :disappointed:

PR #3068, opened 2023-09-08
Add config: buf_read_size

### LINKED ISSUE #2596: Slow processing when receiving large data via POST
Hello,

Being a newbie to GUnicorn, I posted this question on Stackoverflow first, thinking I was getting something wrong, but it looks like it's indeed a problem with GUnicorn and not just an error on my side. I'll describe the problem below, more details and full code to reproduce on the Stackoverflow question here: https://stackoverflow.com/questions/67938278/waitress-and-gunicorn-large-data-input-is-much-slower-than-flask-development-ser/68079761

In summary, if a lot of data is sent to GUnicorn via a POST request, it takes a large time to process (compared to e.g. the Flask development server). A helpful Stackoverflow user has found the blocking point to be the following lines:
```
while size > self.buf.tell():
     data = self.reader.read(1024)
     if not data:
         break
     self.buf.write(data)
```
inside `gunicorn/html/body.py`. I confirmed it, and found that if GUnicorn was receiving 30MB of data (on localhost, so very little I/O overhead), then this snippet alone was taking 0.2s on my machine.

Smelling a slow python while loop (30MB means the loop is done ~30k times) I tested with increasing the amount of data read per step replacing `1024` by `1024**2=1048576`. This helped in my case, reducing time from ~200ms to ~20ms. But I don't know if it's a viable solution (some tests seemed to show that it slowed things down in case the received data was very small, but further tests didn't confirm this, so I'm not sure anymore). Also I haven't checked if the value is optimal (I only noticed that if I further increase to `2**25 = 33554432` it gets slower, about 40ms, so it's not true that more is better or even that hitting the exact amount of data in one go is ideal)

Would you think increasing that amount of data read per step would be feasible? Or would there be a cleaner solution to this problem?

southball:
Resolves #2596.

Add the `--buf-read-size` option to set the block size when reading request body.

### DIFF
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -20,6 +20,20 @@ for reference on setting at the command line.
 
     .. versionadded:: 19.7
 
+
+
+
+.. _buf-read-size:
+
+``buf_read_size``
+~~~~~~~~~~~~~~~~~
+
+**Command line:** ``--buf-read-size``
+
+**Default:** ``1024``
+
+Buffer read size from request data
+
 Config File
 -----------
 
--- a/docs/source/settings.rst
+++ b/docs/source/settings.rst
@@ -20,20 +20,6 @@ for reference on setting at the command line.
 
     .. versionadded:: 19.7
 
-
-
-
-.. _buf-read-size:
-
-``buf_read_size``
-~~~~~~~~~~~~~~~~~
-
-**Command line:** ``--buf-read-size``
-
-**Default:** ``1024``
-
-Buffer read size from request data
-
 Config File
 -----------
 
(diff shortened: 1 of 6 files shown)


docs/source/settings.rst
```diff
@@ -1377,6 +1377,17 @@ Set to ``*`` to disable checking of Front-end IPs (useful for setups
...
+.. _buf-read-size:
+
+``buf_read_size``
+~~~~~~~~~~~~~~~~~
+
+**Command line:** ``--buf-read-size``
+
+**Default:** ``1024``
+
+Buffer read size from request data
```
aamirawan7584, line 1389:
```suggestion
Buffer read size for reading request data from socket
```

gunicorn/config.py
```diff
@@ -208,6 +208,14 @@ def sendfile(self):
 
         return True
 
+    @property
+    def buf_read_size(self):
+        buf_read_size = self.settings['buf_read_size'].get()
+        if buf_read_size is None:
+            return 1024
+
+        return int(buf_read_size)
```
aamirawan7584, line 217: i think we should add some exception handling here if buf_read_size is not convertable to int and return a explanatory error.

aamirawan7584:
@southball do you think we need to use this buffer size config according to content-length because for small request data setting a big buffer size could decrease the performance?

Issue #3079, opened 2023-10-13
worker-class parameter accepts class additionally to class path


odyfatouros:
I have a FastAPI application running on gunicorn with the use of uvicorn workers. I am creating my own custom uvicorn worker because i need to pass specific config parameters to uvicorn that are managed by a central config.yml file by application managers. What i want to do is pass the defined class to the "worker-class" parameter of the gunicorn config and not the class path. This would allow me to define my class inside the function that calls the gunicorn application and also has access to all config parameters.

After investigating the code i found out that actually the util.load_class function has the possibility to distinguish between class and class path (which is str) and return the class. The problem lies in the config module and the properties "worker_class" and "worker_class_str". A simple check about the type of the uri would be enough for the application to start properly with the given class.

tilgovi:
It looks like you created a branch but I don't see a PR. Can you open a PR and I can take a look, please?

tilgovi:
Oh, I'm sorry, I found it at #3079.

benoitc:
i don't see any pull request there... did you close it?

odyfatouros:
No i didn't. I already got one review and I think I need one more in order
for the PR to be merged. I didn't change anything.

Benoit Chesneau ***@***.***> schrieb am Di., 6. Aug. 2024,
22:42:

benoitc:
@odyfatouros i only see code in your repo. Can you point me to the PR?

odyfatouros:
Here it is https://github.com/benoitc/gunicorn/pull/3080

Benoit Chesneau ***@***.***> schrieb am Mi., 7. Aug. 2024,
01:33:

benoitc:
applied thanks.

PR #3080, opened 2023-10-13
Fix for issue #3079, worker_class parameter accepts a class

odyfatouros:


tilgovi (approved):

tilgovi:
Seems reasonable. Any chance you can add a test to `test_config.py? Would be good just to act as documentation and something that would break if someone were to remove the support for loading a class directly in `util.load_class`.

odyfatouros:
Thanks for your feedback. i just added a unit test, i think i covered the case you mentioned.

benoitc:
thanks

PR #2419, opened 2020-09-12
Add additional logs when worker exits abnormally

mildebrandt:
Fixes #1663 

To test out of memory detection:
```
$ cat g.py
print("eating memory")
a = []
while True:
    a.append(bytearray(512000000))

def app(environ, start_response):
    pass

$ docker run -it -m 256m --cpus=0.5 -v $PWD:/app python:3.7 /bin/bash
root@8bf8f22c0928:/# cd app
root@8bf8f22c0928:/app# pip install -e .
Obtaining file:///app
Requirement already satisfied: setuptools>=3.0 in /usr/local/lib/python3.7/site-packages (from gunicorn==20.0.4) (50.3.0)
Installing collected packages: gunicorn
  Running setup.py develop for gunicorn
Successfully installed gunicorn
root@8bf8f22c0928:/app# gunicorn g:app
[2020-09-12 05:37:48 +0000] [18] [INFO] Starting gunicorn 20.0.4
[2020-09-12 05:37:48 +0000] [18] [INFO] Listening at: http://127.0.0.1:8000 (18)
[2020-09-12 05:37:48 +0000] [18] [INFO] Using worker: sync
[2020-09-12 05:37:48 +0000] [21] [INFO] Booting worker with pid: 21
eating memory
[2020-09-12 05:37:52 +0000] [18] [ERROR] Worker (pid:21) was sent SIGKILL! Perhaps out of memory?
[2020-09-12 05:37:52 +0000] [22] [INFO] Booting worker with pid: 22
eating memory
[2020-09-12 05:37:55 +0000] [18] [ERROR] Worker (pid:22) was sent SIGKILL! Perhaps out of memory?
[2020-09-12 05:37:55 +0000] [23] [INFO] Booting worker with pid: 23
eating memory
[2020-09-12 05:37:58 +0000] [18] [ERROR] Worker (pid:23) was sent SIGKILL! Perhaps out of memory?
[2020-09-12 05:37:58 +0000] [24] [INFO] Booting worker with pid: 24
eating memory
```

To test other abnormal exits:
```
$ cat f.py
import sys

def app(environ, start_response):
    sys.exit(5)

$ gunicorn f:app &
$ [2020-09-11 22:52:11 -0700] [19736] [INFO] Starting gunicorn 20.0.4
[2020-09-11 22:52:11 -0700] [19736] [INFO] Listening at: http://127.0.0.1:8000 (19736)
[2020-09-11 22:52:11 -0700] [19736] [INFO] Using worker: sync
[...]

PR #3094, opened 2023-11-14
Update arbiter.py - log SIGTERM as info level message

slimadvies:
I recently started to record gunicorn error messages and everytime I reloaded gunicorn, I received error log messages from every worker. SIGTERM should not generate an error log message. Info level is more appropriate since it is a graceful exit.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -524,7 +524,8 @@ class Arbiter(object):
                     # infinite start/stop cycles.
                     exitcode = status >> 8
                     if exitcode != 0:
-                        self.log.error('Worker (pid:%s) exited with code %s', wpid, exitcode)
+                        self.log.error('Worker (pid:%s) exited with code %s',
+                                       wpid, exitcode)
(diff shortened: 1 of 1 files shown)


benoitc:
please don't break behavior that exist since years. We bneed to check if first the behavior changed here and also how errors are configured at user level. We sometimes uses stderr to not go through the stdout which would be the case with info level.

slimadvies:
How about ignoring the SIGTERM signal then? Because this signal is thrown internally to gracefully shutdown workers. This should NOT generate a error message (introduced by PR #2419)

Parthib:
+1

We recently upgraded gunicorn to the latest version, and this has been a pain point for us. Whenever our containers are taken down for deployment, it emits this error log line even though we exit gracefully. Graceful shutdowns should never result in an error log. This issue is polluting our alarming.

benoitc:
Applied this fix in commit 3e5b102e. SIGTERM is now logged at info level instead of warning, reducing noise during graceful shutdown/reload. Thank you for the contribution!

PR #3095, opened 2023-11-14
Use `utime` instead of `fchmod` in `WorkerTmp.notify`

ramikg:
This PR uses `utime` to update the "Changed" timestamp of the worker's temporary file, instead of calling `fchmod(fd, 0)` & `fchmod(fd, 1)` back-and-forth (which may be problematic in some restrictive environments).

This change was made possible in Python 3.3, which added support for calling `utime` with a file descriptor.


benoitc:
thanks!

PR #3103, opened 2023-12-04
Update wsgi.py fix header string unicode error

hatuan195:
Update to fix header strings encoded in unicode error


benoitc:
The HTTP spec  expect latin1 chars there not utf8 : https://www.rfc-editor.org/rfc/rfc9110.html#section-5.5 You should encode them when you want to provide utf8 inside them.

PR #3108, opened 2023-12-07
restore, and from now on CI-test for entry point

pajod:
1. We want the built package to install a "gunicorn" script.
   * Tell setuptools, now it reads that from `pyproject.toml`
   * Fixes: https://github.com/benoitc/gunicorn/issues/3119
2. We want that script to work and reasonable output when called `gunicorn --help`
   * Have CI `assert` that
3. We want the entry point to work more or less interchangeably with `python3 -m gunicorn` (that is: `__main__.py`)
   * Have CI `assert` that

Driveby fixes:

4. GitHub actions finish faster when the setup-python knows which dependencies are reused
    * potentially unwanted effects for private dependencies - OK, we only use PyPi hosted (and occasionally public GitHub)
    * Verification: In section "Install Dependencies" logs will say "Using cached" in place of "Downloading"
5. Some python tools fail to agree with GitHub runners about console capabilities.
   * Enforce color output
   * Verification: I have seen it work.
6. Removed `cryptography` dependency *again*. 
   * commit 91cb3dc67cb3a711336de8f2bc1f18dfe9ad57bf does not justify including it, if it was indeed needed maybe it should be referenced elsewhere (somewhere not parsed by CI)
   * Verification: Removing it did not noticeably break or slow down CI
7. GitHub actions are versioned. If they improve things, we want that.
   * Since we do not depend on versioned features, we can just bump the major version and automatically use a recent release.
   * Verification: Did not noticeably break or slow down CI

### DIFF
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,4 +1,9 @@
 -r requirements_test.txt
 
+# setuptools v68.0 fails hard on invalid pyproject.toml
+# which a developer would want to know
+# otherwise, oldest known-working version is 61.2
+setuptools>=68.0
+
 sphinx
 sphinx_rtd_theme
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,6 +2,9 @@ name: lint
 on: [push, pull_request]
 permissions:
   contents: read # to fetch code (actions/checkout)
+env:
+  # note that some tools care only for the name, not the value
+  FORCE_COLOR: 1
 jobs:
   lint:
     name: tox-${{ matrix.toxenv }}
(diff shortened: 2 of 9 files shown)


requirements_dev.txt
```diff
@@ -1,4 +1,9 @@
 -r requirements_test.txt
 
+# setuptools v68.0 fails hard on invalid pyproject.toml
+# which a developer would want to know
+# otherwise, oldest known-working version is 61.2
+setuptools>=68.0
```
tilgovi, line 6: Should this be `>` instead of `>=` then?

tilgovi (approved):

requirements_dev.txt
tilgovi: Oh, I see! "Fails hard" is a good thing here :-)

PR #3110, opened 2023-12-08
CI: Pylint 3.x

pajod:
Included:
 * pylint 3.0.0 workarounds (because 3.x is what we want for Python 3.12)
 * removing references to ancient Debian releases (updating which is not helpful as long as the versions distributed remain the same)

Removed:
* everything already merged in https://github.com/benoitc/gunicorn/pull/3131

Not included:
* Arbiter.SIG_NAMES (stdlib since 3.5 provides signal.Signals Enum) and related comments

-----

Suggested order: Merge https://github.com/benoitc/gunicorn/pull/3148 or equivalent first, possibly eradicating the identifier clash in a simpler manner.

### DIFF
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -54,17 +54,18 @@ class Reloader(threading.Thread):
 
 
 has_inotify = False
+inotify = None  # inotify.adapters
+Inotify = None  # inotify.adapters.Inotify
 if sys.platform.startswith('linux'):
     try:
-        import inotify.constants as _
+        from inotify.adapters import Inotify
+        import inotify.constants
         has_inotify = True
     except ImportError:
         pass
 
 
 if has_inotify:
-    from inotify.adapters import Inotify
-    import inotify.constants
 
     class InotifyReloader(threading.Thread):
         event_mask = (inotify.constants.IN_CREATE | inotify.constants.IN_DELETE


cclauss (approved):

.github/workflows/tox.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest] # All OSes pass except Windows because tests need Unix-only fcntl, grp, pwd, etc.
-        python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11", "pypy-3.8" ]
+        python-version: [ "3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "pypy-3.8", "pypy-3.10" ]
```
tilgovi, line 13: Should we add pypy-3.9 here, too?

gunicorn/arbiter.py
```diff
@@ -6,7 +6,7 @@
 import os
 import random
 import select
-import signal
+import signal as _signal
```
tilgovi, line 9: Why this change? To reduce confusion with the `Arbiter.signal` method?

gunicorn/reloader.py
```diff
@@ -54,17 +54,18 @@ def run(self):
...
-    except ImportError:
-        pass
+class OSNoInotifyError(ImportError):
+    pass
+
 
+has_inotify = False
+try:
+    if not sys.platform.startswith('linux'):
+        raise OSNoInotifyError("The inotify mechanism is only supported on Linux")
```
tilgovi, line 64: I think this was maybe nicer when the `try/except` was tightly scoped to just the import.

tilgovi (changes requested):
Would it be possible to just extract the changes to the Python version support and skip the rest, please? I'm not seeing the(pylint comments, inotify import changes, or signal import renaming as relevant to the PR description.

.github/workflows/tox.yml
cclauss: https://www.pypy.org/download.html mentions pypy-3.9 and pypy-3.10, NOT pypy-3.8.

cclauss:
> This branch has git conflicts that must be resolved

tilgovi:
@pajod I'm happy to review the linting-related changes in this PR or separate, your choice. Let's just get the pypy version fixed up and I think we're good to go.

iM-GeeKy:
Is there a timeline when this will be merged?

.github/workflows/tox.yml
benoitc: @pajod can you make the change there?

benoitc:
@pajod why a change to the reloadedr is part of this change?

cclauss:
To fix those two errors, you can change
https://github.com/benoitc/gunicorn/blob/7268a610996eee15e328fce3daf8bf8e748f1627/gunicorn/reloader.py#L63
to read
`    inotify = None`

gunicorn/reloader.py
```diff
@@ -54,17 +54,18 @@ def run(self):
...
+inotify = None  # inotify.adapters
+Inotify = None  # inotify.adapters.Inotify
 if sys.platform.startswith('linux'):
     try:
-        import inotify.constants as _
+        from inotify.adapters import Inotify
+        import inotify.constants
         has_inotify = True
     except ImportError:
         pass
```
cclauss, line 65:
```suggestion
        inotify = None
```

benoitc:
closing since latest version updated to support latest pyylint. Thanks for the patch anyway!

PR #3116, opened 2023-12-15
Set SO_KEEPALIVE on listener socket

### LINKED ISSUE #3115: Setting SO_KEEPALIVE on server socket
I think we should have an option to set SO_KEEPALIVE on the server socket so this will help prevent proxy to close long running connection and also detect disconnected client faster.

cedk:
Closes #3115

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -37,6 +37,7 @@ Brett Randall <javabrett@gmail.com>
 Brian Rosner <brosner@gmail.com>
 Bruno Bigras <bigras.bruno@gmail.com>
 Caleb Brown <git@calebbrown.id.au>
+Cédric Krier <cedric.krier@b2ck.com>
 Chris Adams <chris@improbable.org>
 Chris Forbes <chrisf@ijw.co.nz>
 Chris Lamb <lamby@debian.org>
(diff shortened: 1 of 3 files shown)


tilgovi (commented):
While this might be fine, I think it'd be even easier to say yes and merge with a configuration option. Any chance you want to update the PR to make this configurable?

tilgovi:
While the Linux default is 2 hours before keepalives start, and that is _much_ higher than a typical Gunicorn HTTP keep-alive timeout, it's still a fine idea for cases where someone does set a very high HTTP keep-alive timeout or wants to detect stalled, long-lived requests or something. Maybe they set their system keepalive timeout to be very low.

Nevertheless, it's not enabled at all unless we set this socket option, so I think we probably should make that possible.

tilgovi:
The original linked issue talks about closing long running connections. A good example of where this is useful is probably if someone uses gunicorn with an application that handles WebSocket upgrades.

cedk:
Indeed our host tells us to send keep-alive to avoid the proxy to close the request before it is answered.
But now they came back on this statement and just ask us to use asynchronous request.

So I'm no more sure if this option will be useful for us but I can add the option.

benoitc:
commented related ticket. closing as WON4T DO IT

Issue #1852, opened 2018-08-09
Old bug reproduced：'Response' object has no attribute 'status_code' in wsgi.py with websockets


BoWuGit:
Just like this old [issue 1210](https://github.com/benoitc/gunicorn/issues/1210) said, gunicorn logs error when client disconnects, and my environment is： 

+ Debian GNU/Linux 7.8 

+ nginx 

+ Python3.4 

+ gunicorn(19.8.1) (with one or multiple workers)

+ Flask-SocketIO, client specifies websocket transport 

Everything works well including clients, except for this error log, two cloud independent production instances both persistently logs，but I can't reproduce it in my develop machine, which is a mac. 

Much thanks for your help.

> Error handling request /socket.io/?EIO=3&transport=websocket
Traceback (most recent call last):
  File "/opt/apps/lms/virtualenv/lib/python3.4/site-packages/gunicorn/workers/async.py", line 56, in handle
    self.handle_request(listener_name, req, client, addr)
  File "/opt/apps/lms/virtualenv/lib/python3.4/site-packages/gunicorn/workers/async.py", line 116, in handle_request
    resp.close()
  File "/opt/apps/lms/virtualenv/lib/python3.4/site-packages/gunicorn/http/wsgi.py", line 409, in close
    self.send_headers()
  File "/opt/apps/lms/virtualenv/lib/python3.4/site-packages/gunicorn/http/wsgi.py", line 325, in send_headers
    tosend = self.default_headers()
  File "/opt/apps/lms/virtualenv/lib/python3.4/site-packages/gunicorn/http/wsgi.py", line 306, in default_headers
    elif self.should_close():
  File "/opt/apps/lms/virtualenv/lib/python3.4/site-packages/gunicorn/http/wsgi.py", line 229, in should_close
    if self.status_code < 200 or self.status_code in (204, 304):
AttributeError: 'Response' object has no attribute 'status_code'

benoitc:
do you have any simple example to reproduce it? Also please try with latest master if possible.

BoWuGit:
Previously I have tried several times in my local development environment，which is the same application code with production environment，but I can't reproduce it.  

And I have checked version 19.9.0 release log，not found something related，I'll keep
watching this error log，if find something new，I would post here.

mat10tng:
I too have this problem, specifically when I force all my connection from client to websocket protocol. My settings is the same as BoWuGit. If allow polling protocol before upgrade, this does not show up, but another error :
`
[...]

PR #3126, opened 2023-12-28
Force connection close after protocol upgrade

### LINKED ISSUE #1210: 'Response' object has no attribute 'status_code' in wsgi.py with websockets

tilgovi:
A new protocol, such as WebSockets or HTTP/2, may manage the framing of multiple messages or requests of that protocol over a single connection.

- Close connections after completion of any upgraded response.

- Close connections after a response with status code less than 200. Any such response should be only for a protocol upgrade since applications never see any `Expect` header.

- Ensure that the `status_code` member always exists on instances of the `Response` class (close #1210).


tilgovi:
RFC 6455 states that closing a WebSocket must close the underlying TCP connection:

> After both sending and receiving a Close message, an endpoint considers the WebSocket connection closed and MUST close the underlying TCP connection.

-- https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.1

I couldn't find any language about this in the HTTP/2 protocol, but it seems safer to me to assume that the socket is in an unknown state and any connection persistence is managed at the level of the new protocol.

gunicorn/http/wsgi.py
```diff
@@ -218,15 +219,20 @@ def force_close(self):
...
-        if self.status_code < 200 or self.status_code in (204, 304):
-            return False
-        return True
+        if self.req.should_close():
+            # example: connection close or upgrade header
+            return True
+        if self.upgrade:
+            # close after the new protocol terminates
+            return True
+        if self.response_length is None and not self.chunked:
```
benoitc, line 231: why changing the logic there ? Also this doesn't take care anymore about status code < 100. Can you elaborate? (As a side note I'm not a fan of these one line boolean combination, decoupling in multiple if helps to trace the code .

jblakey-google: @tilgovi are you planning to follow up on this PR? 
It looks like merging it could potentially fix #3126, #1210 and #1852, but I see it's from early 2024.

davidbern2:
Would appreciate an update on this as well! @benoitc if you please get a chance

prav2019:
@benoitc Could you please let us know the updates on this one ? thanks

Issue #2675, opened 2021-11-01
gunicorn cannot run in python 3.10 with gevent = "21.8.0"


ipomoealba:
Hi all, 

We encounter the following errors when we try to upgrade our python environment.

- env: docker.io/python:3.10.0-slim-bullseye
- gunicorn version: "20.1.0"
- gevent version: "21.8.0"

We start gunicorn by the command:
```
gunicorn {{ .Values.flask_app }} --workers "3" --worker-class gevent --bind "[::]:5000" --threads "4" --timeout "300"
```

Error logs
```
[2021-11-01 03:59:04 +0000] [6] [WARNING] Worker with pid 9 was terminated due to signal 15
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 209, in run
    self.sleep()
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 357, in sleep
    ready = select.select([self.PIPE[0]], [], [], 1.0)
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 242, in handle_chld
    self.reap_workers()
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 525, in reap_workers
    raise HaltServer(reason, self.WORKER_BOOT_ERROR)
gunicorn.errors.HaltServer: <HaltServer 'Worker failed to boot.' 3>

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/gunicorn", line 8, in <module>
    sys.exit(run())
  File "/usr/local/lib/python3.10/site-packages/gunicorn/app/wsgiapp.py", line 67, in run
    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
  File "/usr/local/lib/python3.10/site-packages/gunicorn/app/base.py", line 231, in run
    super().run()
  File "/usr/local/lib/python3.10/site-packages/gunicorn/app/base.py", line 72, in run
    Arbiter(self).run()
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 229, in run
    self.halt(reason=inst.reason, exit_status=inst.exit_status)
  File "/usr/local/lib/python3.10/site-packages/gunicorn/arbiter.py", line 342, in halt
    self.stop()
[...]

PR #3134, opened 2023-12-29
CI: minimal dependency versions: specify them and try them in CI

pajod:
We should know our minimum versions, and possibly how they are different depending on python version, e.g. see #2675

This is not the change in minimal dependencies, this is just the necessary duplication (https://github.com/pypa/pip/issues/8085) and hooking it up to GitHub CI. Has everything added as comments though, so after this one is merged, any future PR actually bumping dependencies could have that extra ✅/❌ show up.

The useless header/comment line swap is intentional, works around a bug in GitHub diff syntax highlighting.


benoitc:
Closing this PR as CI configuration has evolved since this was opened over a year ago. Thank you for the contribution.

PR #3147, opened 2024-02-01
partial revert of #2918 due to concurrency issues: Deadlock

### LINKED ISSUE #3146: Gunicorn gthread deadlock
# Problem
On receiving multiple connections at the same time I observe a deadlock that causes gunicorn worker to completely freeze forever (the worker timeout does not kill the frozen worker).

This happens when the number of `worker_connections` is smaller than the number of concurrent connections received.

Affected versions:
- 21.0.0
- 21.0.1
- 21.1.0
- 21.2.0

# How to reproduce

Run a gunicorn app with gthread worker and these options:
```
    options = {
        # ...
        'workers': 1,
        'threads': 3,
        'worker_connections': 4,
    }
```
Then run apache bench with enough concurrency (higher than 4): `ab -n1000 -c100 http://0.0.0.0:8080/`

I am opening a PR soon with the fix and the script to reproduce this easily (UPDATE: PR #3147).

# Proposed solution

~Partial revert of #2918 ~
#3157

javiertejero:
when opening many sockets the worker gets frozen consuming 100% cpu

to reproduce:

```
cd examples
python standalone_app_gthread.py
```

and then using Apache Bench tool:
```
ab -n1000 -c100 http://0.0.0.0:8080/
```

proposed solution: Partial revert of #2918

Fixes #3146

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -83,6 +83,7 @@ Jakub Paweł Głazik <zytek@nuxi.pl>
 Jan-Philip Gehrcke <jgehrcke@googlemail.com>
 Jannis Leidel <jannis@leidel.info>
 Jason Madden <jason@nextthought.com>
+Javier Tejero Jurado <javier.tejero@travelperk.com>
 jean-philippe serafin <serafinjp@gmail.com>
 Jeremy Volkman <jeremy@jvolkman.com>
 Jeryn Mathew <jerynmathew@gmail.com>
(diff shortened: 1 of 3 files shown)


javiertejero:
Most likely this is more convenient but not 100% working in MacOS as of now https://github.com/benoitc/gunicorn/pull/3157

PR #3148, opened 2024-02-01
arbiter: Use waitpid() facilities to handle worker exit status

sylt:
This change is meant to handle the return value of waitpid() in a way that is more in line with the man page of said syscall. The changes can be summarized as follows:

* Use os.WIF(STOPPED|CONTINUED|EXITED|SIGNALED) to determine what caused waitpid() to return, and exactly how a worker may have exited.

* In case of normal termination, use os.WEXITSTATUS() to read the exit status (instead of using a hand rolled bit shift). A redundant log was removed in this code path.

* In case of termination by a signal, use os.WTERMSIG() to determine the signal which caused the worker to terminate. This was buggy before, since the WCOREFLAG (0x80) could cause e.g. a SIGSEGV (code 11) to be reported as "code 139", meaning "code (0x80 | 11)".

* SIGSTOP and SIGCONT are ignored. They are usually used by debuggers for stopping/continuing the process as they attach/detatch.


sylt:
As for the motivation of this change: I got very confused when debugging a gunicorn error log reading "Worker (pid:xxx) was sent code 139!". Since code 139 wasn't anything I immediately recognized, I started digging into the code to get more clues, and I think I found what I was searching for. I thought that this could be made more clear.

When it comes to testing, all automatic tests should pass, and I have tested the following cases manually (with good results):

1. Send SIGSEGV to a worker.
2. Attach debugger (gdb) to worker process.
3. Incidentally, but still, make workers exit normally in different ways (e.g. APP_LOAD_ERROR).

sylt:
Oh no, I see that there already is a myriad of PRs related to signal handling. Sorry about adding another one... this one I did solely to get better log output, but I didn't consider that the logging might be done in the wrong context. Adding a commit which solves that particular problem should probably be feasible as well.

gunicorn/arbiter.py
```diff
@@ -186,12 +188,17 @@ def init_signals(self):
         # initialize all signals
         for s in self.SIGNALS:
             signal.signal(s, self.signal)
-        signal.signal(signal.SIGCHLD, self.handle_chld)
+
+    def write_message_to_pipe(self, one_byte_message):
+        try:
+            os.write(self.PIPE[1], one_byte_message.to_bytes(1))
```
piskvorky, line 194:
```suggestion
            os.write(self.PIPE[1], one_byte_message.to_bytes(1, byteorder='big'))
```

On pre-3.11 Pythons, the signature is:

```python
to_bytes(length, byteorder, *, signed=False)
```

I.e. there was no default for the `byteorder` argument.

piskvorky (changes requested):

gunicorn/arbiter.py
sylt: Thanks for the heads up! :+1: I have incorporated your suggested change in the next change-set!

benoitc:
Thanks for the patch I appreciate. But I'm not sure I agree with this change or #2314.  It  removes the asynchronous way to monitor the workers by mixing it with arbiter signaling. Afaik we don't to wait for a worker signal when we just want to quit Gunicorn or reload it. These messages should be taken in priority among the rest. 

Also likely we want to process all signals related to the workers and not skip one which is done when the signal queue is used. I see this is removed in this patch but  by removing the signal queue you miss the fact that it is used a s a buffer that protect us to receive to much messages to handle. Sometimes this happening due to a badly configured alert handler or just a supervisor like systemd going crazy.  We need to keep it around. This has been added on purpose.

To fix the orignal issue (not having the ability to print without crashing), I think we should handle logging in a separate thread out of the signal handler. We can redirect all messages to a pipe or unix socket that will be consumed by this thread to  print or actually write the logs.  A patch in this direction would not break the logic exposed above and continue.  This thread could also be used to handle stats.  Would you make this patch?

sylt:
@benoitc Thanks for having a look! It's great to know some history as well behind the code -- git blame didn't help me much there unfortunately. I'll answer your findings in the same order as you wrote them:

> It removes the asynchronous way to monitor the workers by mixing it with arbiter signaling. Afaik we don't to wait for a worker signal when we just want to quit Gunicorn or reload it. These messages should be taken in priority among the rest.

Maybe I've misunderstood then, because I believed that the code actually did wait for workers when one quits/reloads Gunicorn? I am basing this on Arbiter.stop():

```python
    def stop(self, graceful=True):
        ... # lots of lines omitted

        self.kill_workers(sig)
        # wait until the graceful timeout
        while self.WORKERS and time.time() < limit:
            time.sleep(0.1)
```
This function won't exit until either the timeout has been reach, or all workers have been removed. So, while the code above is (most likely) executing the time.sleep(0.1), SIGCHLD keeps coming in, removing the workers. Reading #2314, I believe this is the same analysis which @tilgovi has done as well.

Just so that I don't misunderstand, was this what you were referring to?

> Also likely we want to process all signals related to the workers and not skip one which is done when the signal queue is used. I see this is removed in this patch but by removing the signal queue you miss the fact that it is used a s a buffer that protect us to receive to much messages to handle. Sometimes this happening due to a badly configured alert handler or just a supervisor like systemd going crazy. We need to keep it around. This has been added on purpose.

Yes, I suspected that the '5' max limit was due to something like that, but if there is a supervisor process going bananas, spamming signals to a process, I doubt that such a limit would help in a precarious situation like that since the queue would keep being full as it gets new signals? 

The reason the queue was removed has to do with signal safety: I think the signal handler should be kept as simple as possible, with no memory allocation nor print-outs nor locking of any kind. Since I believe the list manipulation in python involves some kind of malloc(), I don't think it's strictly signal safe (although in practice, it might be).

And since I saw that SIG_QUEUE has been around since the [very first revision of arbiter.py](https://github.com/benoitc/gunicorn/commit/8be02267636206914decffa332e55e635a560cca) (in 2009), I thought that maybe things have changed since back then. And since there's no logging when the buffer is full, I didn't know if we actually had any data which supported its existence?

> To fix the orignal issue (not having the ability to print without crashing), I think we should handle logging in a separate thread out of the signal handler. We can redirect all messages to a pipe or unix socket that will be consumed by this thread to print or actually write the logs. A patch in this direction would not break the logic exposed above and continue. This thread could also be used to handle stats. Would you make this patch?

But how would we implement such a proposal, send formatted logs from a signal context to a pipe, without doing any dynamic memory allocation in the signal context? In C I would have understood -- use a static buffer on the stack and use snprintf(). In python however, we don't have that kind of control over memory. Also, then we'll have another thread to manage with all problems and considerations that comes with it (life-time, locking).

So before introducing another thread, I think we need to carefully consider if there is no other way to solve the problem. Although I am of course very biased in all of this, I think the solution offered in this PR solves the logging problem and also manages to remove some complexity rather than adding more on top. 

If you still want to go in this direction however, perhaps creating a separate issue would be a good idea? But please consider these thoughts above!

gunicorn/arbiter.py
```diff
@@ -237,10 +249,12 @@ def run(self):
                 self.pidfile.unlink()
             sys.exit(-1)
 
-    def handle_chld(self, sig, frame):
+    def handle_cld(self):
```
piskvorky, line 252: `SIGNALS` installs only `SIGCHLD`, not `SIGCLD`. What is this alias for?

piskvorky (commented):
I saw @tilgovi also singles out `CLD` in his PR.

Whichever PR gets continued / merged, it may be worthwhile documenting this special treatment / aliasing of `CLD` in the code, to ease future maintenance.

tilgovi:
@sylt Python signal handler code does not get executed in the C signal handler context. I believe the interpreter just executes the signal handler at certain times before returning to the main execution of the script. I think it's perfectly safe to allocate memory and so forth.

So while we don't have to worry about C-level concerns, we do still have to reason about the Python interpreter switching execution between the signal handler code and the main loop.

This is why my PR really only suggests that the benefit is a lack of risk of concurrent modification of the workers list and some code simplification in the form of reaping only happening in one place.

But I think our analysis is otherwise correct.

tilgovi:
https://docs.python.org/3/library/signal.html#execution-of-python-signal-handlers

sylt:
@tilgovi Alright, thanks for the correction, that's a relief to hear it works in such a way! :) Well, then I guess commit 2 in this PR shouldn't be needed from a memory safety point of view, so perhaps I should throw it out.

gunicorn/arbiter.py
sylt: Heh, this is a good question, I had to dig a little bit :)

Short answer: Because self.SIG_NAMES[Signals.SIGCHLD] => 'cld'.

Long answer: Consider the following code:

```python
import signal

# I love dynamic languages
SIGNALS = [getattr(signal, "SIG%s" % x)
               for x in "CHLD HUP QUIT INT TERM TTIN TTOU USR1 USR2 WINCH".split()]
SIG_NAMES = dict(
    (getattr(signal, name), name[3:].lower()) for name in dir(signal)
    if name[:3] == "SIG" and name[3] != "_"
)
```
While SIGNALS looks correct, the bug is inside SIG_NAMES, specifically how we loop over `dir(signal)`. On Linux, this list of attributes contains both the strings 'SIGCHLD' and 'SIGCLD' in this order. So what happens? Well, the same thing as `{ 'foo': 1, 'foo': 2 }`: we end up with `{ 'foo': 2 }`. 

One way to fix this would be to loop over `[s.name for s in signal.Signals]` instead, which doesn't seem to include both SIGCLD and SIGCHLD (only the latter is included).

jcjones1515:
> I actually came here to make a PR to fix it so both CHLD and CLD get handled the same, but it looks like you beat me to it by a few days?

misunderstood what was going on earlier -- it's looking for the wrong alias I think.

just in case this on takes a while to get merged, I made this quick fix as an option: https://github.com/benoitc/gunicorn/pull/3151

sylt:
@benoitc I have made an alternative approach [here](https://github.com/benoitc/gunicorn/compare/master...sylt:gunicorn:dont_remove_workers_in_signal_handler) with the following highlights:

* SIGCHLD is handled separately, just as before.
* The signal queue with rate limiting (< 5 items) is kept just as before.
* In the handler for SIGCHLD, we don't do any logging, we just do waitpid() and save the status to the worker object.
* In the main loop where we manage workers, we go through all workers and see if they have an exit status. If so, print it and then remove the worker so that new ones can spawn.

Would such an approach seem more reasonable?

benoitc:
@sylt i will revisit this PR for 23.1 this month  and come back to you.  

One thinng I would keep is a separate sleep function. I understand you timeout in the queue, but I think it's good to keep a way to wake up the master separately. One of the reason for it is to allow later a way  to optimize the scheduling and extend its usage. I would like to be more near the system than the python stdlib there.

sylt:
@benoitc OK, thanks for giving it a look!

Hmm, instead of adding an Arbiter:sleep() method, I added an Arbiter:wakeup() method, so that "fast" wakeups of the arbiter can be requested (instead of potentially having to wait 1 second before it runs again due to timeout). Would that solve the problem do you think?

benoitc:
> @benoitc OK, thanks for giving it a look!
> 
> 
> 
> Hmm, instead of adding an Arbiter:sleep() method, I added an Arbiter:wakeup() method, so that "fast" wakeups of the arbiter can be requested (instead of potentially having to wait 1 second before it runs again due to timeout). qqqqsWould that solve the problem do you think?

yes this can work. I will review the pr tomorrow as I plane to make a new release during the week

sylt:
@pajod I've reverted the latest two patches! I'm a little bit short on time ATM, but I'll try and answer your comments in near time. But order is at least restored :)

gunicorn/arbiter.py
```diff
@@ -573,7 +535,12 @@ def manage_workers(self):
         workers = sorted(workers, key=lambda w: w[1].age)
         while len(workers) > self.num_workers:
             (pid, _) = workers.pop(0)
-            self.kill_worker(pid, signal.SIGTERM)
+            worker = self.WORKERS[pid]
+            if worker.killed['by'] is None:
+                self.kill_worker(pid, signal.SIGTERM)
+            elif (worker.killed['by'] == signal.SIGTERM
+                  and time.monotonic() > worker.killed['when'] + worker.timeout):
+                self.kill_worker(pid, signal.SIGKILL)
```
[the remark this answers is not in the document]
sylt: Yes, looking at it again, it is true that the USR1 handling didn't fit into the suggestion! Once and if this PR lands, I could look into posting a follow-up PR.

Resolving since the patch is removed from the patch set.

gunicorn/arbiter.py
```diff
@@ -499,11 +467,11 @@ def murder_workers(self):
...
                 continue
 
-            if not worker.aborted:
+            if worker.killed['by'] is None:
                 self.log.critical("WORKER TIMEOUT (pid:%s)", pid)
-                worker.aborted = True
                 self.kill_worker(pid, signal.SIGABRT)
-            else:
+            elif (worker.killed['by'] == signal.SIGABRT
+                  and time.monotonic() > worker.killed['when'] + worker.timeout):
```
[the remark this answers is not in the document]
sylt: Yes, what I sought to fix was that the "_You have 2 seconds!_" threat would be honored :) Because right now, it just depends on how often we enter the `manage_workers()` method, which could be more often than once per second.

gunicorn/arbiter.py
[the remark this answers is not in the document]
sylt: I tried a slightly different approach at implementing this here: https://github.com/sylt/gunicorn/pull/1. Hopefully, the USR1 problem should be solved!

Even though the new approach applies cleanly on master, it depends on work done in this PR. So, I'll take it one thing at a time.

gunicorn/arbiter.py
```diff
@@ -203,25 +202,17 @@ def run(self):
...
-                    self.log.info("Ignoring unknown signal: %s", sig)
-                    continue
+                try:
+                    sig = self.SIG_QUEUE.get(timeout=1)
+                    if sig != self.WAKEUP_REQUEST:
+                        if sig != signal.SIGCHLD:
+                            self.log.info("Handling signal: %s", signal.Signals(sig).name)
+                        self.SIG_HANDLERS[sig]()
+                except queue.Empty:
+                    pass
```
suligap, line 212:
```suggestion
                timeout = 1
                while True:
                    try:
                        sig = self.SIG_QUEUE.get(timeout=timeout)
                    except queue.Empty:
                        break
                    timeout = 0
                    if sig != self.WAKEUP_REQUEST:
                        self.log.info("Handling signal: %s", signal.Signals(sig).name)
                        self.SIG_HANDLERS[sig]()
```

Can you consider some of the suggested changes?:

1. Log the info message unconditionally for all signals.
1. Limit the try block.
1. Make sure we go through all queued signals (without waiting for subsequent ones) before .murder_workers() is called.

suligap (changes requested):
This looks rather fantastic to me. It's not just the regular printing due to logging that was problematic/unsafe here. Gunicorn's configurable `child_exit` was also being called inside of the signal handler.

I've proposed one small change and a couple of nitpicks if you can take a look.

gunicorn/arbiter.py
```diff
@@ -170,27 +171,25 @@ def init_signals(self):
         Initialize master signal handling. Most of the signals
         are queued. Child signals only wake up the master.
```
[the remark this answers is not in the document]
sylt: I removed the docstring, I think the code should be clear enough that a comment isn't needed.

gunicorn/arbiter.py
```diff
@@ -170,27 +171,25 @@ def init_signals(self):
...
+
+        # Some UNIXes require SIGCHLD to be reinstalled, see python signal docs
+        if sig == signal.SIGCHLD:
+            signal.signal(sig, self.signal)
+
+    def wakeup(self, due_to_signal=None):
+        """\
+        Wake up the main master loop.
+        """
+        self.SIG_QUEUE.put(due_to_signal or self.WAKEUP_REQUEST)
```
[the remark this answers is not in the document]
sylt: Sure, I added it where I assign queue.SimpleQueue to SIG_QUEUE. That way, I hope that the comment won't get outdated that easily.

gunicorn/arbiter.py
```diff
@@ -36,16 +36,12 @@ class Arbiter:
...
 
     # I love dynamic languages
-    SIG_QUEUE = []
-    SIGNALS = [getattr(signal, "SIG%s" % x)
-               for x in "HUP QUIT INT TERM TTIN TTOU USR1 USR2 WINCH".split()]
-    SIG_NAMES = dict(
-        (getattr(signal, name), name[3:].lower()) for name in dir(signal)
-        if name[:3] == "SIG" and name[3] != "_"
-    )
+    SIG_QUEUE = queue.SimpleQueue()
```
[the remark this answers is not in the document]
sylt: Sounds good, added!

gunicorn/arbiter.py
sylt: Sure! I did as follows:

1. Made so that SIGCHLD gets logged with debug severity. In the normal case, I think logging with info severity gets too spammy.
2. Sure! I refactored getting the list of signals into its own method, this way I found it easier to limit the scope.
3. Same here, solved by getting a list of signals from a method instead.

gunicorn/arbiter.py
```diff
@@ -168,31 +171,38 @@ def start(self):
...
+        """\
+        Returns list of pending signals (or empty list if timeout)
+        """
+        signals = []
+
+        def append_signal(sig):
+            if sig != self.WAKEUP_REQUEST:
+                signals.append(sig)
+        try:
+            append_signal(self.SIG_QUEUE.get(timeout))
```
suligap, line 201: There's a nasty bug here with `timeout` being passed as the `block` arg to `.get()` (`Queue.get(block=True, timeout=None)`) and preventing `.murder_workers()` and `.manage_workers()` from doing the maintenance.

```suggestion
            append_signal(self.SIG_QUEUE.get(timeout=timeout))
```

suligap (changes requested):
Hi @sylt and thanks for the changes. Found one bug (easily fixed).

Really looking forward to this getting merged.

suligap (approved):
Thank you, +1.

oelhammouchi:
Hi, bumping into this issue right now, are you still looking at having it merged? Thanks a lot for all the great work!

PR #3151, opened 2024-02-06
alias CLD to CHLD in signal handlers

jcjones1515:
image

some systems have 17 mapped to both CHLD and CLD, this prevents a CHLD signal from not being handled in this case

I observed on our systems this issue prevented the arbiter from recognizing a worker getting killed and did not restart it.


benoitc:
@jcjones1515 what systems have this ? Can you give an example?

benoitc:
Closing in favor of the more comprehensive approach in #3148. Thank you for identifying this issue!

PR #3154, opened 2024-02-11
chore: Add clarity to SECURITY.md

chris-griffin:
The security policy currently contains a contradiction where it says that the latest release is the only version supported, but also notes that v21.2.0 is not supported. However, according to the repo, [v21.2.0](https://github.com/benoitc/gunicorn/releases/tag/21.2.0) is the latest release.

image

This updates the policy to make it clear that v21.2.0 is currently eligible for security reports as the latest version and also explicitly mentions master as well.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -12,8 +12,8 @@ At this time, **only the latest release** receives any security attention whatso
 
 | Version | Status          |
 | ------- | ------------------ |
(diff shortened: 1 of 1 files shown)


benoitc:
change applied slightly differently to master. Thanks for the suggestion anyway !

PR #3173, opened 2024-03-23
refactor: replace IOError with OSError

e-kwsm:
<https://peps.python.org/pep-3151/>


benoitc:
Closing as duplicate. IOError is an alias of OSError in Python 3, so the current code is correct. Related cleanups were done in #3189. Thank you!

PR #3179, opened 2024-03-26
fix(deps): exclude eventlet 0.36.0 to avoid WebSocket bug

dhdaines:
Eventlet 0.36.0's WebSocket implementation has a bug that prevents it from running under a Gunicorn worker (as is done by Flask-SocketIO and possibly other things): https://github.com/eventlet/eventlet/issues/946

Probably this will be fixed in the next Eventlet release but Gunicorn should exclude 0.36.0 to avoid nasty surprises downstream!


benoitc:
thanks!

PR #3181, opened 2024-04-03
Add on_stopping hook to set anything before master is killed

ilpianista:
As suggested in #2944. My use case is to postpone the shutdown of gunicorn.


ilpianista:
@benoitc may you have a look at this please? Thanks

PR #3189, opened 2024-04-22
chore: eat Python 2 leftovers

pajod:
* No harm in merge conflicts: can always be trivially regenerated thanks to https://pypi.org/project/pyupgrade/ with its `--py37-plus` flag
* This is an updated version of https://github.com/benoitc/gunicorn/pull/2836

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # Gunicorn documentation build configuration file
 #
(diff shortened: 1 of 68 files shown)


docs/source/conf.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
benoitc, line 1: why removing this line?  can't we still have files using latin 1 or other encoding?

Issue #2650, opened 2021-09-12
Gunicorn's handling of PATH_INFO and SCRIPT_NAME can lead to security issues when placed behind a proxy


itszn:
This might not be something you can fix, but I wanted to report it since people might have vulnerable configs.

Gunicorn [reads some WSGI variables from request headers](https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/wsgi.py#L127). This can be an issue when a proxy allows and passes along these special header values.

For example [`SCRIPT_NAME` is used to transform `PATH_INFO`](https://github.com/benoitc/gunicorn/blob/ff58e0c6da83d5520916bc4cc109a529258d76e1/gunicorn/http/wsgi.py#L183). If a proxy previously parsed the request path to determine if access is allowed, and attacker could now use a `SCRIPT_NAME` header to bypass the proxy.

Here is an example nginx config:

```nginx
server {
    listen 80;
    underscores_in_headers on;
    location / {
        include proxy_params;
        proxy_pass http://unix:/tmp/gunicorn.sock;
        proxy_pass_request_headers on;
        location ^~ /admin/ {
            deny all; # disallow anyone from accessing any routes starting with /admin/
        }
}
```

This request will call the `/admin/something/bad` route in the WSGI app despite nginx's `deny all`.
```python3
requests.get(URL+'/REMOVED/admin/something/bad', headers={'script_name':'REMOVED/'})
```

benoitc:
Fair enough. This is a part of the spec. I think most of the time it is harmless. But what about adding a ˋ—standalone ˋ option? that would  make gunicorn safer

apollo13:
@benoitc Just stumbled upon this. Which spec would mandate reading script_name from the request headers?

javabrett:
@itszn thanks for documenting this.

Noting that `underscores_in_headers` default is `off`, and the potential exploit requires it to be `on` - slight mitigation but can be enabled without understanding the impact.

Would it be possible to document adding:

```
proxy_set_header SCRIPT_NAME "";
```

It's possible that Gunicorn isn't strictly required to _read_ the`SCRIPT_NAME` header, but I expect a lot of reverse-proxy setups are going to be relying on that now by-default.

douardda:
[...]

PR #3192, opened 2024-04-24
22.0.0 regression: We need a better default treatment of SCRIPT_NAME

pajod:
As pointed out in [#2650](https://github.com/benoitc/gunicorn/issues/2650#issuecomment-2064250689),  I have     introduced a regression in the now released 22.0 version. What commonly worked in the *default* installation without additional configuration broke: Passing *dynamic* path information from a front-end when the application lives in a subfolder.

- ✅ Deployments where the application does not require additional path information from the front-end proxy keep working
- ✅ Deployments where the SCRIPT_NAME is statically configured keep working
  - e.g. in the application, in its environment, or in the environment of the gunicorn runtime (including in a systemd unit file)
- ❌ Deployments that depended on the ability of arbitrary requests containing a `SCRIPT_NAME` (underscore, not dash) header to transmit the path where the application lives are broken after updating to gunicorn 22.0
  - and can at this time only be reverted to the previous behaviour by resorting to the [--header-map dangerous](https://docs.gunicorn.org/en/stable/settings.html#header-map) escape hatch. Which is bad, because when intending to only allow one particular header from one particular allowed forwarder, this options re-introduces the security concern by permitting all headers from all sources.

Affected Nginx config might look like this:
```nginx.conf
location /foo {
    proxy_redirect off; proxy_http_version 1.1;
    proxy_set_header SCRIPT_NAME /foo;
    proxy_pass http://unix:/run/www-data/gunicorn_foo.sock;
}
location /bar {
    proxy_redirect off; proxy_http_version 1.1;
    proxy_set_header SCRIPT_NAME /bar;
    proxy_pass http://unix:/run/www-data/gunicorn_bar.sock;
}
```
This is not *necessarily* unsafe. When leaving the defaults of `underscores_in_headers off; ignore_invalid_headers on;` untouched, this header name would not be automatically forwarded, so it has come from Nginx.

> [!CAUTION]
> `SCRIPT_NAME` is not the only header that can lead to dangerous results when an application gets confused about which headers were inserted by an authorized proxy. But is the only one gunicorn absolutely must take care of, being the one gunicorn picks up from headers and places into environ.

### I propose we do:
- Push out a patch release before the affected version even propagates everywhere and people are left with no option other than applying a less specific workaround.
- [x] By default extend "localhost & unix peers are authorized" treatment of *secure scheme* headers also to now-banned field names
  - [x] New list-type option, similar to *secure scheme* headers. There may be other cases where (especially nginx users) want to in-band transmit `map{}-ed` data to the application.
  - The fact that non-IP peers simple ignore the allowed IP list probably deserves documentation. For code readability, it should also be flipped in the code form "ignore anything that is not a tuple" to "ignore for UNIX sockets, specifically, and only then"
- [x] check the syntax of the allowed-ip-addrs setting to make that option easier to figure out
  - see also https://github.com/benoitc/gunicorn/pull/2390 which goes beyond just validation
  - Possibly also change logging around this
- [x] modify the default of allowed-ip-addrs setting to permit the (singular) IPv6 loopback address as well
- [x] Documentation adding examples, explanation, and cross-links to popular frameworks on how SCRIPT_NAME works
  - leading/trailing slashes deserve an extra sentence or two. A common source of confusion.
  - [x] https://docs.gunicorn.org/en/latest/faq.html#how-do-i-set-script-name
  - https://docs.gunicorn.org/en/stable/deploy.html#systemd
    - Documentation & Examples to encourage making use of externally-managed (systemd) or named ("UNIX socket") sockets, eliminating the reason for defaulting to "blindly trust all local users" for those cases.
  - https://docs.djangoproject.com/en/5.0/ref/settings/#force-script-name

## Related issues
* Caused while attempting to address: https://github.com/benoitc/gunicorn/issues/2650
* Fixes: https://github.com/benoitc/gunicorn/issues/3200

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1263,17 +1263,18 @@ class ForwardedAllowIPS(Setting):
     cli = ["--forwarded-allow-ips"]
     meta = "STRING"
     validator = validate_string_to_list
-    default = os.environ.get("FORWARDED_ALLOW_IPS", "127.0.0.1")
+    default = os.environ.get("FORWARDED_ALLOW_IPS", "127.0.0.1,::1")
     desc = """\
         Front-end's IPs from which allowed to handle set secure headers.
         (comma separate).
 
-        Set to ``*`` to disable checking of Front-end IPs (useful for setups
+        Set to ``*`` to disable checking of Front-end IPs. This is useful for setups
         where you don't know in advance the IP address of Front-end, but
-        you still trust the environment).
+        instead have ensured via other means that none other than your
+        authorized Front-ends can access gunicorn.
 
         By default, the value of the ``FORWARDED_ALLOW_IPS`` environment
-        variable. If it is not defined, the default is ``"127.0.0.1"``.
+        variable. If it is not defined, the default is ``"127.0.0.1,::1"``.
 
         .. note::
 
@@ -2364,6 +2365,9 @@ class HeaderMap(Setting):
         The value ``refuse`` will return an error if a request contains *any* such header.
         The value ``dangerous`` matches the previous, not advisabble, behaviour of mapping different
         header field names into the same environ name.
+
+        The (at this time, not configurable) header `SCRIPT_NAME` is permitted
+         without consulting this setting, if it is received from an allowed forwarder.
 
         Use with care and only if necessary and after considering if your problem could
         instead be solved by specifically renaming or rewriting only the intended headers
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -1250,7 +1250,7 @@ class SecureSchemeHeader(Setting):
 
         A dictionary containing headers and values that the front-end proxy
         uses to indicate HTTPS requests. If the source IP is permitted by
-        ``forwarded-allow-ips`` (below), *and* at least one request header matches
+        :ref:`forwarded-allow-ips` (below), *and* at least one request header matches
         a key-value pair listed in this dictionary, then Gunicorn will set
         ``wsgi.url_scheme`` to ``https``, so your application can tell that the
         request is secure.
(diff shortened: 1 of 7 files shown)


jhominal:
How will `forwarded-allow-ips` work when listening on a UNIX socket?

gunicorn/config.py
```diff
@@ -402,6 +403,17 @@ def validate_list_of_existing_files(val):
     return [validate_file_exists(v) for v in validate_list_string(val)]
 
 
+def validate_string_to_addr_list(val):
+    val = validate_string_to_list(val)
+
+    for addr in val:
+        if addr == "*":
+            continue
```
benoitc, line 411: shouldn't we return if wildcard is set?

benoitc:
sounds good for me. Do you plane any other changes?

benoitc:
Thanks for your answer :)  We will launch a 23.1 asap this month with more feature. Let's focus on making a release bringing gunicorn on a par with latest HTTP 1.1 updates for 3.0   

PATH_INFO and SCRIPT_NAME are required  the wsgi spec : https://peps.python.org/pep-3333/#environ-variables  . 

I think it's unfortunate that people don't understand this pooint when they talk about security. As for gunicorn I think that we should provide 2 modes:

1. is running behind a proxy/HTTP server . In this case some secuirtiy could be relaxed and CGI variables shuoudl be passed and secured from this server . This is the advised way until now
2. gunicorn acting itself as main server. In such case we should ensure only Gunicorn and system env can set this variables .

This let me think that a simple "--mode" variable set as "server" or "gateway" could enforce some rules and options. Ie no CGI-VARIABLES should be accepted when it's come from HTTP in server mode. These variables can only be set using OS env or configuration in such case.  In gateway mode  we should probably accept them transparently and let the security benig handled by the server on top. 

I will merge this change fr now but let's revisit this asap following the definition above in a searate ticket.

benoitc:
merged the patch. Thanks :) Discussion can continue there https://github.com/benoitc/gunicorn/issues/3262

PR #3193, opened 2024-04-24
Update util.py

paoloemilioserra:
TypeError: 'int' object is not callable


benoitc:

Issue #3195, opened 2024-04-26
InvalidHTTPVersion / TypeError: not all arguments converted during string formatting


tgz-cz:
Hello,

Our production log contains follwing error:

```
Traceback (most recent call last):
  File "/opt/python/lib/python3.11/site-packages/gunicorn/workers/sync.py", line 134, in handle
    req = next(parser)
          ^^^^^^^^^^^^
  File "/opt/python/lib/python3.11/site-packages/gunicorn/http/parser.py", line 42, in __next__
    self.mesg = self.mesg_class(self.cfg, self.unreader, self.source_addr, self.req_count)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python/lib/python3.11/site-packages/gunicorn/http/message.py", line 257, in __init__
    super().__init__(cfg, unreader, peer_addr)
  File "/opt/python/lib/python3.11/site-packages/gunicorn/http/message.py", line 60, in __init__
    unused = self.parse(self.unreader)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python/lib/python3.11/site-packages/gunicorn/http/message.py", line 281, in parse
    self.parse_request_line(line)
  File "/opt/python/lib/python3.11/site-packages/gunicorn/http/message.py", line 445, in parse_request_line
    raise InvalidHTTPVersion(self.version)
gunicorn.http.errors.InvalidHTTPVersion: <exception str() failed>
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/opt/python/lib/python3.11/site-packages/gunicorn/arbiter.py", line 609, in spawn_worker
    worker.init_process()
  File "/opt/python/lib/python3.11/site-packages/gunicorn/workers/base.py", line 142, in init_process
    self.run()
  File "/opt/python/lib/python3.11/site-packages/gunicorn/workers/sync.py", line 126, in run
    self.run_for_one(timeout)
  File "/opt/python/lib/python3.11/site-packages/gunicorn/workers/sync.py", line 70, in run_for_one
    self.accept(listener)
  File "/opt/python/lib/python3.11/site-packages/gunicorn/workers/sync.py", line 32, in accept
    self.handle(listener, client, addr)
  File "/opt/python/lib/python3.11/site-packages/gunicorn/workers/sync.py", line 158, in handle
[...]

PR #3197, opened 2024-04-26
unhandled exception via invalid configuration

pajod:
Whenever ` "string %r" % single_arg` is used, the string formatting can raise confusing exception whenever `single_arg` is a tuple with more than one item.

This is exposed via invalid configuration read from a python configuration file, or more likely and seriously because of the switched Exception argument type in https://github.com/benoitc/gunicorn/commit/7ebe442d089a6fe0c51abb19a598b3d0d6a6d128 whenever refusing requests based on their (parsed) version tuple. The latter affecting only the case where gunicorn can receive such requests (read: not sitting behind Nginx) and only the 22.0.0 release.

Solution: Make the argument string-only again, and for good measure wrap the string argument in a tuple unconditionally, which also helps with automatically converting it to `"string {!r}".format(single_arg)` or even `f"string {single_arg!r}"` later on.

* Fixes: https://github.com/benoitc/gunicorn/issues/3195
* Suggested merge order: merge https://github.com/benoitc/gunicorn/pull/3196 first (note: now that this is applied, this PR is reduced to stylistic issues and rare edge cases)

### DIFF
--- a/tests/requests/invalid/version_03.http
+++ b/tests/requests/invalid/version_03.http
@@ -1,0 +1,2 @@
+GET /foo HTTP/0.9\r\n
+\r\n
--- a/tests/requests/invalid/version_03.py
+++ b/tests/requests/invalid/version_03.py
@@ -1,0 +1,2 @@
+from gunicorn.http.errors import InvalidHTTPVersion
+request = InvalidHTTPVersion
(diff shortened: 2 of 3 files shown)


washeck:
@pajod I think this PR should be merged instead of mine https://github.com/benoitc/gunicorn/pull/3196

What are the next steps to merge it?

benoitc:
>  "string %r" % single_arg is used, the string formatting can raise confusing exception whenever single_arg is a tuple with more than one item.

I agree for this but why for` %s` ? I think using a tuple where we expect a simple string is kind of a little more confusing, verbose and not natural there when you read it.

Issue #2833, opened 2022-07-26
Add support for PEP 484 Type Hints


kkirsche:
Would adding type hints to the gunicorn codebase be something the team would be open to? If so, what can I do to help with this initiative?

tilgovi:
I closed #2374 with a positive answer. I think it would be a great addition.

kkirsche:
Awesome! Regarding contributing types, do you have a preference for how that's delivered? For example, do you prefer to have smaller, focused merge requests that add various pieces or a larger merge request which lays the full foundation?

tilgovi:
I think smaller contributions are more likely to get reviewed and merged.

kkirsche:
Perfect, thanks for the feedback. So looking at the codebase, the following is the approach that makes sense to me to take, but wanted to outline it so that others are aware and can provide feedback, if desired.

The gunicorn codebase is decently large and comes with complexity. To avoid errors in the type hints, the safest approach is to work from the innermost "edges" of the gunicorn API and work outward to the user-facing API. To elaborate, this means identifying functions, classes, modules, etc. which are self-contained or use other libraries that have existing types. This will allow these methods to be typed appropriately based on the requirements of those functions. This can be done either on a per-file or per-module basis, I lean towards per-file to keep it nice and small.

During this stage, the `py.typed` marker will be added, mypy configuration, etc. Once the innermost edges are complete, I'll start to work outwards to functions that call the typed functions, and work up the API until coverage is complete.

I think the `config` module may be a good place to begin this process from, but I'll review the code more to determine if there is a better starting point. Once identified, I'll create an initial pull request to begin this process and will put a checklist into this issue of the various files to track completeness of the type hints

kkirsche:
It's worth noting, that mypy's variable type hints for classes are only supported on Python 3.6+, so that may delay certain annotations until support for EOL Python 3 versions (per https://devguide.python.org/versions/#versions) is dropped

kkirsche:
Clarified the above statement with alternative that using `.pyi` stubs rather than inline stubs would be an alternative to bypass this issue.

benoitc:
what would be the results of this types? How would they be used?

kkirsche:
Can you be more specific? I don’t understand your question.

With that said, Type hints are a feature of Python that allows you to explicitly declare the data type of a variable when declaring it. They are only available in Python 3.5 and later.

[...]

Issue #2940, opened 2023-02-06
Access log formatting is wrong when using gevent_pywsgi worker class


skolyszewski:
Hi!

When using `gevent_pywsgi` worker class, status code in access log is present as a bytestring, for example `b'404 NOT FOUND'`, instead of a stringified int (e.g. `404`). As you can imagine, this breaks downstream parsing and causes some a stir in our logging aggregation.

Expected:
```
- - - [06/Feb/2023:17:46:12 +0100] "GET / HTTP/1.1" 404 19484 "-" "curl/7.85.0" 3 3553692 -
```
Got:
```
- - - [06/Feb/2023:17:44:01 +0100] "GET / HTTP/1.1" b'404 NOT FOUND' 19754 "-" "curl/7.85.0" 3 3590796 -
```

This issue doesn't occur with `gevent` worker class, nor it should occur with any different classes (based on my limited understanding of gunicorn internals).

As stated, I have very limited understanding of the internals, but the code that would address the issue would be one that doesn't run (because of the magic):

https://github.com/gevent/gevent/blob/1e412d35526183b26c1abf2eb658cbef661f5f70/src/gevent/pywsgi.py#L925
```
        return '%s - - [%s] "%s" %s %s %s' % (
            client_address or '-',
            now,
            self.requestline or '',
            # Use the native string version of the status, saved so we don't have to
            # decode. But fallback to the encoded 'status' in case of subclasses
            # (Is that really necessary? At least there's no overhead.)
            (self._orig_status or self.status or '000').split()[0],
            length,
            delta)
```
(the split()[0] piece).

This happens on gunicorn 19.x and 20.x, gevent 20.9.0.

benoitc:
Closing - gevent_pywsgi is a third-party worker class. Access log formatting issues with custom worker classes should be reported to the worker class maintainer.

PR #3201, opened 2024-05-05
add type annotations and rework some None/unset/default cases

pajod:
Apparently we can ship type annotations in separate `*.pyi` files right now, with no need to await dropping 3.7 compat and no obvious harm to people that have no use for them. Reworking #2377 I ended up with a certainly still imperfect, but already somewhat useful set of annotations. A few default settings in gunicorn popped out as odd:

- setting `--config=""` does not disable reading configuration. the next best thing, setting `--config=/dev/null` loudly complains
- enabling `--reload` has non-obvious side-effects in error handling, and setting `--reload-extra=` is silently ignored when set in isolation
- settings that take a callable put the staticmethod in the next line instead of as a decorator. this way static checkers cannot tell the default callables are never meant to receive the usual `self` method arg.
  - refactoring this requires a minor fix in `docs/gunicorn_ext.py` to keep documentation unaffected
- uid/gid settings default to trying even when not expecting any changes.. is this needed? the only effect appears to be breaking some WSL corner cases (where thats not a no-op but a hard error)
- `gid = abs(gid) & 0x7FFFFFFF` <= now unnecessary and broken for systems that actually use group id > 2**32 (see #3212)
- setting --gid=number without setting --uid=number is.. surprising. There is still that root group in the supplementary group list. Looks like [Debian did not like this and patched it](https://sources.debian.org/src/gunicorn/20.1.0-6/debian/patches/0001-drop-supplemental-groups.patch/) - though I would probably branch the initgroups case.
- pytest-cov does not mix well with running just coverage run -m pytest (telling me no coverage was collected..) - is it even needed for anything other than invocation brevity, which is already dealt with by inclusion in tox & Makefile?
- There is a str/bytes type confusion in the `gevent_pywsgi` worker class (but this was already noticed independently, see #2940)

This is primarily an invitation to discuss potential default settings changes, the attached patches serve as clarification.

### Advantages of shipping type annotations:
- highlight oddities in current behaviour stemming from ambiguous/overloaded methods
- can be checked in CI: automatically receive notification before introducing potential type confusion bugs
- assists in understanding existing code, even where currently little to no documentation exists
- assists in reviewing  changes which incorporate type changes, easing complex rebase work of outstanding PRs
- helpful in extracting a list of items to be touch when planning to reclassify previously public/unspecified API into private
- enhances IDE features such as quickly navigating to definitions of types or method arguments

### Disadvantages of shipping type annotations:
- people submitting new PRs without updating annotations may see less self-explanatory feedback from CI runners
- as long as any annotation is incorrect or incompatible with a popular tool such as mypy, expect some people to show up and report those bugs in annotations at a time where there already are quite a lot open bugs/PRs

### Related
Fixes: #2833 #2374 #1393

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest] # All OSes pass except Windows because tests need Unix-only fcntl, grp, pwd, etc.
+        os: [ubuntu-latest, macos-13] # All OSes pass except Windows because tests need Unix-only fcntl, grp, pwd, etc.
         python-version:
          # CPython <= 3.7 is EoL since 2023-06-27
          - "3.7"
--- a/gunicorn/workers/ggevent.py
+++ b/gunicorn/workers/ggevent.py
@@ -169,7 +169,7 @@ class PyWSGIHandler(pywsgi.WSGIHandler):
         finish = datetime.fromtimestamp(self.time_finish)
         response_time = finish - start
         resp_headers_bytes = getattr(self, 'response_headers', [])
-        resp_headers = [(n.decode(), v.decode()) for n,v in resp_headers_bytes]
+        resp_headers = [(n.decode(), v.decode()) for (n, v) in resp_headers_bytes]
         resp = GeventResponse(self.status.decode(), resp_headers, self.response_length)
         if hasattr(self, 'headers'):
             req_headers = self.headers.items()
--- a/gunicorn/workers/gtornado.py
+++ b/gunicorn/workers/gtornado.py
@@ -74,7 +74,7 @@ class TornadoWorker(Worker):
         req_headers = []
         # FIXME: unfinished
         sent = 0
-        response_time = datetime.timedelta(seconds = 1000.0 * request.request_time())
+        response_time = datetime.timedelta(seconds=1000.0 * request.request_time())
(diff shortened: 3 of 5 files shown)


benoitc:
Thanks for the patch. However I don't think having type annotation  external to the project will be convenient on the long term. This introduces too much constraints as you mention.

Also do we really need to introduce type annotation? What would be the addition there for the project? I think the synttax makes the code barely readable and only adapted for those runing an IDE or vscode. Also I don't want to go to the path we have to set specific rules or properties to ignore some results during type checking.  IMO we would better invest in adding property base checking or equivalent there.

Dreamsorcerer:
> only adapted for those runing an IDE

It would make it easier to maintain projects that depend on this library (I personally don't use an IDE, but we use very strict type checking throughout all our projects with mypy). For example, we don't have anyone on aiohttp at the moment that is familiar with the gunicorn worker, so more type information that can be verified with a type checker would make it easier for us to maintain.

I'd also argue that those annotations may actually improve readability. Looking at the diff, most annotations are single word types. Including them means that I can see what types are being used at a glance, whereas right now I'd have to dig through the code figure that out, or just guess.

sth:
It would make sense to crate an independent `types-gunicorn` package and move the `*.pyi` files there. Then the main `gunicorn` can remain unchanged and untyped, but users of the library can still choose to get type checking of the interface by additionally installing the `types-gunicorn` package.

Issue #2742, opened 2022-02-08
Missing test cases for workers


ankitkpandey1:
Issues: There are only import test case for gevent and eventlet workers. I think there should be more cases needed to be covered for these workers or is it sufficient for now?

benoitc:
no activity since awhile. closing feel free to create a new ticket if needed.

PR #3127, opened 2023-12-28
Use plain socket objects instead of wrapper classes

tilgovi:
Refactor socket creation to remove the socket wrapper classes so that these objects have less surprising behavior when used in worker hooks, worker classes, and custom applications.

Close #3013.

Issue #1170, opened 2015-12-28
test OSX and other platforms in CI


benoitc:
we should automate tests on all platform so things like d199e76cea45bb006eb15748da3fcb24754b21f0 can never happen again. 

IS there any CI service open-source friendly around that allows that? If not I can borrow a machine to someone that knows how to setup such things :)

Note the CI should support the following:
- Linux
- OSX/ Darwin
- FreeBSD
- NetBSD
- OpenBSD
- DragonflyBSD

if I forgot an OS just let me know. Also we should prepare us to test gunicorn on windows. 

Feedback/Help is welcome!

benoitc:
https://buildkite.com looks interesting for that. Any opinion?

tilgovi:
Indeed. Never used it. I don't see alternatives other than Jenkins or some other such thing we host ourselves.

benoitc:
gitlab allows to have multiple runner we can plug to it:

https://gitlab.com/gitlab-org/gitlab-ci-multi-runner

I'm thinking to switch to it, at least for the CI. Thoughts?

benoitc:
cc @tilgovi @berkerpeksag

berkerpeksag:
Anything we won't have to maintain and can be used with GitHub is good IMO :) Do we have need to support all *BSDs? We can probably start with FreeBSD and OpenBSD. I can ping some FreeBSD people and ask for help if we can't find a suitable CI service.

benoitc:
stalled issue

PR #3210, opened 2024-05-12
Python 3.4+: use os.set_inheritable + os.set_blocking in place of unreadable fcntl calls

pajod:
We should cleanup those rather unreadable low-level fcntl calls. The stdlib provides single-call functions for these now.

(Replacement is not 100% equivalent, as [they might call](https://github.com/python/cpython/blob/ec1398e117fb142cc830495503dbdbb1ddafe941/Python/fileutils.c#L1504) `ioctl(.., FIO.., NULL)` instead, but I see no reason we should care.)

* Fixes: #2742
* Related: #1170
----
Suggested order: 
* Merge #3211 before this one
* Merge before #3127 - alternatively verify `.fileno()` suffixes that might to be dropped

### DIFF
--- a/.github/workflows/illumos.yml
+++ b/.github/workflows/illumos.yml
@@ -36,9 +36,10 @@ jobs:
           # /tmp/.nginx must exist because nginx will not create configured tmp
           # build-essential shall point to suitable gcc13/gcc14/..
           # TODO: replace python-MN with version-agnostic alias
-          # release: "r151048"
+          # pinning below r151050 to avoid gcc14 incompat
+          release: "r151050"
           prepare: |
-            pkg install pip-312 python-312 sqlite-3 nginx gcc14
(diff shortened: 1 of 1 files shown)


benoitc:
Not disagreeing about this change but this must be validated on BSD. I think we need to add FreeBSD to our test platform to ensure it works there.  (an dmaybe other OpenBSD).

benoitc:
i will look at this branch but I am questionning the goal pursued there. One of the point of keeping low level though is to not depends on how one external dev think things should work.  Knowing and controlling low level access allows us to understand and track more accurately some issues. Thoughts?

benoitc:
Closing this PR for now. The idea of using os.set_inheritable and os.set_blocking instead of fcntl calls is good for code modernization - we'll revisit this in a dedicated cleanup PR. Thank you for the contribution.

PR #3217, opened 2024-05-22
CI: verify docs are up to date

pajod:
New CI task: complain if `docs/source/settings.rst` has changed after invoking `sphinx-build`, indicating someone (me) has forgotten to run `make -C docs html`

* Its not a tox env because I found it easier to ask git whether the file has changed.
* It will also fail if sphinx-build fails. This is good. We should know when that happens (note however, that this does not yet fail on a mistyped reference. that is, so far, just a warning.)

Proof that this works can be found in the earlier CI run of this very PR:
 * https://github.com/benoitc/gunicorn/actions/runs/9185099713/job/25258482749
 * it fails - which is to be expected, because only a later commit fixes includes the updated rst file.

Assumption:
* no need to duplicate between `news.rst` / `2023-news.rst` - just move the content to referenced-in-toc subfiles when the main news file gets large.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,13 +7,18 @@ env:
   FORCE_COLOR: 1
 jobs:
   lint:
-    name: tox-${{ matrix.toxenv }}
+    name: ${{ matrix.python-version }} / tox-${{ matrix.toxenv || '(other)' }}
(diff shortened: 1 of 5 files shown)


benoitc:
thanks!

Issue #3227, opened 2024-06-13
Waiting socket reads block worker restart


michdan:
I have noticed in logs that there is regular GreenletExit exception printed - this do not create an error in clients but it has impact on a performance.
Steps to reproduce:

Server configuration:
Gunicorn version: 22.0.0
Gevent version: 24.2.1
Python version: 3.11.9
Gunicorn settings:
max_requests = 2 (low value only in order to show problem faster)
timeout = 60
worker_class = 'gevent'
workers = 1 (only 1 worker in order to show problem better)
graceful_timeout = 30 (default)
keepalive = 75

The important thing here is that keepalive > graceful_timeout

Client:
Although server has a keepalive=75 client does not reuse connections and each time opens new one. Client sends requests in a loop in parallel 

Observations:
When client and server are running, from time to time server restarts worker and during it stops accepting new connections as  and waits until graceful period  passes- client is waiting for 30 seconds to be able to continue processing. After graceful period there is a message in logs:

```[2024-05-28 13:54:12 +0000] [17990] [ERROR] Error handling request (no URI read)

Traceback (most recent call last):

File "/usr/local/lib/python3.11/site-packages/gunicorn/workers/base_async.py", line 48, in handle

req = next(parser)

^^^^^^^^^^^^

File "/usr/local/lib/python3.11/site-packages/gunicorn/http/parser.py", line 42, in __next__

self.mesg = self.mesg_class(self.cfg, self.unreader, self.source_addr, self.req_count)

^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

File "/usr/local/lib/python3.11/site-packages/gunicorn/http/message.py", line 257, in __init__

super().__init__(cfg, unreader, peer_addr)

File "/usr/local/lib/python3.11/site-packages/gunicorn/http/message.py", line 60, in __init__

unused = self.parse(self.unreader)

^^^^^^^^^^^^^^^^^^^^^^^^^

File "/usr/local/lib/python3.11/site-packages/gunicorn/http/message.py", line 269, in parse

self.get_data(unreader, buf, stop=True)

File "/usr/local/lib/python3.11/site-packages/gunicorn/http/message.py", line 260, in get_data

data = unreader.read()

^^^^^^^^^^^^^^^

File "/usr/local/lib/python3.11/site-packages/gunicorn/http/unreader.py", line 37, in read

[...]

PR #3228, opened 2024-06-13
3227: shutdown socket when worker is not alive

michdan:
More info: https://github.com/benoitc/gunicorn/issues/3227

<s>Simply close the waiting reads (pending due to keepalive) and allow the worker to restart without having to wait for the graceful period.

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -17,6 +17,7 @@ from gunicorn.http.errors import (
 from gunicorn.http.errors import InvalidProxyLine, ForbiddenProxyRequest
 from gunicorn.http.errors import InvalidSchemeHeaders
 from gunicorn.util import bytes_to_str, split_request_uri
+from gunicorn.util import Status
 
 MAX_REQUEST_LINE = 8190
 MAX_HEADERS = 32768
@@ -31,7 +32,7 @@ VERSION_RE = re.compile(r"HTTP/(\d)\.(\d)")
 
 
 class Message(object):
-    def __init__(self, cfg, unreader, peer_addr):
+    def __init__(self, cfg, unreader, peer_addr, status=None):
         self.cfg = cfg
         self.unreader = unreader
         self.peer_addr = peer_addr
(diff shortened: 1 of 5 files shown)


michdan:
> Would prefer a more obviously consistent behaviour for [both GreenletExit](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/ggevent.py#L129) origins. This looks more like a spot-fix to me. Suggest awaiting / working towards something we can run from pytest that gracefully restarts the worker in each of the possible states (idle, keepalive, busy) and asserts the workers behave consistently and without emitting the warning (so far, I only [have a test](https://github.com/benoitc/gunicorn/pull/3201/files#diff-ed3b8af48cd21dc1ac735b839d89b172a6f45569c324af70a2bdb4f60e5fbcdd) for auto-reload of _idle_ workers, and not even that one is close to ready for merge).

Thanks for the quick response!

This is indeed a quick fix. It's up to you whether you want to release it before making major changes, but this fix would have a positive impact on Gunicorn's performance. It's not about handling GreenletExit, but rather preventing it and making worker restarts faster. To demonstrate this, I've created a repository with some showcases and recorded a screen from those showcases. You can watch it on YouTube here: https://youtu.be/xqv4vGKjGBM

michdan:
I have moved it to a draft because it would require comprehensive testing, there is a risk that it can introduce another bug.
However, the problem described in https://github.com/benoitc/gunicorn/issues/3227 is still unresolved

michdan:
> I have moved it to a draft because it would require comprehensive testing, there is a risk that it can introduce another bug. However, the problem described in #3227 is still unresolved

I have an idea on how to resolve #3227, and I believe the current change should work fine. This merge request does not add any additional operations on the socket as before but instead makes the 'is_alive' property shared between objects that make decisions based on it.
I tested it manually but of course more tests are required. But after applying this change ggevent worker is working much better. @pajod @benoitc

michdan:
New version soon

PR #3236, opened 2024-07-04
3227: Wait for the keepalive period to end before stopping the acceptance of new requests

michdan:
This change will prevent potential disruptions in request handling. 
Currently, when a worker is restarted due to the max_request limit and the keepalive period is greater than zero, there is a gap between stopping the acceptance of new requests and starting a new worker. This happens because the process must wait for the lesser of the grace period and the keepalive period, even if there are no pending requests. It waits due to open connections that remain active for the duration of the keepalive period.
cc: @benoitc @pajod


michdan:
To test such functionality, consider using an approach inspired by Apache httpd: https://github.com/apache/httpd/blob/trunk/test/modules/core/test_002_restarts.py. This approach involves making assertions based on log output. While it would require implementing some additional tools, this method seems to simplify testing complex scenarios such as multiprocessing and greenlets.

gunicorn/workers/ggevent.py
```diff
@@ -117,6 +121,19 @@ def run(self):
         except Exception:
             pass
 
+    def _wait_for_keepalive(self, servers):
+        # Retrieve all active greenlets and repeatedly check, until the keepalive period ends,
+        # if any of those greenlets are still active. If none are active, exit the loop.
+
+        greenlets = {id(greenlet) for server in servers for greenlet in server.pool.greenlets}
```
michdan, line 128: When this line is executed, self.alive is already False, which means that all keepalive connections are either closed or are being handled by the existing greenlets. No new keepalive connections will be created until the worker restarts.
Therefore, if the greenlets found in this line have finished, we do not need to wait any longer, as we are assured that there are no more active keepalive connections.

gunicorn/workers/ggevent.py
```diff
@@ -117,6 +121,19 @@ def run(self):
...
+        ts = time.time()
+
+        while time.time() - ts <= self.cfg.keepalive:
```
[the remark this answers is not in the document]
michdan: Thanks for pointing this out, I changed it to use monotonic clock. Btw, shouldn't it be changed also in other places? I can see for example: https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/ggevent.py#L98

gunicorn/workers/ggevent.py
```diff
@@ -87,6 +87,10 @@ def run(self):
             self.notify()
             gevent.sleep(1.0)
 
+        # Wait for pending keepalive connections to complete before stopping the acceptance of new requests
+        if self.cfg.keepalive:
```
benoitc, line 91: i disagree there, we shouldn't care about waiting for  keepalive connections when closing. CLose should be done as fast as possible.

gunicorn/workers/base_async.py
```diff
@@ -36,7 +36,8 @@ def handle(self, listener, client, addr):
             parser = http.RequestParser(self.cfg, client, addr)
             try:
                 listener_name = listener.getsockname()
-                if not self.cfg.keepalive:
+                # do not allow keepalive if the worker is about to be restarted
```
benoitc, line 39: this change is OK

benoitc (changes requested):
I think only  the first part of this change (ensuring we don't process keepalive during closing is OK. 

I disagree with the second part. Can you update your patch to only take care of the first one?

gunicorn/workers/base_async.py
michdan: ok

michdan:
@benoitc, I have already provided a feedback (I am writing this explicite because MR is still tagged as waiting for feedback)

PR #3237, opened 2024-07-09
don't fail upon error while chdir()ing *before* loading config

svmhdvn:
When gunicorn is invoked by `root` using `# su -m <nonprivileged_user> -c gunicorn --config <absolute_path_to_config_file> --chdir <some_world_accessible_directory>`, the following sequence of steps occurs:

1. `su` runs `gunicorn` as the `<nonprivileged_user>`, BUT with `PWD=/root` due to the `-m` flag
2. Due to https://github.com/benoitc/gunicorn/commit/e908ec33594f50619b7aae2d29e6fc15454a9338 , gunicorn takes its current working directory from the `PWD` envvar (QUESTION: is this the right behaviour? It seems to me that `gunicorn` should just use `os.getcwd()` every time.)
3. gunicorn then reaches the code from this patch
4. the config parser correctly parses `cfg.chdir = <some_world_accessible_directory>`, but hasn't yet set `self.cfg = cfg`
5. gunicorn fails with `self.chdir: ERROR: permission denied` after trying to `chdir()` to `/root`

This patch makes this event a warning instead of a failure and continues onwards with searching for the config file. Afterwards, we will `chdir()` to the correct user-configured location as normal.


benoitc:
@pajod gunicorn is supposed to be run from shell if it's wanted. 

I think however that we should exit when such error happen instead of continue since it can be a security issue. @svmhdvn  can you rework your patch for it.

svmhdvn:
I won't be using gunicorn for a while now so I don't have the time to change and test the patch, however someone can definitely take my work!

PR #3244, opened 2024-07-12
reloader: follow symlinks for reload_extra_files

cfelder:


benoitc:
@cfelder can you dcument this change in the config module?

Issue #3251, opened 2024-07-30
Add option to purge largest worker


TyeMcQueen:
In getting Gunicorn running in Kubernetes and tuning resource specifications, I found that the memory consumption of each pod would just grow over time.

In the pull request I'll attach to this issue, I implemented 3 new config options that allowed me to make the memory consumption quickly flatten out and stay stable (and lower) over long periods with Production load.  Example configuration:

```
import psutil

def proc_vmsize(pid):
    # Return how many MB of virtual memory is being used by a worker process
    try:
        p = psutil.Process(pid)
        mb = p.memory_info().vms/1024/1024
        return mb
    except psutil.NoSuchProcessError: # In case worker dies on its own
        return 0

prune_seconds = 5*60            # Prune largest worker every 4.5-5m
prune_function = proc_vmsize    # Measure worker size in MB of VM
prune_floor = 300               # Don't kill workers using <= 300 MB of VM
```

The crux of the problem is that relatively rare requests are much larger (or generate much larger responses or load a lot more resources when computing the response) than typical requests and cause Python to allocate quite a bit more memory and hold on to that memory until the process is killed. Over time, the number of oversized workers would grow and the size of the oversized workers would also creep up, causing pod memory consumption to just keep growing.

This is somewhat counteracted by max_requests, but not very effectively. Setting max_requests low enough to make a dent in this growth can make process restarts so frequent during busy times that you are likely to get overlapping restarts reducing your capacity (and spending more CPU doing all that restarting). And then it is still not effective when traffic volume drops during slow times of day.

Having each worker kill itself when it gets too big risks unbounded and uncoordinated restarting which could be a disaster.

So I added a time-based controlled restart of just the worker that is currently the largest, using whatever measure you prefer or that best fits how your code behaves.

The time-based approach works well in Kubernetes because when traffic ramps up, the number of pods increases and the rate of workers being restarted also goes up proportionally.  And yet you never have two workers in the same pod restarting even close to the same time, so no significant dips in capacity.

And the prune_floor allows you to stop doing any restarts during periods when they aren't really needed.

TyeMcQueen:
> The documentation sounds a lot like leaking memory was in any way normal.

The documentation and this issue description have no mention of memory leaks and memory leaks are not what is being addressed by this feature. Granted that the "the size of the oversized workers would also creep up" (only from the issue description) would fit a memory leak, but it was actually talking about something quite different.

[...]

PR #3252, opened 2024-07-30
Add options to purge big workers

TyeMcQueen:
Described in Issue #3251


gunicorn/arbiter.py
```diff
@@ -203,6 +204,13 @@ def run(self):
 
             while True:
                 self.maybe_promote_master()
+                if 0 < self.cfg.prune_seconds:
+                    if self.next_prune is None:
+                        self.next_prune = time.monotonic() + self.cfg.prune_seconds
+                    elif self.next_prune <= time.monotonic():
+                        self.prune_worker()
+                        self.next_prune += self.cfg.prune_seconds * (
+                            0.95 + 0.10 * random.random())
```
sylt, line 213: I wonder if it would make sense to put all this logic into prune_worker? I mean, something like:

```python
    def prune_worker(self):
        """\
        Kill the worker with highest prune score
        """
        if self.cfg.prune_seconds <= 0:
            return

        if self.next_prune is None:
            self.next_prune = time.monotonic() + self.cfg.prune_seconds

        if time.monotonic() < self.next_prune:
            return
        
        maxi = self.cfg.prune_floor
        victim = 0
        for pid in list(self.WORKERS.keys()):
            score = self.cfg.prune_function(pid)
            if maxi < score:
                maxi = score
                victim = pid
        if victim != 0:
            self.log.info(f"Pruning worker (pid: {victim}) with score {score}")
            self.kill_worker(victim, signal.SIGTERM)

        self.next_prune += self.cfg.prune_seconds * random.uniform(0.95, 1.05)
```

... and then call `self.prune_worker()` just after `self.maybe_promote_master()`? This has the benefit of keeping all worker prune logic at the same place.

I had two other minor suggestions as well (incorporated in the proposed method above):

 * Would `random.uniform(0.95, 1.05)` work instead of `0.95 + 0.10 * random.random()`? I recon it's probably the same, but the former is a little bit shorter.
 * Since we're not interested in the workers themselves but only their PIDs, perhaps calling `self.WORKERS.keys()` would suffice?

Issue #3144, opened 2024-01-31
fix header value to be compliant with new RFC 9110


kenballus:
I reported this privately on June 2, so I'm reporting it publicly now.

Try sending Gunicorn the following payload:
```
GET / HTTP/1.1\r\n
Host: a\r\n
Useless:\n\nGET / HTTP/1.1\nHost: a\r\n
\r\n
```
Gunicorn will see an HTTP request with two headers, `(b'host', b'a')`, and `(b'useless', b'\n\nGET / HTTP/1.1\nHost: a')`.

Since the `\n`s persisted into the header value, this technically violates RFC 9110, which says this:
> Field values containing CR, LF, or NUL characters are invalid and dangerous, due to the varying ways that implementations might parse and interpret those characters; a recipient of CR, LF, or NUL within a field value MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message.

Gunicorn will also violate this rule when it receives a header containing `\x00` or `\r`.

However, the more interesting consequence of this bug is that some HTTP proxies (lookin' at you, OpenLiteSpeed) accept and forward bare LF line endings without normalizing them into CRLFs. This renders Gunicorn vulnerable to request smuggling when used in conjunction with those servers.

benoitc:
> This renders Gunicorn vulnerable to request smuggling when used in conjunction with those servers.

Which chain expose that ?  Was it tested on latest master? Also Shouldn't it be the responsibility to test the headers it's using are safe?

kenballus:
Yes, this was tested on the latest master.

Any proxy that doesn't normalize LF into CRLF as a header line terminator would work. This is a pretty easy mistake to make. One proxy that makes this mistake is OpenLiteSpeed.

So the chain would be this:
Client -> OpenLiteSpeed -> Gunicorn

benoitc:
WSGI applications must check there header to be safe.  I don't see in the current example how an application could be in atatcked unless it's using an header  like this. We protect already against long lines. 

However it's true we should be compliant with the HTTP RFC. A fix for it is needed.

kenballus:
> I don't see in the current example how an application could be in atatcked unless it's using an header like this. We protect already against long lines.

I'll give you a PoC then.

Suppose you have OpenLiteSpeed acting as a reverse proxy for Gunicorn.

OpenLiteSpeed treats `\n` as a line ending in field-lines. This is permitted by RFC 9112 section 2.2:
[...]

PR #3253, opened 2024-07-30
Refuse requests with invalid and dangerous CR/LF/NUL in header field value, as demanded by rfc9110 section 5.5

pajod:
**Backwards incompatible behavior** - but only in clear-cut cases of requests that are, and always have been, invalid and dangerous. Newer standards are more explicit in that they are not to be tolerated.

* first commit is just to clean up diff view, should be empty if applied after #3217
* drive-by change: attach deprecation notices to 22.0 escape hatch options
* fixes: #3144


benoitc:
this will need to be a 23 release though since it's introducing an incompatible behavior.

Issue #3245, opened 2024-07-17
Gunicorn doesn't strip tabs and spaces from header values on the right


kenballus:
The RFCs specify that header values can be prefixed and/or suffixed with any number of spaces and/or tabs, and that this whitespace should be ignored.

Currently, Gunicorn only ignores this whitespace on the left. For example, if you send the following request to Gunicorn:
```
GET / HTTP/1.1\r\n
Host: a\r\n
Test: \t abc \t \r\n
\r\n
```
...it sees a `Test` header value of `abc \t `, but it should see `abc`. Most other HTTP implementations, including AIOHTTP, Apache httpd, Cheroot, Go net/http, H2O, HAProxy, Hyper, Hypercorn, Jetty, Libsoup, Lighttpd, Mongoose, Netty, Node.js, LiteSpeed, Passenger, Tomcat, Tornado, Twisted, Unicorn, Uvicorn, Waitress, and WEBrick, strip the whitespace appropriately.

PR #3254, opened 2024-07-31
Strip trailing optional whitespace (OWS) from single-line header field values

pajod:
**Backwards incompatible behavior** - but only in rare instances of trailing padding on header field values. Old behavior was wrong, new behavior is sanctioned by https://datatracker.ietf.org/doc/html/rfc9112#name-field-syntax

* <del>first commit is just to clean up diff view, should be empty if applied after #3217</del>
* drive-by change: introduce a new HTTP error class & <del>config to</del> squish deprecated folding altogether
* drive-by change: handle the other newly introduced HTTP error classes as well
* Fixes: #3245


benoitc:
The patch doesn't have to be suc complicated IMO ;) The *updated* specifications sasys explicitely:

> Field values containing CR, LF, or NUL characters are invalid and dangerous, due to the varying ways that implementations might parse and interpret those characters; a recipient of CR, LF, or NUL within a field value MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message. Field values containing other CTL characters are also invalid; however, recipients MAY retain such characters for the sake of robustness when they appear within a safe context (e.g., an application-specific quoted string that will not be processed by any downstream HTTP parser).

And 

> A field value does not include leading or trailing whitespace. When a specific version of HTTP allows such whitespace to appear in a message, a field parsing implementation MUST exclude such whitespace prior to evaluating the field value.


So let's just handled it as the default and remove the option.  We will release a major version of it. Can you take care of it?

gunicorn/config.py
```diff
@@ -2241,6 +2241,24 @@ class PasteGlobalConf(Setting):
         """
 
 
+class RefuseObsoleteFolding(Setting):
```
benoitc, line 2244: let just make it the default if it's not supported by the specidication.

benoitc (changes requested):
can you make it the default since it's part of the spec.

also please make sure seetiings.rst and other documentation fiile update is part of a separate commit

benoitc (changes requested):
@pajod can you make the change none optionnnal and link it to major 23.0 coming ?

gunicorn/config.py
benoitc: the spec says

> Field values containing CR, LF, or NUL characters are invalid and dangerous, due to the varying ways that implementations might parse and interpret those characters; a recipient of CR, LF, or NUL within a field value MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message. Field values containing other CTL characters are also invalid; however, recipients MAY retain such characters **for the sake of robustness when they appear within a safe context (e.g., an application-specific quoted string that will not be processed by any downstream HTTP parser).**


So you're right there to keep that extra option.

Issue #3235, opened 2024-07-04
Gunicorn accepts requests with empty URIs


kenballus:
The ABNF rule that defines request URIs (`request-target`) doesn't match the empty string. Thus, requests with empty URIs don't match the grammar, and should be rejected.

Currently, Gunicorn doesn't do this. You can see this for yourself by sending the following request to an instance of Gunicorn:
```
GET  HTTP/1.1\r\n
Host: whatever\r\n
\r\n
```
(Notice the two spaces between the method and version.)

benoitc:
i have edited this issue. There is nonissue to list all others implementations when you post an issue. Just describing the issue and linking to the spec is enough.

PR #3255, opened 2024-07-31
refuse empty request-target in HTTP request

pajod:
The shortest origin-form is a single slash, *not* empty. So refuse request lines that have two consecutive spaces where the URI should be. There should be *something* there.

Python [stdlib explicitly tells us](https://docs.python.org/3/library/urllib.parse.html#url-parsing-security ) it will not perform validation - and neither does this patch! This patch merely closes the one trivial hole that is of special interest, as it might confuse setups involving proxies and pipe-lining.

* Fixes: #3235

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -426,6 +426,17 @@ class Request(Message):
         # URI
         self.uri = bits[1]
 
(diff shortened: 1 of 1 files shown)


benoitc:
LFTM. thanks

Issue #3234, opened 2024-07-04
Gunicorn ignores the `Content-Length` header when the connection is half-closed.


kenballus:
# Description of the bug
From RFC 9112:
> If a valid Content-Length header field is present without Transfer-Encoding, its decimal value defines the expected message body length in octets. If the sender closes the connection or the recipient times out before the indicated number of octets are received, the recipient MUST consider the message to be incomplete and close the connection.

Gunicorn does not enforce this rule. When it receives a request, and the sender half-closes the connection, Gunicorn responds regardless of whether the request's body has been fully received.

# To reproduce
1. Start a Gunicorn-based HTTP server that echos the message body. (e.g., [something like this](https://raw.githubusercontent.com/narfindustries/http-garden/main/images/gunicorn/server.py))
2. Send it a request with an incomplete message body, followed by half-closing the socket, and observe that it still responds:
```sh
printf 'GET / HTTP/1.1\r\nHost: a\r\nContent-Length: 10\r\n\r\nA' | nc localhost 80
```
```
HTTP/1.1 200 OK
Server: gunicorn
Date: Thu, 04 Jul 2024 16:21:12 GMT
Connection: keep-alive
Content-type: application/json
Content-Length: 133

{"headers":[["SE9TVA==","YQ=="],["Q09OVEVOVF9MRU5HVEg=","MTA="]],"body":"QQ==","version":"SFRUUC8xLjE=","uri":"Lw==","method":"R0VU"}
```
5. Decode the response to see that it accepted the incomplete message body:
```sh
printf '{"headers":[["SE9TVA==","YQ=="],["Q09OVEVOVF9MRU5HVEg=","MTA="]],"body":"QQ==","version":"SFRUUC8xLjE=","uri":"Lw==","method":"R0VU"}' \
    | jq '.["body"]' \
    | xargs echo \
    | base64 -d \
    | xxd
```
```
00000000: 41                                       A
```

Gunicorn version: 22.0.0
Python version: 3.11.9

benoitc:
Body is streamed to the application. There is no reason to pass even incomplete header to the appplication. Otherwise that would force the gateway to buffer the whole body which is properly inneficient. The application beg-hind must take care of an incomplete body as a result.  What do I miss there?

kenballus:
> Body is streamed to the application. There is no reason to pass even incomplete header to the appplication.

I don't know what you're trying to say here.

> Otherwise that would force the gateway to buffer the whole body which is properly inneficient.

[...]

PR #3256, opened 2024-07-31
detect incomplete body before reading next request

pajod:
**DO NOT MERGE: ** The attached patch only checks that we have read the entire expected body after processing is done. It merely assist in testing and demonstrating that there is a problem, it does not prevent replying to truncated input yet.

I wonder if there is a way to approach #3234 with just one length check. We have one NoMoreData now, but it appears we want two variants - a) done reading, and b) unable to read more.

### DIFF
--- a/gunicorn/http/parser.py
+++ b/gunicorn/http/parser.py
@@ -3,6 +3,7 @@
 # This file is part of gunicorn released under the MIT license.
 # See the NOTICE for more information.
 
+from gunicorn.http.errors import IncompleteBody
 from gunicorn.http.message import Request
 from gunicorn.http.unreader import SocketUnreader, IterUnreader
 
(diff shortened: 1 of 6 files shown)


benoitc:
I don't understand the point there. We don't crash in the midldle of the appliication processing. How is the application supposed to handle such error there?

benoitc:
> > We don't crash in the middle of the application processing
> 
> We do raise `InvalidHeaderName` when we see something funny in chunked trailer fields. To me that feels like sufficient precedent for demanding that the application (and, where failing that, Gunicorn) can reasonably deal with the result of invalid or truncated input _up to and including processing even the very last octet of arbitrary-length input_.

fair enough :) please go ahead then :) We shoudl probably document this somewhere also

gunicorn/http/parser.py
```diff
@@ -27,15 +28,17 @@ def __iter__(self):
...
-        if self.mesg and self.mesg.should_close():
-            raise StopIteration()
-
         # Discard any unread body of the previous message
         if self.mesg:
             data = self.mesg.body.read(8192)
             while data:
                 data = self.mesg.body.read(8192)
+            if not self.mesg.body.finished:
+                raise IncompleteBody()
```
benoitc, line 37: maybe adding a reason there?

benoitc:
Closing this draft PR due to incomplete state. Feel free to reopen or create a new PR when the implementation is ready. Thank you!

PR #3258, opened 2024-07-31
gracefully handle chunked encoding missing size

pajod:
Integer conversion from zero-length strings is invalid: `ValueError: invalid literal for int() with base 10: ''`

Treat it the same as invalid characters where size should be.


benoitc:
thanks!

Issue #3125, opened 2023-12-28
Consistent exception handling for 4XX and 5XX errors


pajod:
I find it odd how different ParseException (or any error outside the application, really) is handled in different ways depending on configuration and specific error. I think this can be simplified by pushing the status, reason and message up and treating all exceptions in one code block.

They are all alike, really. They could subclass something that gives the basic structure of error messages associated with certain HTTP status codes and ASCII reason summaries e.g.
* https://github.com/pajod/gunicorn/blob/e9f0fa8/gunicorn/workers/base.py#L220
* https://github.com/pajod/gunicorn/blob/e9f0fa8/gunicorn/http/errors.py#L12

And the fact that `--reload` has an *implicit* side-effect that cannot be independently configured is also a bit surprising and inflexible. One idea I have is to introduce a new option, which merely *defaults* to automatically selecting the error-handler, but can be overridden if so desired.

I was thinking of a `--on-fatal=[world-readable|refuse|quiet|guess]` option, where only `guess` exposes potentially-meant-private code depending on whether it believes this is a non-public non-production environment e.g.

* now submitted in [PR #3360 Patch 5](https://github.com/benoitc/gunicorn/pull/3360/commits/dda47750afa563cb62401508a15573f9f0a87cd5#diff-96ece661a9027be952f6f1b9f2da9266707444ae6d4e7470fcc35d7f2041c633R148-R169)
* (previous draft was: [e9f0fa8/gunicorn/config.py#L2403](https://github.com/pajod/gunicorn/blob/e9f0fa8/gunicorn/config.py#L2403) [e9f0fa8/gunicorn/workers/base.py#L142](https://github.com/pajod/gunicorn/blob/e9f0fa8/gunicorn/workers/base.py#L142))

But maybe instead of 3-4 fixed choices, it should be a hook that can do anything (make_fail_app or otherwise). I am not certain about how that hook would work and be documented, to cover those 3-4 cases (and possibly more), and still not produce unexpected resources if that hook in turn triggers another exception.

Ideas?

tilgovi:
I've also considered that maybe we should have a `--dev` option. This would cause `--reload` to default on, `--reuse-port` to default off (and we make it on by default otherwise), and maybe there are other options. Sending tracebacks to the browser in development mode is convenient.

PR #3264, opened 2024-08-11
assert only, no change intended: never reflect invalid input

pajod:
This does not fix any bug, nor does it add any feature.

But if my less obviously correct patch for #3125 that deletes all the duplicate branches in error handling is flawed, then adding this assertion should reduce the impact of bugs from somewhat worrying "reflected injection" to much less worrying "crashes with proper traceback".

(re: pylint: No, those are not `superfluous-parens`. (Using the binding precedence of `not`) `is not` producing readable code. I recommend disabling that pylint warning)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -17,7 +17,7 @@ disable=
     broad-except,
     duplicate-bases,
     duplicate-code,
-    eval-used,
+    superfluous-parens,
     fixme,
     import-error,
     import-outside-toplevel,


benoitc:
Closing this PR as it adds assertions only with no functional change, and has been open for over a year. Thank you for the contribution.

pajod:
> no functional change

That was the point.

This thing:

https://github.com/benoitc/gunicorn/blob/e75c353/gunicorn/workers/base.py#L222-L257

should be refactored. **But ideally it would not be changed in one giant commit.**
Because right now it is trivial to introduce security bugs simply by failing to consider the interaction between user input in `Exception.__str__()` methods and the unscoped `mesg`, `reason` and `req` variables.

Issue #3267, opened 2024-08-12
sdist is missing `tox.ini`


mtelka:
The sdist package at PyPI is missing `tox.ini`.  Please add the missing file to sdist to make downstream testing easier.  Thank you.

PR #3269, opened 2024-08-12
packaging: include tox.ini in sdist

pajod:
* Fixes: #3267 (not immediately, the change merely requests inclusion in *future* source distributions)

Argument from authority: See what setuptools does & documents: https://github.com/pypa/setuptools/blob/1d2fd37368fc2381e21cd5466aeb222fc11db2db/MANIFEST.in

Confirmation:
```shell
$ python -m build
$ tar -tf dist/gunicorn-*.tar.gz | grep tox
gunicorn-23.0.0/tox.ini
```
note: if ever using setuptools-scm: .gitattributes / MANIFEST.in precedence could undo this change

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ include README.rst
 include THANKS
 include requirements_dev.txt
 include requirements_test.txt
+include tox.ini
 recursive-include tests *
 recursive-include examples *
 recursive-include docs *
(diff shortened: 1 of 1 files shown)


jayaddison:
@pajod why should `tox.ini` be included in a distributable?  Naively I would assume that a distributable release includes only the application code -- that is, the code necessary to run a project on a production system.

jayaddison:
Sorry!  This is an `sdist` -- a _source_ distributable.  And those can include test code.  In that case it seems more reasonable to include `tox.ini`.

PR #3273, opened 2024-08-13
style: re-verbosify HTTP commentary

pajod:
The HTTP parser code is difficult to understand, even with the spec on a second screen. Add (back) some commentary regarding the places where we see a request that is invalid for rather minuscule issues, yet we *intentionally* refuse supporting it.


benoitc:
Closing this PR as it's a style-only change (comment re-verbosification) that has been open for over a year. Thank you for the contribution.

PR #3280, opened 2024-08-17
Python 3 style cleanup (part 2)

pajod:
Cleanup, follow-up to #3189 - There is a change of config default in here, but for our workers that only impacts who sees the warning about the *ignored* setting. Rest is intended to be no-op.

* the eventlet warning is coped from https://github.com/eventlet/eventlet/blob/master/README.rst

* Unfinished consideration: `InvalidChunkSize` (and most of the meaning of `NoMoreData` and `ChunkMissingTerminator`, though that needs to be split first) subclassing IOError is just wrong, replacing with OSError does not fix that. Input *reading* worked perfectly fine, its the *content* of that input that our parser found issues with. Maybe some of it could re-use the built-in EOFError, but only while avoiding confusing unexpected SSL tunnel shutdown with previously announced end of input.

* Suggested merge order: merge *last*

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -681,7 +681,8 @@ class WorkerClass(Setting):
         * ``gthread``
         * ``eventlet`` - Requires eventlet >= 0.24.1 (or install it via
           ``pip install gunicorn[eventlet]``).
-          Usages in new projects are discouraged.
+          Usages in new projects `are discouraged
+          <https://github.com/eventlet/eventlet/blob/master/README.rst>`_.
         * ``gevent``   - Requires gevent >= 1.4 (or install it via
           ``pip install gunicorn[gevent]``)
         * ``tornado``  - Requires tornado >= 0.2 (or install it via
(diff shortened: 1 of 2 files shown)


tests/requests/invalid/007.py
```diff
@@ -1,2 +1,3 @@
 from gunicorn.http.errors import LimitRequestHeaders
+
```
benoitc, line 2: why bothering adding a line there?

gunicorn/config.py
```diff
@@ -678,14 +678,14 @@ class WorkerClass(Setting):
         A string referring to one of the following bundled classes:
 
         * ``sync``
+        * ``gthread``
         * ``eventlet`` - Requires eventlet >= 0.24.1 (or install it via
-          ``pip install gunicorn[eventlet]``)
+          ``pip install gunicorn[eventlet]``).
+          Usages in new projects are discouraged.
```
benoitc, line 684: why?

benoitc:
Closing this PR as it's a pure style cleanup that has been open for over a year without review. The codebase has evolved and this would need significant rebasing. Thank you for the contribution, but we're focusing on functional changes at this time.

pajod:
> this would need significant rebasing

@benoitc Actually does not. I just rebased to confirm: https://github.com/benoitc/gunicorn/compare/master...pajod:gunicorn:style-py3-part2?expand=1

Issue #3281, opened 2024-08-19
module invocation breaks USR2 upgrading, as modified argv results in modified path


pajod:
Mixing sys.executable and *parsed* argv produces unintentional sys.path changes on re-exec:

```
# python -m gunicorn my:app
Booting worker with pid: 201117
[..]
# kill -USR2 $(pgrep --oldest -f "gunicorn")
Handling signal: usr2
[..]
  File "/venv/lib/python3.11/site-packages/werkzeug/serving.py", line 28, in <module>
    from http.server import BaseHTTPRequestHandler
ModuleNotFoundError: No module named 'http.server'
```

* this is *not* an issue when you run `gunicorn my:app` (with `argv[0] == '[..]/gunicorn'`)
* this is an issue when `argv[0]` is significant, such as `python3 -m gunicorn my:app`
  * then parsed sys.argv may point to `[..]/__main__.py`
  * then relaunching using parsed argv will result in `sys.path` listing directory containing `__main__.py`
  * then wsgi apps using `werkzeug` (such as `flask`) will crash because we clobbered the `http` name
  * which we never meant to do anyway
  * and crucially, *did not do* .. until we launched new master after USR2

https://github.com/benoitc/gunicorn/blob/903792f152af6a27033d458020923cb2bcb11459/gunicorn/arbiter.py#L68-L76

(Side note: The `p` is wrong and scaring static analyzers, but harmless in cPython. I suggest cleaning it up while we are at it.)

https://github.com/benoitc/gunicorn/blob/903792f152af6a27033d458020923cb2bcb11459/gunicorn/arbiter.py#L430-L431

I *think* we just did not have the option to [use the original arguments](https://docs.python.org/3/library/sys.html#sys.orig_argv) (added in 3.10):

```diff
diff a/gunicorn/arbiter.py b/gunicorn/arbiter.py
-    os.execve(sys.executable, [sys.executable, *sys.argv[:]], environ)
+    os.execve(sys.executable, sys.orig_argv, environ)
```

Objections against just doing just that, on Python versions where we can? Even if we cannot do that for some reason, I would still add a *warning* when launching with *clearly* unintended `sys.path` - it is a common source of confusion.

* see also https://docs.python.org/3/using/cmdline.html#cmdoption-P

benoitc:
what does mean `Mixing sys.executable and parsed argv produces unintentional sys.path changes on re-exec:` ? Can you provide an example?

Issue #2104, opened 2019-09-04
Bind to unix socket *and* IP: gunicorn only binds to socket.


rafalmp:
I am trying to get gunicorn to listen simultaneously on unix socket and IP address:
```
$ gunicorn --bind unix:/run/testapp/socket --bind 10.0.0.2:8000 test:app
```
Only the unix socket works - gunicorn does not listen on the IP.

tilgovi:
What version of Gunicorn, please?

rafalmp:
Sorry, should have mentioned that. 19.9.0.

benoitc:
I tried from the example folder the following command line : 

```python
→ gunicorn --bind unix:/tmp/gunicorn.sock --bind localhost:8080 test:app
[2019-09-05 10:46:56 +0200] [15865] [INFO] Starting gunicorn 19.9.0
[2019-09-05 10:46:56 +0200] [15865] [INFO] Listening at: unix:/tmp/gunicorn.sock,http://127.0.0.1:8080 (15865)
[2019-09-05 10:46:56 +0200] [15865] [INFO] Using worker: sync
[2019-09-05 10:46:56 +0200] [15868] [INFO] Booting worker with pid: 15868
```

I can query `http://localhost:8080 ` .  There is a bug though since this binding should be displayed in the output.

rafalmp:
Interesting, it does work from CLI for me too, but when I try to run gunicorn from a systemd service file, it doesn't. My unit file is as follows (based on the official documentation):
```
[Unit]
Description=myapp gunicorn daemon
Requires=myapp.socket
After=network.target

[Service]
PIDFile=/run/myapp/pid
User=myappuser
Group=myappuser
RuntimeDirectory=myapp
WorkingDirectory=/home/myappuser/myapp
ExecStart=/home/myappuser/.local/share/virtualenvs/myapp-OsxhwLj6/bin/gunicorn \
    --pid /run/myapp/pid --bind unix:/run/myapp/socket --bind 127.0.0.1:8080 \
    --worker-class aiohttp.GunicornWebWorker main:app
ExecReload=/bin/kill -s HUP $MAINPID
ExecStop=/bin/kill -s TERM $MAINPID
PrivateTmp=true

[Install]
WantedBy=multi-user.target
```
journalctl output:
```
Sep 05 13:59:38 myappuser gunicorn[32545]: [2019-09-05 13:59:38 +0200] [32545] [INFO] Starting gunicorn 19.9.0
[...]

Issue #1948, opened 2019-01-11
more status to systemd


benoitc:
followup to #1897, we can provide the arbiter [status to systemd](https://www.freedesktop.org/software/systemd/man/sd_notify.html) via sd_notiy. I propose for a start to add the following status:

- [ ] `RELOADING`: when the arbiter receive an `HUP` signal, code happen there it can be done in the [handle_hup](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L247) function
- [ ] `STOPPING`: when the arbiter receive a `TERM` signal.  it can be done in the [stop function](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L372)
- [ ] `MAINPID`: when the arbiter receive a USR2 signal. This one is tricky as the user can choose to not keep the new arbiter. So it needs to be set when the server is finally killed using `QUIT` maybe in the [maybe_promote_master function](https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L315)

Thoughts?

benoitc:
cc @gdamjan @tilgovi @berkerpeksag

gdamjan:
It's a good idea. I didn't dare to try implementing it, since I'm not familiar with the gunicorn states.

benoitc:
no activity since awhile. closing feel free to create a new ticket if needed.

PR #3285, opened 2024-08-23
Retain argv + fds on re-exec, fix USR2 under systemd by notifying new PID

pajod:
This PR stops mangling sys.argv and socket fd numbers on re-exec, thus permitting much improved integration with systemd.

* While not meant to break anything before that, this patch is only effective in cPython >= 3.10.
  * Python docs explicitly say that relaunching with `-m` switch is the [recommended procedure](https://docs.python.org/3/library/subprocess.html#subprocess.Popen), so that *should* work
```diff
diff --git a/gunicorn/arbiter.py b/gunicorn/arbiter.py
-        args = sys.argv[:]
-        args.insert(0, sys.executable)
+        if sys.version_info < (3, 10):
+            args = sys.argv[:]
+            args.insert(0, sys.executable)
+        else:
+            args = sys.orig_argv[:]
```
* New feature: send timestamp to properly synchronize reload cycles using `Type=notify-reload` (systemd `>= v253`)

* Fixes: #3281
* Fixes: #1948
* Fixes: #2104
  * insofar as the surprising behavior is *documented* at least
* Fixes: #2856
  * insofar as `systemctl -s USR2 gunicorn.service` now automatically issues graceful quit
  * `systemctl reload gunicorn.service` and `systemctl restart gunicorn.service` unchanged in .service example
* Improves systemd-specific case of: #3050
* Improves sock fd leak pointed out in: #3339

* Follow-Up plans not included here:
  * the choice between announcing new MAINPID as early or as late as possible 
    * ideally, we would not trigger the `systemd[1]: gunicorn.service: Supervising process 1337 which is not our child. We'll most likely not notice when it exits.` warning, at all
  * document difference between behavior of `< 3.10` and Python `>= 3.10`
  * could be improved with `BARRIER=1` so reloading/re-execing in quick success is reliable as well
  * could be improved with `EXTEND_TIMEOUT_USEC=` so automatic timeouts are reliable as well

### DIFF
--- a/docs/source/deploy.rst
+++ b/docs/source/deploy.rst
@@ -247,7 +247,8 @@ to the newly created unix socket:
 
     [Service]
     # gunicorn can let systemd know when it is ready
-    Type=notify
+    # if systemd versions >= v253, otherwise use 'Type=notify'
+    Type=notify-reload
     NotifyAccess=main
     # the specific user that our service will run as
     User=someuser
@@ -257,7 +258,8 @@ to the newly created unix socket:
     RuntimeDirectory=gunicorn
     WorkingDirectory=/home/someuser/applicationroot
     ExecStart=/usr/bin/gunicorn applicationname.wsgi
-    ExecReload=/bin/kill -s HUP $MAINPID
+    # if 'Type=notify' instead of 'Type=notify-reload' (depending on systemd version)
+    # ExecReload=/bin/kill -s HUP $MAINPID
     KillMode=mixed
     TimeoutStopSec=5
     PrivateTmp=true


benoitc:
what , does thsi patch exactly in simple word and how do you guarantee the compatibility with current systems not running systemd?

pajod:
This PR is a superset of #3419 and I will rebase and fix it once that one is merged.

> what , does thsi patch exactly in simple word

Use fd numbers and argv as-is, and keep systemd informed about status changes.

> how do you guarantee the compatibility with current systems not running systemd

The status changes are effectively no-op outside of systemd, so only 3 changes remain to be judged in that regard:

 1. using original argv is the same except for one special case -- one probably nobody used, as suggested by me being 1st to complain about the crash
 2. not duplicating file descriptors on re-exec and the `fd://` setting -- getting exactly the same thing is the *point* of duplication. If that operation changed things, it would not really be a "duplicate"
 3. extra warning -- only emitted on very clearly unintended configuration. also, its just one extra line logged per master startup, should not cause compat issues.

I deliberately did not guard these 3 changes behind a `if systemd:` branch, because they resolve issues that are merely unlikely to be encountered outside of systemd, not specific to systemd.

bvallant:
Would be interesting to get this into the next version. Currently trying to setup Reloading with systemd and `SIGUSR2`, but the main problem really seems to be that there seems no other descent way of telling systemd about the new `MAINPID` besides having Gunicorn itself use `sd_notify`.

PR #3303, opened 2024-09-30
Fix request body handling in gthread worker

### LINKED ISSUE #3301: gthread worker will block
The gthread worker will block and try to read a request that wasn't sent on the wire in a specific scenario:

- A POST request is sent with a total size (request header and body) greater than 8192 bytes.
- The application does not read the request body
- The connection is kept alive

The worker will then be blocked on that connection until it is closed, and is not able to accept any requests.

Reproduction:

Create myapp.py:

```python3
    data = b"Hello, World!\n"
    start_response("200 OK", [
    ])
    return iter([data])
```

Run with `gunicorn myapp:app --keep-alive 120 --worker-class gthread`

Then create another Python file to open a connection and make a POST request `httppost.py`

```python3
from time import sleep
import sys, socket

host, port, numbytes = sys.argv[1], int(sys.argv[2]), int(sys.argv[3])

with socket.create_connection((host, port)) as sock:
    get = b"POST /test HTTP/1.1\r\nHost: " + host.encode("ascii") + b"\r\nContent-Length: " + bytes(str(numbytes), "ascii") + b"\r\n\r\n" + (b"f" * numbytes)
    sock.sendall(get)
    while True:
            data = sock.recv(1024)
            print(data)
```

Execute with `python3 httppost.py 127.0.0.1 8000 9000`

Then try to make another request on a different connection such as with `curl "http://127.0.0.1:8000/test"` and notice that you do not get a response. Ctrl+C the httppost.py program and the curl will complete as expected.

oallenj:
Fixes #3301.

The problem is that if the request body is not read and the gthread worker finishes handling a request with a large request body, it will see that the socket is still readable and call on_client_socket_readable. However, the readable bytes are just the rest of the request body. The part of the code annotated "Discard any unread body of the previous message" will discard that request body, but gthread will then proceed to subsequently try and read a full HTTP request, which may have not been sent. This change will discard the request body after handling a request, and before on_client_socket_readable can be called so this doesn't happen.

I'm not sure how to make a test for this because it requires multiple connections and doesn't seem to fit in with the rest of gunicorn's test suite.


somerobmitchell:
@benoitc Do you have any update on the review of this PR?

Issue is here: https://github.com/benoitc/gunicorn/issues/3301

PR #3311, opened 2024-10-04
fix: change log level of sigterm signal received from error to warning

pauljeannot:
Hi,

This PR addresses the issue reported here: https://github.com/benoitc/gunicorn/issues/3050.

I have reverted the log level back to warning, as it was previously. This prevents the log from being sent to stderr and cluttering the monitoring, as it’s not truly an error.

Thansk!

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -550,7 +550,7 @@ class Arbiter:
                         # Additional hint for SIGKILL
                         if status == signal.SIGKILL:
                             msg += " Perhaps out of memory?"
-                        self.log.error(msg)
+                        self.log.warning(msg)
 
                     worker = self.WORKERS.pop(wpid, None)
                     if not worker:


drichardson:
Can this get merged in @benoitc? This error message causes a lot of error log false positive noise during deployments where SIGTERMs are common.

panterz:
Any updates on this? Thanks

shetty-777:
Please merge this...
Is that what should be done? In gunicorn 23.0.0, the arbiter.py has `error` and not `warning`

ryuichi1208:
@pajod 

> Where this generally happens, it is caused by explicit administrative action. Gunicorn can and should be shutdown properly - by signalling the master, without triggering an error.

Isn't it possible that what's happening now is that after sending SIGTERM to the master, the SIGTERM is propagated to the child process, resulting in an error? Is it possible to determine whether sending SIGTERM directly to the child process is a problem?

pajod:
@ryuichi1208

> Isn't it possible that what's happening now is that after sending SIGTERM to the master, the SIGTERM is propagated to the child process, resulting in an error?

You are thinking of just "receiving SIGTERM", but that is not what goes to the `WIFSGINALED(waitpid(..))` branch. That one  is triggered by unexpectedly learning that the child process received a signal **and *failed* to catch it**. Which is not what should happen (and generally: what does happen) on ordinary shutdown.

> Is it possible to determine whether sending SIGTERM directly to the child process is a problem?

For the master process, that distinction is easy. The answer is yes. It is a problem, *by default*, if a live worker is terminated by a signal it could and should have handled.

ryuichi1208:
@pajod 

What about cases where a child process reports to the parent process that it has terminated due to a signal it received? Even if processes like uvicorn catch the signal with a signal handler, they still ultimately send the signal to themselves. Is that also an error?

This is only an issue with SIGTERM and the like, and may not be something gunicorn should be concerned with.

PR #3319, opened 2024-10-29
perf(gthread): reduce unnecessary polling

ankush:
gthread calls `epoll_wait` (and 2 other syscalls) every second because it
specifies timeout to be 1 second.

```
λ sudo strace -p `pgrep -f "gunicorn: worker" | head -n1`
strace: Process 30815 attached
epoll_wait(7, [], 1, 666)               = 0
getppid()                               = 30800
utimensat(6, NULL, [{tv_sec=3157, tv_nsec=198136276} /* 1970-01-01T06:22:37.198136276+0530 */, {tv_sec=3157, tv_nsec=198136276} /* 1970-01-01T06:22:37.198136276+0530 */], 0) = 0
epoll_wait(7, [], 1, 1000)              = 0
getppid()                               = 30800
utimensat(6, NULL, [{tv_sec=3158, tv_nsec=204192934} /* 1970-01-01T06:22:38.204192934+0530 */, {tv_sec=3158, tv_nsec=204192934} /* 1970-01-01T06:22:38.204192934+0530 */], 0) = 0
epoll_wait(7, [], 1, 1000)              = 0
getppid()                               = 30800
utimensat(6, NULL, [{tv_sec=3159, tv_nsec=210145196} /* 1970-01-01T06:22:39.210145196+0530 */, {tv_sec=3159, tv_nsec=210145196} /* 1970-01-01T06:22:39.210145196+0530 */], 0) = 0
epoll_wait(7, [], 1, 1000)              = 0
getppid()                               = 30800
utimensat(6, NULL, [{tv_sec=3160, tv_nsec=215517372} /* 1970-01-01T06:22:40.215517372+0530 */, {tv_sec=3160, tv_nsec=215517372} /* 1970-01-01T06:22:40.215517372+0530 */], 0) = 0
epoll_wait(7, ^Cstrace: Process 30815 detached
 <detached ...>
 ```

Timing out every second wakes up the process and loads it on CPU even
if there is nothing to service.

This can be detrimental when you have `total workers >> total cores` and
a multi-tenant setup where most tenants might be sitting idle. (but not
"idle enough" because of 1s polling timeout)

This can possibly keep a completed future in the queue until the next 
request arrives, but I don't see any obvious problem with it except a few 
bytes of extra memory usage? I could be wrong here, please check this.

fixes https://github.com/benoitc/gunicorn/issues/3317 (more details on issue)

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -206,8 +206,13 @@ class ThreadWorker(base.Worker):
 
             # can we accept more connections?
             if self.nr_conns < self.worker_connections:
+                # Block for new events until worker times out or keepalive timeout.
+                select_timeout = self.timeout or 1.0
+                if self._keep:
+                    select_timeout = min(select_timeout, self.cfg.keepalive)
+
                 # wait for an event
-                events = self.poller.select(self.timeout)
+                events = self.poller.select(select_timeout)
                 for key, _ in events:
                     callback = key.data
                     callback(key.fileobj)
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -207,7 +207,7 @@ class ThreadWorker(base.Worker):
             # can we accept more connections?
             if self.nr_conns < self.worker_connections:
                 # wait for an event
-                events = self.poller.select(1.0)
+                events = self.poller.select(self.timeout)
                 for key, _ in events:
                     callback = key.data
                     callback(key.fileobj)
--- a/THANKS
+++ b/THANKS
@@ -20,6 +20,7 @@ Andreas Stührk <andy-python@hammerhartes.de>
 Andrew Burdo <zeezooz@gmail.com>
 Andrew Svetlov <andrew.svetlov@gmail.com>
 Anil V <avaitla16@gmail.com>
+Ankush Menat <ankushmenat@gmail.com>
 Antoine Girard <antoine.girard.dev@gmail.com>
 Anton Vlasenko <antares.spica@gmail.com>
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by ankush** [comment]
@pajod timeout is divided by 2 when the worker is created. So self.timeout is half of request timeout.


I'll look into 2nd part of your comment in some time.

**Comment by ankush** [comment]
@pajod I think best course of action is to keep timeout as `min(req_timeout/2, keep_alive)`, this addresses your concern and mine. 

1. I can set keep-alive to a much higher value. The setup I have like most others, run gunicorn behind a reverse proxy.
2. TTFB is minimally affected by this change.
3. `keep-alive` keeps working as advertised... with worst-case deviation of up to 2x keep-alive value only occurring when workers go idle. We can document this. E.g. a socket had almost reached keep-alive timeout but we just blocked on epoll_wait. So, nearly 2x keep-alive time will be required in idle conditions.

File: gunicorn/workers/gthread.py [hunk 1]
```diff
@@ -206,8 +206,13 @@ def run(self):
 
             # can we accept more connections?
             if self.nr_conns < self.worker_connections:
+                # Block for new events until worker times out or keepalive timeout.
+                select_timeout = self.timeout or 1.0
+                if self._keep:
```
ankush, line 211: I can hoist this change out of the while loop, but then folks running the default config won't see any noticeable difference. 

Keeping it inside the loop allows making timeout adaptive to the presence of current keepalived sockets.

ankush:
@benoitc kind reminder :)

oallenj:
Could you use timer file descriptors instead? Then the keep alive timeout could be enforced accurately and polling would be unnecessary.

PR #3327, opened 2024-11-24
fix: problematic parsing leniency in parsing chunk extensions

JeppW:
Stop allowing newline characters in chunk extensions. This can cause request smuggling issues with some reverse proxies. 

Reference: https://grenfeldt.dev/2021/10/08/gunicorn-20.1.0-public-disclosure-of-request-smuggling/

### DIFF
--- a/gunicorn/http/body.py
+++ b/gunicorn/http/body.py
@@ -6,7 +6,7 @@ import io
 import sys
 
 from gunicorn.http.errors import (NoMoreData, ChunkMissingTerminator,
-                                  InvalidChunkSize)
+                                  InvalidChunkSize, InvalidChunkExtension)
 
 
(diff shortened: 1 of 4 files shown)


JeppW:
AFAIK only the newline has ever been shown to be exploitable in practice. Perhaps we should disallow \r as well, it seems plausible that a proxy could misinterpret \rX as a line terminator. Other than that, I think further validation would just be unnecessary overhead.

JeppW:
I added the \r as we agreed.

benoitc:
On hwich specification is based this change?

pajod:
(only) CR rejected since https://github.com/benoitc/gunicorn/commit/bdb2ebd5a4913fff1e92800f3763e4a879526d3e

Issue #3330, opened 2024-11-28
Preserve worker exit codes in worker class


tback:
It'd be really great if the exit code of workers could be preserved in the worker class.
That way we could log unusual ones.
My reason for asking for this is that I'm trying to find processes that have be targeted by the oom killer. Those receive exit code 137. 
Then I could define a `child_exit` funktion that log what happened:

```python
def child_exit(_, worker):
        worker.log.info(f"Child Exit: Worker with PID {worker.pid} exited with code {worker.exit_code}.")
```

tback:
I just noticed this should be logged already. Yet I don't see exit code 137 while I'm quite sure it happens. Will update (and hopefully close) in about two hours https://github.com/benoitc/gunicorn/blob/bacbf8aa5152b94e44aa5d2a94aeaf0318a85248/gunicorn/arbiter.py#L528

tback:
Conclusion: Exit code 137 is not magically produced in case of an OOMKill. It's neither a clear indicator for it. Exit code 137 is the result of adding the number of the signal (`9` for `SIGKILL`) and 128 (the base number for signal exits).

PR #3331, opened 2024-11-28
Use library function to get exitcode from status

tback:
Shifting by 8 bits will hide the correct exit code in some cases.

On linux sending `SIGKILL` will produce the log entry.

```
Worker (pid:22) was sent SIGKILL! Perhaps out of memory
```

The logline printing the exitcode is not triggered because shifting the status will zero it.

The patched version using `os.waitstatus_to_exitcode` produces.
```
Worker (pid:7) exited with code -9
Worker (pid:7) was sent SIGKILL! Perhaps out of memory?
```

Yes, this is a minute detail. I discovered it when investigating #3330.


tback:
thanks, closing in favor of #3148

PR #3340, opened 2025-01-19
Prevent gthread connection reset on max-requests restart

aszubarev:
Fixes: #3038

***

Using #3039 as an example, I suggest adding a setting that will optionally enable the processing mode #2918.

I think the original behavior (before #2918) should be the default, as it matches the expected behavior. See [comment](https://github.com/benoitc/gunicorn/issues/2917#issuecomment-1407638555).

And for users who use Gunicorn without a reverse proxy, make it possible to enable the current connection processing mode optionally. This way they can avoid blocking workers like here #2917

### DIFF
--- a/gunicorn/config.py
+++ b/gunicorn/config.py
@@ -126,6 +126,10 @@ class Config:
         return worker_class
 
     @property
+    def worker_connections_enqueue_async(self):
+        return self.settings['worker_connections_enqueue_async'].get()
+
+    @property
     def address(self):
         s = self.settings['bind'].get()
         return [util.parse_address(util.bytes_to_str(bind)) for bind in s]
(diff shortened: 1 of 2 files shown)


benoitc:
Closing this PR as the underlying issue has been addressed by the gthread worker refactoring in PR #3440.

The refactored implementation now always submits connections directly to the thread pool after accept (see `accept()` method), rather than registering with the poller and waiting for data first. This eliminates the connection reset issue on max-requests restart because:

1. Connections are immediately being processed by worker threads
2. Graceful shutdown properly drains in-flight connections within the grace period
3. No connection state is held in the poller that could be lost on restart

Thank you for the PR and for identifying this issue!

PR #3341, opened 2025-01-31
Remove RuntimeDirectory from systemd service

vanschelven:
In ab81ae3b4d15 this was added with the following remark:

> Without this /run/gunicorn isn't created

However, in ba0d784960be the final reference to `/run/gunicorn/` (for `/run/gunicorn/pid`) was removed. I've checked like so:

find -type f | x grep run/\gunicorn\/


vanschelven:
I merged main into this branch;

```
find -type f | x grep run/\gunicorn/
```

still has no results. 

@benoitc feel free to merge :smile:

PR #3342, opened 2025-01-31
Drop Python 3.7 support (fixes Github CI)

vanschelven:
it was EoL since 2023-06-27 and since today prevents the CI from running


vanschelven:
https://github.com/benoitc/gunicorn/pull/3343#issuecomment-3369261726

it seems that Gunicorn is going a bit more radical with reducing the list of supported Python versions. Whatever the outcome of that discussion, the present PR is now outdated.

PR #3343, opened 2025-01-31
Drop python 3.8

vanschelven:
it was EoL since 2024-10-07

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -21,8 +21,7 @@ jobs:
          - macos-13
          # Not testing Windows, because tests need Unix-only fcntl, grp, pwd, etc.
         python-version:
-         # CPython <= 3.7 is EoL since 2023-06-27
-         - "3.7"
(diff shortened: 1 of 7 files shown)


vanschelven:
This is similar to, and a continuation of, #3342; by opening 2 PRs we'll be able to discuss the merits and timing of each independently.

darjeeling:
@vanschelven I think EOL of 3.9 is coming soon.
also CI is still failing. can we go with other PR?

vanschelven:
With 2 open PRs waiting on a decision for so long I'm not all to eager to open a third. Once this starts moving, I just might. Looking ahead at the moment of "Drop Python 3.9", the following commit (in my own project) may provide some relevant background info on popular distributions' Python versions: https://github.com/bugsink/bugsink/commit/12af5302efdd6c33f171ba9ba43efc420465f30b

benoitc:
Gunicorn is now supporting only N,N-1 releases of python. Closing it as longer needed then. Thanks for it anyway :)

PR #3346, opened 2025-02-12
Disable SO_REUSEPORT for unix sockets

e-nomem:
Because of the fix for [CVE-2024-57903](https://nvd.nist.gov/vuln/detail/CVE-2024-57903), the linux kernel started returning an error if an attempt is made to set the `SO_REUSEPORT` socket option on non-inet/inet6 sockets.

In gunicorn, if `reuse_port` is set to `True` and one of the bind addresses is a unix socket, this causes the arbiter to hang on newer kernels.

### DIFF
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -36,10 +36,13 @@ class BaseSocket:
     def __getattr__(self, name):
         return getattr(self.sock, name)
 
+    def supports_reuseport(self):
+        return hasattr(socket, 'SO_REUSEPORT')
+
     def set_options(self, sock, bound=False):
         sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
         if (self.conf.reuse_port
-                and hasattr(socket, 'SO_REUSEPORT')):  # pragma: no cover
+                and self.supports_reuseport()):  # pragma: no cover
             try:
                 sock.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEPORT, 1)
             except OSError as err:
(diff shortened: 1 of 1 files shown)


e-nomem:
> Thanks. What is the advantage of failing silently? The admin specifically requested it. We should let them know, no?

I guess, but doesn't that code already silently skip if the `SO_REUSEPORT` attribute is missing (I'm guessing if the kernel is older than v3.9 when the option was added...) even if it was explicitly requested in the config?

> If they specified a mix of UNIX and INET sockets on the command line / config, I imagine we can safely assume that was by mistake, [...]

I don't think that's a safe assumption. I discovered this issue after a kernel upgrade in a situation where the configuration specified binding to both `unix:/run/gunicorn/gunicorn.sock` and `127.0.0.1:8080` and I think that's valid.

> [...] and if they combined INET socket on cmdline/config yet passed some bound UNIX socket via systemd.. that is broken anyway, because the `Arbiter.start` won't pick up both anyways.

I can't speak to the systemd socket activation interactions. My application runs in a containerized environment without systemd.

> (Also: Could this patch break some edge case outside the intersection of `isinstance(sock.getsockname(), tuple)` and `sk_is_inet(sk)`?)

I have no idea if any other address family returns names that are tuples 😅 I think the only way to sneak them into the system is via systemd socket activation though because the gunicorn code will only open sockets in one of the 3 'normal' address families of INET, INET6, and UNIX.

> If there is an argument for silencing the discrepancy between requested config and applied settings, then the config help text should say so.

Agreed. Regardless of if this PR is merged, the docs should be updated. I guess it really comes down to deciding if attempting to bind to both a unix and inet socket at the same time is valid. If it is, we should silently skip setting the option (and maybe log the skip). If it is not valid, we should error out and fail. The latter seems like it would be a breaking change.

e-nomem:
Looks like cpython has also decided to silently skip setting reuseport instead of erroring out: https://github.com/python/cpython/pull/128933

PR #3355, opened 2025-03-11
Fix typo in the settings.rst

mondwan:
getgrnam is under grp module instead of pwd


mondwan:
Sure, let me take a look

PR #3357, opened 2025-03-13
Inherit Application instead of BaseApplication

mondwan:
Some of the settings are not implemented in BaseApplication But Application


benoitc:
will not happen. works as intended.

Issue #3212, opened 2024-05-16
gunicorn/util.py set_owner_process Mangling gid on RHEL8 + Google Cloud Platform (very large int group id)


jbisbee:
I'm using Python 3.9.18 on Red Hat Enterprise Linux 8.   My uid and gid values are crazy high under RHEL8 and Google Cloud Platform.  As a result, they are hitting a hard coded condition within gunicorn/util.py

151         # versions of python < 2.6.2 don't manage unsigned int for
152         # groups like on osx or fedora
153         gid = abs(gid) & 0x7FFFFFFF

Where my gid is being changed from 4133075255 to 1985591607 and then because they don't match, gid is attempted to be set and gunicorn dies because I'm not a priveledged user and I don't have permission to set gid (which is only firing off because my gid was unjustly molested 👎)

benoitc:
> top messing with uid/gid/groups altogether when there is no need to.

@pajod care to elaborate? We need to be abble to set user and group to detach worker process and handle permissions approriately. UNsure what you mean there

nicholas-johnson-techxcel:
Running this with Langflow is a nightmare. It should not be trying to change gid but it does, seemingly constantly, and it does not give up nor work properly.

benoitc:
the question is rather, why the user you launch can't assign its own gid. I don't reproduce suc issue in any env I use. if gid is mismatching then we should handle the case wit h oses not capable to follow a spec. Not removing a feature.

PR #3360, opened 2025-03-18
clearly point out the surprising behavior of --reload, --user, --group, --config (+ unlock a few previously impossible combinations)

pajod:
Resubmitting a subset of #3201 - [less mixing topics](https://github.com/benoitc/gunicorn/pull/3201#issuecomment-2707223540)

Retained:
* at least *document* the weird status quo of `--user` and `--group` settings
* permit linting `config.py` without tons of complaints about class methods not taking `self` arg
* permit `--reload` without publishing tracebacks (idea from [#3125](https://github.com/benoitc/gunicorn/issues/3125))
* permit `--reload-extra-files`  without `--reload` (separately in [#3271](https://github.com/benoitc/gunicorn/pull/3271))
  * a test that showcases how this works (similar to [#3210](https://github.com/benoitc/gunicorn/pull/3210) - will refactor if both accepted)
* permit `--config=/dev/null` (without warning, when execution of `gunicorn.conf.py` file is not intended)

Not submitted here:
* dropped: permit *not* using `--user` and `--group` without anything bad happening ([reason](https://github.com/benoitc/gunicorn/issues/3212#issuecomment-2282751357))
  * <del>this removes the not generally true assumption that `os.setuid(os.geteuid())` is no-op</del>
  * see https://sources.debian.org/src/gunicorn/23.0.0-1/debian/patches/0001-drop-supplemental-groups.patch/
  * see https://sources.debian.org/src/gunicorn/23.0.0-1/debian/patches/0004-Set-supplementary-groups-when-changing-uid.patch/
* no `*.pyi` type stubs
* no docs linting
* no pyproject.toml/pylintrc reformatting
* no longer includes https://github.com/benoitc/gunicorn/commit/7268a610996eee15e328fce3daf8bf8e748f1627
* no longer includes https://github.com/benoitc/gunicorn/pull/3264

Related; I shall fix merge conflicts once this or one of these is merged:
* https://github.com/benoitc/gunicorn/pull/3272
* https://github.com/benoitc/gunicorn/pull/3110

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -21,8 +21,7 @@ jobs:
          - macos-13
          # Not testing Windows, because tests need Unix-only fcntl, grp, pwd, etc.
         python-version:
-         # CPython <= 3.7 is EoL since 2023-06-27
-         - "3.7"
+         # CPython <= 3.8 is EoL since 2024-10-07 https://peps.python.org/pep-0569/
          - "3.8"
          - "3.9"
          - "3.10"


gunicorn/config.py
```diff
@@ -456,7 +456,7 @@ def _validate_callable(val):
 
 def validate_user(val):
     if val is None:
-        return os.geteuid()
```
benoitc, line 459: what does this change fix?

gunicorn/config.py
```diff
@@ -1171,14 +1183,21 @@ class Group(Setting):
     cli = ["-g", "--group"]
     meta = "GROUP"
     validator = validate_group
-    default = os.getegid()
```
benoitc, line 1174: again what does this change fix ? what is the error caused by this? We need to ensure the worker process run under the specified or running user.

benoitc (commented):
can you describe what the user change fix? I don't see any real reason to change a behavior in place since many year . can you expand?

Issue #3364, opened 2025-03-20
Gunicorn incorrectly strips certain characters from `transfer-coding`s


kenballus:
When Gunicorn receives a request with a `Transfer-Encoding: \x0b\x0c\x1c\x1d\x1e\x1f\x85\xa0chunked\x0b\x0c\x1c\x1d\x1e\x1f\x85\xa0` header, it treats this equivalently to `Transfer-Encoding: chunked`.

This is useful as a request smuggling primitive.

You can reproduce this as follows:
1. Start a Gunicorn server that echoes request bodies, such as [this one](https://raw.githubusercontent.com/narfindustries/http-garden/main/images/gunicorn/server.py).
2. Send it a request that triggers the bug:
```sh
printf 'POST / HTTP/1.1\r\nHost: whatever\r\nTransfer-Encoding: \x0b\x0c\x1c\x1d\x1e\x1f\x85\xa0chunked\x0b\x0c\x1c\x1d\x1e\x1f\x85\xa0\r\n\r\n1\r\nZ\r\n0\r\n\r\n' \
    | ncat localhost 80 \
    | grep "body" \
    | jq '.["body"]' \
    | xargs echo \
    | base64 -d \
    | xxd
```
3. Observe that the message body was interpreted as chunked:
```
00000000: 5a                                       Z
```

benoitc:
on which version did you test? also no need to use the catch-all "request smuggling".  If we strip there it's a parsing issue.

benoitc:
this has been there for a while and never been an issue. What should we do now? return a 400 ?

kenballus:
> What should we do now? return a 400 ?

Yes.

benoitc:
@kenballus will see. I still have to see rthis happen and still have to see how can somecall this a security issue when any app would simply just not do anything about an unknown request.

kenballus:
Once again, you are demonstrating that you do not understand request smuggling. It's not about the app; it's about **upstream load balancers**.

Suppose that you have the following scenario:
```
       Request                Request
       ------->               ------->
Client          Load Balancer          Gunicorn
       <-------               <-------
       Response               Response
```

The bug is exploitable when there are discrepancies between the load balancer and Gunicorn. The app that Gunicorn invokes **does not matter**.

The load balancer cannot do its job when its interpretation of the request stream differs from that of the origin server (Gunicorn). Exploiting these discrepancies is "request smuggling," which can be used for
- cache poisoning,
[...]

PR #3369, opened 2025-03-20
remove whitespace in transfer-encoding

pajod:
HTTP parser Whitespace cleanup

1. one no-op replacement
2. one variable reuse after casting to tuple, to clarify these are still just SP and HTAB
3. another no-op replacement
4. a meaningful change for https://github.com/benoitc/gunicorn/issues/3364
   * N.B. this still disallows `Content-Encoding: identity,` (trailing comma), probably a [robust](https://en.wikipedia.org/wiki/Robustness_principle) choice
   * no longer meaningful after 2073e13dc80969074954cabd61c01d5e1f427d7f - `\x0b` is rejected by the new `RFC9110_5_5_INVALID_AND_DANGEROUS`
   * N.B. responding 501 is not [what the original reporter recommended](https://github.com/benoitc/gunicorn/issues/3364#issuecomment-2741541765) - but differentiating invalid and unsupported is is complicated because of potential parameters, and I do no see how the difference matters.
5. a no-op replacement in checking the `CONNECTION` header.. which should probably just be dropped. How could `v != v.strip(" \t")` legitimately happen, unless something was wrong with obsolete folding?

Searching for `strip()` revealed additional references in `cfg.strip_header_spaces` (known dangerous) in `util.is_hoppish()` (unreachable after 72238fcf8d29db62fdbb9d2481ced801e27fe139 - left as-is but did add the extra header rfc9112 tells us to drop) and in the websockets example (not reviewed yet).

References:
* https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.3
* https://datatracker.ietf.org/doc/html/rfc9110#section-5.6.1.2

### DIFF
--- a/tests/requests/valid/obs_fold_whitespace.http
+++ b/tests/requests/valid/obs_fold_whitespace.http
@@ -1,0 +1,18 @@
+POST /1 HTTP/1.1\r\n
+Host: example.com\r\n
+Connection: keep-alive\r\n
+ \r\n
+Content-Length: 3\r\n
+\r\n
+123
+POST /2 HTTP/1.1\r\n
+Host: example.com\r\n
+Connection: close\r\n
+ \r\n
+Transfer-Encoding: chunked\r\n
+\r\n
+3\r\n
+123\r\n
+0\r\n
+Trailer-Checksum: 0\r\n
+\r\n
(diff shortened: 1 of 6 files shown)


tests/requests/valid/obs_fold_whitespace.http
```diff
@@ -0,0 +1,18 @@
+POST /1 HTTP/1.1\r\n
+Host: example.com\r\n
+Connection: keep-alive\r\n
+ \r\n
```
pajod, line 4: I don't see any legitimate use for this. I suggest Gunicorn removes the `.strip()` call.

PR #3376, opened 2025-04-01
Fix the build for modern setuptools

vanschelven:


vanschelven:
I ran into this when doing [a similar fix on one my own projects](https://github.com/bugsink/bugsink/commit/bb3740ed127e5010c2a0e96a43d1b76dccc57802) and then grepping in my "under active development" directory for other packages running into the same.

My super-non-scientific process was: read `setuptools`' error messages and do as they say. This also means: I didn't bother to check which particular version of setuptools broke the build.

vanschelven:
I just noticed on another one of my projects that this "fix" [breaks against Python 3.8](https://github.com/bugsink/verbose_csrf_middleware/commit/c515b6e53efd404e5e1465e52e11616b50fd1297)

Thus, we'll need to wait until gunicorn drops Python 3.8 support https://github.com/benoitc/gunicorn/pull/3343

However, the present PR will need to make it in before By 2026-Feb-18, because according to the message:

> By 2026-Feb-18, you need to update your project and remove deprecated calls

vanschelven:
@pajod inasmuch as I have taken the time to understand what you're saying, I agree with your analysis of the situation, including its root cause :cry: 

Here's the proof of [Trixie's setuptools version](https://packages.debian.org/trixie/python/python3-setuptools ): (note that Trixie's EOL hasn't even been announced yet, but is likely to be at least in 2030).

My personal way forward would probably be the "do it damnit" solution of getting rid of the license info altogether (nr. 2) but I have no strong opinion on this.

vanschelven:
Python 3.8 support has been dropped by Gunicorn so the build passes now

benoitc:
fixed in 24.0 . Thanks for your PR :)

PR #3408, opened 2025-08-08
fix gthread infinite loop bug

### LINKED ISSUE #3407: gthread workers hang when `worker_connections` is low in some scenarios
Looking for advice on a PR to resolve an issue we're seeing specifically relating to keepalive connections when using gthread workers and a low `worker_connections` value.

I've observed that in some scenarios when stress-testing a gunicorn server, some of my workers would go unresponsive. The root cause was the loop in [gthread.py](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gthread.py#L203), where the following was happening:

1. `self.nr_conns = self.worker_connections` and `self.futures` was an empty `deque`.
2. The worker does not handle any new connections via `self.poller` but because `self.futures` is empty, there are no requests being handled in the task pool that could decrement `self.nr_conns`.
3. Thus, the worker ends trapped in an infinite loop where it neither polling from `self.poller` nor is it actively processing any connections.

This ended up being due to an edge case where many connections were handled simultaneously: because [`on_client_socket_read` is scheduled on `self.poller` alongside new connections without decrementing `self.nr_conns`](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gthread.py#L261), it's possible that the application can enter a state where too many open connections are allocated and the gunicorn stops enqueing new connections.

My question here is what y'all would think the best fix for this bug might be. One simple solution would be to upload [L208](https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/gthread.py#L208) to:

```python
if self.nr_conns < self.worker_connections or not self.futures:
```

but there could be a risk that this could process entirely new connections and push `self.nr_conns` beyond the limit. Another option might be to poll the selector and if `self.futures` is empty, re-register the method and data if it is not calling `on_client_socket_readable`. I guess it might be possible to set up a separate poller as well for specifically checking up on keep-alive connections, but I'm not sure how useful it would be to overcomplicate this codebase for a minor edge case.

In any of these cases I'd be willing to draft up a PR.

chadac:
resolves #3407 . The problem arises from `on_client_socket_readable`, which shares a queue with `self.accept`. When the number of active connections reaches the `worker_connections` limit and if all of those connections are expected to be terminated via `on_client_socket_readable`, the worker reaches a state where it no longer polls from `self.poller` (`self.nr_conns` >= self.worker_connections`) and hence is incapable of closing any connections. This happens more often when the application is under stress and `worker_connections` is set to a low value (such as 4).

This solution adds a selector specifically for sockets that are expected to close and uses this in case the application has finished handling all other connections and has no capacity to queue up any more. This should ensure that the number of active connections never exceeds `self.worker_connections` while avoiding this rare case of deadlock.

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -74,6 +74,8 @@ class ThreadWorker(base.Worker):
         # initialise the pool
         self.tpool = None
         self.poller = None
+        # poller for active connections
+        self.current_poller = None
         self._lock = None
         self.futures = deque()
         self._keep = deque()
@@ -90,6 +92,7 @@ class ThreadWorker(base.Worker):
     def init_process(self):
         self.tpool = self.get_thread_pool()
         self.poller = selectors.DefaultSelector()
+        self.current_poller = selectors.DefaultSelector()
         self._lock = RLock()
         super().init_process()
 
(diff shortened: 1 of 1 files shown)


chadac:
closing in favor of #3157

khomyakov42:
This issue is reproducible in version 23.0.0.

benoitc:
> This issue is reproducible in version 23.0.0.
24.0 is out.

PR #3419, opened 2025-09-21
Add more systemd notifications

zigarn:
- adding binding to socket when launching Gunicorn in systemd
- reloading notification, allow to switch to `Type=notify-reload` and remove `ExecReload` in the systemd unit
- stopping notification, allow systemd to know in case of stopping from non-systemd


pajod:
I submitted a very similar patch as part of #3285 which is only slightly different:
 * your code for the timestamps is both correct where mine was wrong and more obviously the correct clock
 * I have an extra branch because I try to also gracefully handle the edge case of aborted master-reexec without confusing systemd about which is the new-old `$MAINPID`

pajod (approved):

benoitc (changes requested):
please ensure systemd is only available for system that supports it.

zigarn:
@benoitc Adapted according to your feedbacks and rebased on master for new documentation.

zigarn:
Closing because stale PR.
@benoitc feel free to reopen, copy/adapt the code or whatever if you want to integrate this one day.

PR #3422, opened 2025-10-05
Update to support only Python 3.12 and 3.13

benoitc:
## Summary
This PR updates Gunicorn to support only the latest (N) and previous (N-1) Python versions, as requested.

## Changes
- Update CI workflows to test only Python 3.12 and 3.13
- Remove support for Python versions 3.7 through 3.11
- Update project metadata to require Python >= 3.12
- Update documentation to reflect new Python requirements
- Clean up CI configurations (GitHub Actions, AppVeyor, tox)

## Rationale
Supporting only N and N-1 Python versions:
- Reduces CI complexity and build times
- Allows use of modern Python features
- Ensures security updates are available
- Simplifies maintenance burden


vanschelven:
I've [started a discussion about this here](https://github.com/benoitc/gunicorn/discussions/3424)

.github/workflows/lint.yml
```diff
@@ -42,6 +42,7 @@ jobs:
         run: |
           # this will update docs/source/settings.rst - but will not create html output
           (cd docs && sphinx-build -b "dummy" -d _build/doctrees source "_build/dummy")
+          git update-index --assume-unchanged docs/source/settings.rst
```
pajod: Huh? That was the primary goal of having CI regenerate docs. So that introducing deviations between code & documentation is pointed out in pull requests. If the default is no longer OS-independent, then the docs should say so.

PR #3426, opened 2025-10-06
Migrate docs build to MkDocs

benoitc:
## Summary
- replace legacy Sphinx site with MkDocs Material
- move docs content into Markdown with refreshed landing page
- add automation to build and deploy via MkDocs

## Testing
- mkdocs build


pajod:
```diff
--- a/site/2010-news.html
+++ b/site/2010-news/index.html
```
Looks like `use_directory_urls=false` is the correct choice here, because [Cool URIs don't change](https://www.w3.org/Provider/Style/URI)

```diff
- -
+ &mdash;
```

- Is this needed - I seem to be able to use ASCII hyphen-minus inside bullets points in Markdown just fine?

Issue #3434, opened 2025-11-15
There is a pattern where signal names are not correctly retrieved on Linux


ryuichi1208:
When a child process dies due to a signal such as SIGABRT, the signal name and signal number may not be displayed correctly and may become invalid.

The following is the log when sending SIGABRT to a child process. gunicorn is using the latest version 23.0.0.

```
Worker (pid:74) was sent code 134!
```

benoitc:
Fixed in commit b650332c (Arbiter signal handling improvements #3441). Signal names are now properly displayed using `signal.Signals(sig).name` instead of showing 'code X'.

PR #3435, opened 2025-11-15
Fix: Display signal names instead of exit codes in worker termination logs

ryuichi1208:
Fix signal name display in worker termination logs

Previously, when a worker process was terminated by a signal like SIGABRT,the log would display "Worker (pid:X) was sent code 134!" instead of the actual signal name. This happened because the code incorrectly interpreted shell exit codes (128 + signal_number) as signal numbers.

### Changes:

- Use os.WIFSIGNALED() and os.WTERMSIG() to correctly detect and extract
    signal numbers from process exit status
- Use os.WIFEXITED() and os.WEXITSTATUS() to correctly extract exit codes
- Handle shell convention where exit code > 128 indicates signal termination
    (e.g., 134 = 128 + SIGABRT)
- Add tests for both direct signal termination and shell-style exit codes

After this fix, the log correctly displays:
"Worker (pid:X) was sent SIGABRT!" instead of "Worker (pid:X) was sent code 134!"

Fixes signal name logging for all signals including SIGABRT, SIGKILL, SIGTERM, etc.

issue: https://github.com/benoitc/gunicorn/issues/3434

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -523,36 +523,47 @@ class Arbiter:
                     # A worker was terminated. If the termination reason was
                     # that it could not boot, we'll shut it down to avoid
                     # infinite start/stop cycles.
-                    exitcode = status >> 8
-                    if exitcode != 0:
-                        self.log.error('Worker (pid:%s) exited with code %s', wpid, exitcode)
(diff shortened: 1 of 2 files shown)


benoitc:
Closing as the related issue #3434 has been resolved. Thank you for your contribution!

PR #3441, opened 2026-01-22
Arbiter signal handling improvements

### LINKED ISSUE #3004: Deadlock after SIHCHLD signal handling
# Description

## Environment

`Python 3.9`
`Gunicorn 20.1.0`

## Current state

I have an application running with Gunicorn with 5 `gthread` workers.
I'm using a framework that in turn calls the `Arbiter.run` function. Before calling that function I start a separate thread which will sleep for a period of time and perform some logic before signalling the process (`os.getpid()`) with SIGHUP.
I implemented this in two different ways (both have the same issue which I'm describing in a section below).

1. Using an infinite `while True` loop.
2. Removing the loop and starting the thread in the beginning (pre-fork) and on each `on_reload' server hook.

I'm also using the `child_exit` and `worker_exit` server hooks which contain a log statement
in each function.

## Behavior

We've had this issue in 2 different scenarios.
1. When using the `max-requests` and `max-requests-jitter` configurations - After some successful autorestarts, at a random point in time, workers die after serving the max number of requests and no new workers are booted up.
2. Removing the previous configurations - After some successful reloads,  at a random point in time I see no progress being made by my separate thread and no new reloads being made.

What I've noticed is that there is some weird behavior going on with the `reap_workers` function whenever a SIGCHLD needs to be handled.
In that function we're looping over dead child process ids and later on `cfg.child_exit` is called which means my log statement should be printed each time. What I find weirdly suspicious is that each time the reload stops happening, the previous reload is not printing my log statement in `cfg.child_exit` for each dead child process id (for some process ids that log is missing). However I see the log statement in `cfg.worker_exit` which is called everytime (for all process ids) without failure, which indicates that that process indeed terminated. 

# Steps to reproduce
1. Create a script which calls the `Arbiter.run` function.
2. In the same process that calls that function create and start a second thread. The thread must be used to periodically reload the arbiter.
3. Wait some time and check if at a random point in time, there is a "deadlock/livelock" situation where no new reloads are being made and no progress is being made by the arbiter.

You need to preload the app.

### LINKED ISSUE #3050: Error instead of warning in logs when sending SIGTERM since v21

### LINKED ISSUE #3056: Signal exit log has pid hardcoded in message

### LINKED ISSUE #3198: Logging in signal handlers may be unsafe
In the arbiter's signal-handling code, there are many uses of logging. E.g.

https://github.com/benoitc/gunicorn/blob/master/gunicorn/arbiter.py#L252

however, the docs say this:

> If you are implementing asynchronous signal handlers using the [signal](https://docs.python.org/3/library/signal.html#module-signal) module, you may not be able to use logging from within such handlers. This is because lock implementations in the [threading](https://docs.python.org/3/library/threading.html#module-threading) module are not always re-entrant, and so cannot be invoked from such signal handlers.

https://docs.python.org/3/library/logging.html#thread-safety

[elsewhere](https://github.com/python/cpython/issues/118143#issuecomment-2070834944) Gregory Smith says the following:

>  I'd still rank "doing IO" even in a python signal handler (which is effectively possible to be executed between any given bytecode and between some operations using the C API that check the internal signal context) as a bad idea.

What the consequences of the above remarks are... may depend on how many signals you send.

### LINKED ISSUE #3274: Gunicorn in docker sends SIGKILL to workers when SIGHUP sent to master process
Hello all, I have flask-gunicorn api running in docker but the issue is this:
when sending SIGHUP to the master process, in docker it is 1, workers get SIGKILL signals - not gracefully stopped.
this is my config in docker
gunicorn version - 23.0.0
```
import multiprocessing

bind = "0.0.0.0:5001"
workers = multiprocessing.cpu_count() * 2
pidfile = "api.pid"
errorlog = "/logs/gunicorn-error.log"
keepalive = 10
max_requests = 500
max_requests_jitter = 30
backlog = 800
timeout = 4000
graceful_timeout = 20
capture_output = True
wsgi_app = "wsgi:app"
```
logs show
```
[2024-08-14 07:13:07 +0000] [1] [INFO] Starting gunicorn 23.0.0
[2024-08-14 07:13:07 +0000] [1] [INFO] Listening at: http://0.0.0.0:5001 (1)
[2024-08-14 07:13:07 +0000] [1] [INFO] Using worker: sync
[2024-08-14 07:13:07 +0000] [7] [INFO] Booting worker with pid: 7
[2024-08-14 07:13:07 +0000] [8] [INFO] Booting worker with pid: 8
[2024-08-14 07:13:07 +0000] [9] [INFO] Booting worker with pid: 9
[2024-08-14 07:13:07 +0000] [10] [INFO] Booting worker with pid: 10
[2024-08-14 07:14:01 +0000] [1] [INFO] Handling signal: hup
[2024-08-14 07:14:01 +0000] [1] [INFO] Hang up: Master
[2024-08-14 07:14:01 +0000] [28] [INFO] Booting worker with pid: 28
[2024-08-14 07:14:01 +0000] [29] [INFO] Booting worker with pid: 29
[2024-08-14 07:14:01 +0000] [30] [INFO] Booting worker with pid: 30
[2024-08-14 07:14:01 +0000] [9] [INFO] Worker exiting (pid: 9)
[2024-08-14 07:14:01 +0000] [10] [INFO] Worker exiting (pid: 10)
[2024-08-14 07:14:01 +0000] [7] [INFO] Worker exiting (pid: 7)
[2024-08-14 07:14:01 +0000] [8] [INFO] Worker exiting (pid: 8)
[2024-08-14 07:14:01 +0000] [31] [INFO] Booting worker with pid: 31
[2024-08-14 07:14:01 +0000] [1] [ERROR] Worker (pid:8) was sent SIGTERM!
[2024-08-14 07:14:01 +0000] [1] [ERROR] Worker (pid:7) was sent SIGTERM!
```
gunicorn version - 20.0.4 all works as expected
```
[2024-08-14 07:17:08 +0000] [1] [INFO] Handling signal: hup
[2024-08-14 07:17:08 +0000] [1] [INFO] Hang up: Master
[2024-08-14 07:17:08 +0000] [120] [INFO] Booting worker with pid: 120
[2024-08-14 07:17:08 +0000] [121] [INFO] Booting worker with pid: 121
[2024-08-14 07:17:08 +0000] [122] [INFO] Booting worker with pid: 122
[2024-08-14 07:17:08 +0000] [123] [INFO] Booting worker with pid: 123
[2024-08-14 07:17:08 +0000] [71] [INFO] Worker exiting (pid: 71)
[2024-08-14 07:17:08 +0000] [67] [INFO] Worker exiting (pid: 67)
[2024-08-14 07:17:08 +0000] [76] [INFO] Worker exiting (pid: 76)
[2024-08-14 07:17:08 +0000] [64] [INFO] Worker exiting (pid: 64)
```
what am I doing wrong?

benoitc:
## Summary

- Improve signal handler safety by removing logging from signal context
- Replace PIPE+select with queue.SimpleQueue for simpler, safer code
- Fix waitpid status parsing to correctly identify signal vs exit code
- Wait for old workers to terminate on SIGHUP reload
- Change SIGTERM termination log level to warning
- Log SIGCHLD at debug level to reduce log noise

## Closes

- Fixes #3198 (logging in signal handlers unsafe)
- Fixes #3004 (deadlock after SIGCHLD)
- Fixes #3050 (SIGTERM logged as error)
- Fixes #3056 (signal names not displayed correctly)
- Fixes #3274 (SIGHUP sends double SIGTERM)
- Closes #3148, #3312, #3311, #3151, #3435, #3057

### DIFF
--- a/tests/test_arbiter.py
+++ b/tests/test_arbiter.py
@@ -2,7 +2,6 @@
 # This file is part of gunicorn released under the MIT license.
 # See the NOTICE for more information.
 
-import errno
 import os
 import signal
 from unittest import mock
@@ -452,7 +451,6 @@ class TestSighupReload:
         # Mock app.reload to prevent it from resetting config
         arbiter.app.reload = mock.Mock()
         # Mock setup to prevent it from resetting num_workers
-        original_setup = arbiter.setup
         arbiter.setup = mock.Mock()
 
         arbiter.reload()


gunicorn/arbiter.py
```diff
@@ -394,9 +388,12 @@ def stop(self, graceful=True):
         self.kill_workers(sig)
         # wait until the graceful timeout
         while self.WORKERS and time.time() < limit:
+            self.reap_workers()
             time.sleep(0.1)
```
pajod: Why the change in order from #3148?
What would be the benefit of waiting *after* the final `waitpid()` in that loop?

PR #3454, opened 2026-01-23
Refactor: plain socket objects instead of wrapper classes

pajod:
* #3127 still works for my limited use case
* still looks like a straightforward and useful refactor
* I would still volunteer to work on it, if there is any work to be done

So, here is a rebased version with the fixup I commented earlier.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -158,7 +158,7 @@ class Arbiter:
             if not (self.cfg.reuse_port and hasattr(socket, 'SO_REUSEPORT')):
                 self.LISTENERS = sock.create_sockets(self.cfg, self.log, fds)
 
-        listeners_str = ",".join([str(lnr) for lnr in self.LISTENERS])
+        listeners_str = ",".join([sock.get_uri(lnr, self.cfg.is_ssl) for lnr in self.LISTENERS])
         self.log.debug("Arbiter booted")
         self.log.info("Listening at: %s (%s)", listeners_str, self.pid)
         self.log.info("Using worker: %s", self.cfg.worker_class_str)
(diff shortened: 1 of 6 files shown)


pajod:
Whoops. There is no regression test for `--enable-backlog-metric`!

benoitc (commented):
Thanks for the refactor, code is cleaner :)

Few things to address:

**ASGI Worker needs updating**

The ASGI worker (`gunicorn/workers/gasgi.py`) still accesses `sock.sock` (line 181) and uses `str(sock)` for logging (line 187) - won't work with plain sockets.

**listen() for fd-passed sockets**

Old code called `sock.listen(self.conf.backlog)` including for systemd-passed fds. New `set_socket_options()` doesn't call it. Could this cause issues with systemd socket activation?

**SSL validation timing**

Changing to `validate_file_exists` at config time might break setups where certs are generated dynamically before start. Maybe keep runtime validation?

**Tests**

As you noted, need regression test for `--enable-backlog-metric`. Also missing tests for `get_uri()`, `get_backlog()`, `set_socket_options()`.

Otherwise looks good - the `close_sockets()` error handling improvement is nice.

pajod:
> ASGI Worker needs updating

Copied the changes there, but did not check beyond running tests.

> listen() for fd-passed sockets

Right. I dont *think* it does anything, now that I dropped the `fromfd` duplication. Its the same socket, so why would the kernel mess with the backlog hint? But I'll put it back to avoid unforeseen consequences. I don't see any pressing need to change that right now.

> SSL validation timing - Maybe keep runtime validation?

Fair point. Reverted.

> Tests

I pushed some more unit tests (and the whole thing still passes my nginx test).

----

I have one more commit to fix edge cases like `--bind="[fe80::1%enp21s0]:80"` but that is fairly verbose with new socket.getaddrinfo/socket.getnameinfo calls (and also known to be broken for eventlet) so I will leave this PR as-is and submit that separately.

----

Things I attempt to not touch here, but would like to follow up with:
 * Make reuse_port behaviour clear and consistent on non-supporting platforms
 * Make LISTEN_FDS= compatible with --bind being used simultaneously (and fix reload/re-exec for all resulting edge cases)

pajod:
I would be interested to receive clear confirmation from someone using the `uvicorn-worker` package and receiving superfluous "Bad file descriptor" messages on whether these changes resolve the issue, as suggested regarding the original PR this is a rebase for.

qubydev:
@pajod  I just quit using python man.

PR #3457, opened 2026-01-23
detect obviously unintended address mask

pajod:
When there is a digit missing from the ip address mask, then Python stdlib offers an option to pretend it did not notice.
We do **not** want to use that option. Because when the caller went through the extra effort to spell out the address including non-zero bits covered by the mask, then what they meant was clearly not permitting half the internet!

Revert that bit from e52ac46e297d9d483ad8cf16e43487545185a997 and add to regression test.


pajod:
Closing because rolled into #3459

PR #3464, opened 2026-01-25
asgi: get valid/invalid http tests to pass

pajod:
There are 101 perfectly fine HTTP test cases in `tests/requests/**.http` many of which of no use for testing the ASGI workers right now. Let's fix that.

* change asgi HTTP error exceptions to match base sync worker
  * did not look intentional, link to diff in PR here: https://github.com/benoitc/gunicorn/pull/3444#discussion_r2724675416
* <del>use StopAsyncIteration in async def - my reasoning here is that sticking with this signalling is *exclusively* practical when the exception can bubble in async iterator</del>
* override config in cases tests/test_asgi.py::MockConfig has non-default
  * those test cases actually care about the configs set there, and though this could equally be fixed by switching the test to use the default settings instead, *explicit is better than implicit*

(updated `tests/treq.py` and new `tests/test_nginx.py` intentionally omitted here, they match most of my parallel PRs)

### DIFF
--- a/gunicorn/asgi/message.py
+++ b/gunicorn/asgi/message.py
@@ -16,6 +16,7 @@ import struct
 from gunicorn.http.errors import (
     ExpectationFailed,
     InvalidHeader, InvalidHeaderName, NoMoreData,
+    InvalidChunkSize, ChunkMissingTerminator,
     InvalidRequestLine, InvalidRequestMethod, InvalidHTTPVersion,
     LimitRequestLine, LimitRequestHeaders,
     UnsupportedTransferCoding, ObsoleteFolding,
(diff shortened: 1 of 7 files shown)


pajod:
Not resolved here (because invalid requests were never checked as thoroughly as valid ones - so its not strictly needed for those tests), but a good candidate after some refactoring: dumping a traceback inplace of logging the warning and responding

> gunicorn.http.errors.InvalidRequestLine: Invalid HTTP request line: ''

benoitc:
i suggest to wait for the deduplication refactoring .  Coming week I will release the 25.0?0 with current state. Then what about making it for 25.1 ?

pajod:
reason for close: mostly superseded by b00f125755ec3f509a3e82dc5568d9f2d8bddba7

Issue #3386, opened 2025-04-16
make test fails: setup.py: [Errno 2] No such file or directory


adk-swisstopo:
I am trying to run the packaged tests. I don't really see any specific instructions but there is a Makefile with a promising "test" target. Unfortunately it is broken as it relies on a setup.py file that does not exist and that the build target does not generates.

Can you please provide instructions on how to run the tests?

make build && make test output follows

```
╰─ make build                                                                                                                                                                                 
virtualenv venv                                                                                                                                                                               
created virtual environment CPython3.10.12.final.0-64 in 62ms                                                                                                                                 
  creator CPython3Posix(dest=[...]/gunicorn/venv, clear=False, no_vcs_ignore=False, global=False)                           
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=[...]/.local/share/virtualenv)
    added seed packages: pip==24.3.1, setuptools==75.5.0, wheel==0.45.0                                                                                                                       
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
venv/bin/pip install -e .                                                                      
Obtaining file:///[...]/gunicorn                                                                                                                                                     
  Installing build dependencies ... done                                                       
  Checking if build backend supports build_editable ... done                             
  Getting requirements to build editable ... done                              
  Preparing editable metadata (pyproject.toml) ... done                                
Collecting packaging (from gunicorn==23.0.0)                                                                                                                                                  
  Using cached packaging-24.2-py3-none-any.whl.metadata (3.2 kB)                        
Using cached packaging-24.2-py3-none-any.whl (65 kB)                          
Building wheels for collected packages: gunicorn                                      
  Building editable for gunicorn (pyproject.toml) ... done                      
  Created wheel for gunicorn: filename=gunicorn-23.0.0-0.editable-py3-none-any.whl size=5325 sha256=b898e4c78f9cb5f9a3485bcb43989bdfa7bb6ce52bc720bdebfbeb3f44e83b1c
  Stored in directory: /tmp/pip-ephem-wheel-cache-[...]
Successfully built gunicorn                                                                                                                                                                   
Installing collected packages: packaging, gunicorn                                                                                                                                            
Successfully installed gunicorn-23.0.0 packaging-24.2                 

[notice] A new release of pip is available: 24.3.1 -> 25.0.1
[notice] To update, run: python -m pip install --upgrade pip
[...]

PR #3471, opened 2026-01-27
Remove obviously broken make targets.

adk-swisstopo:
The `test` and `coverage` have been broken for "about 10 years" and there is apparently no interest in fixing them. This PR removes these targets.

As per https://github.com/benoitc/gunicorn/issues/3386


benoitc:
thanks!

PR #3490, opened 2026-02-06
fix: passing maxsplit in re.split() as positional argument is deprecated

tyrossel:
This has been deprecated in python 3.13 [source](https://docs.python.org/3/library/re.html#re.split)


pajod (approved):
Reference: https://github.com/python/cpython/issues/56166

benoitc:
Thanks for the fix!

PR #3493, opened 2026-02-09
Strip body and Content-Length from 204/304 responses (RFC 9112)

### LINKED ISSUE #3413: 204/304 responses should not send body or Content-Length (RFC 9112 compliance)
Currently, Gunicorn forwards 204 and 304 responses even when the app includes a non-zero `Content-Length` or a body.

According to RFC 9112 , these responses must not include a message body, and `Content-Length` must be absent or zero. 

### Expected behavior
For 1xx, 204, 304 responses (and HEAD requests):
- Any response body should be ignored or dropped.
- `Content-Length` should be removed.

### Actual behavior
Gunicorn sends the body and headers as provided by the WSGI app, even when they violate the spec.

daniel7an:
## Summary

Per RFC 9112 Section 6.3, responses with status codes 204 and 304 must not contain a message body, and `Content-Length` should be absent.

Currently, Gunicorn forwards 204/304 responses as-is from the WSGI app, including any body and `Content-Length` header. This can cause issues with downstream proxies (e.g., AWS ALB returning 502).

## Changes

- **`gunicorn/http/wsgi.py`**: Strip `Content-Length` header during `process_headers()` for 204/304 responses. Silently discard body data in `write()` for these status codes.
- **`tests/test_http.py`**: Added tests for 204, 304 (Content-Length stripped, body discarded) and 200 (normal behavior preserved).

## Behavior

| Status | Before | After |
|--------|--------|-------|
| 204 | Body + Content-Length sent as-is | Body discarded, Content-Length stripped |
| 304 | Body + Content-Length sent as-is | Body discarded, Content-Length stripped |
| 200 | Normal | Normal (unchanged) |

Other frameworks (Flask, FastAPI, Express, Spring) already enforce this behavior.

Fixes #3413

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -330,6 +330,10 @@ class Response:
             value = value.strip(" \t")
             lname = name.lower()
             if lname == "content-length":
+                # RFC 9112 6.3: 1xx, 204, and 304 responses must not
+                # contain a message body, so Content-Length is stripped.
+                if self.status_code in (204, 304):
+                    continue
                 self.response_length = int(value)
             elif util.is_hoppish(name):
                 if lname == "connection":
@@ -395,6 +399,12 @@ class Response:
         self.send_headers()
         if not isinstance(arg, bytes):
             raise TypeError('%r is not a byte' % arg)
+
+        # RFC 9112 6.3: 204 and 304 responses must not contain
+        # a message body. Silently discard any body data.
+        if self.status_code in (204, 304):
+            return
+
         arglen = len(arg)
         tosend = arglen
         if self.response_length is not None:
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -126,6 +126,69 @@ def test_http_invalid_response_header():
         response.start_response("200 OK", [('foo\r\n', 'essai')])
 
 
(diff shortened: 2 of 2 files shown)


benoitc:
either we strip or we raise an issue. But we shouldn't do it transparently. That will let the app continue to send a body without beeing fixed. 

I suggest that on 304/204 and similar we have such behaviour:

1. gunicron should assume no content length is given
2. if content lenght is given , raise an exception
3. if no content lenght or chunked encoding given but we still get a body raise an exception

optionally,  but this would have to be configured, we should catch the exception, log a warning, and ignore. But this will have to be configured. 

Thoughts?

pajod:
> if content lenght is given , raise an exception

With 204 status only.
A server MAY send a Content-Length header field in a 304 (Not Modified) response to a conditional GET request.

> if no content lenght or chunked encoding given but we still get a body raise an exception

With both 204 and 304 status raise on body.

I wonder if there might be any 0/NULL edge cases (write/yield empty bytestring) here? If not, order that check *before* `send_headers()`.

benoitc:
@daniel7an any reasonyou closed the PR? Would be nice to say why.

benoitc:
bump

PR #3502, opened 2026-02-12
docs: fix Markdown formatting in /configure

Juneezee:
Fixes https://github.com/benoitc/gunicorn/issues/3498.

Before:

before

After:

after


benoitc:
thanks!

PR #3504, opened 2026-02-12
feat(dirty): add TTIN/TTOU signal support for dynamic worker scaling

### LINKED ISSUE #3489: h2c support
Bring H2C support to HTTP2 support.

### Discussed in https://github.com/benoitc/gunicorn/discussions/3477

<div type='discussions-op-text'>

<sup>Originally posted by **timkofu** February  2, 2026</sup>
### Type

Feature Request

### Description

>HTTP/2 cleartext (h2c) is not supported due to security concerns and lack of browser support.

PaaS like [fly.io](https://community.fly.io/t/http-2-from-fly-proxy-to-an-app-is-now-supported-with-http-handler/16344) and [cloudrun](https://docs.cloud.google.com/run/docs/configuring/http2) support h2c, so we can have http2 features and the capabilities they unlock in our apps (without terminating TLS ourselves) and keep the advantages of their proxies.

### Steps to Reproduce (for bugs)

_No response_

### Configuration

_No response_

### Logs / Error Output

_No response_

### Gunicorn Version

gunicorn (version 25.0.0)

### Python Version

Python 3.14.2

### Worker Class

gthread

### Operating System

Linux

### Additional Context

_No response_

### Checklist

- [X] I have searched existing discussions and issues for duplicates
- [X] I have checked the documentation and FAQ</div>

benoitc:
## Summary

- Add SIGTTIN/SIGTTOU signal support to dirty arbiters for dynamic worker scaling
- Allows scaling dirty workers up/down at runtime without restarting gunicorn
- Respects minimum worker constraints from app configurations

## Changes

- Add `num_workers` instance variable for dynamic count
- Add `_get_minimum_workers()` to calculate minimum required workers from app specs
- Add signal handlers for TTIN (increase) and TTOU (decrease)
- Update `manage_workers()` to use dynamic count instead of config
- Add documentation for dynamic scaling in dirty.md
- Add unit tests (17 tests)
- Add Docker integration tests

## Usage

```bash
# Find the dirty arbiter process
ps aux | grep dirty-arbiter

# Increase workers
kill -TTIN <dirty-arbiter-pid>

# Decrease workers
kill -TTOU <dirty-arbiter-pid>
```

## Minimum Worker Constraint

The dirty arbiter will not decrease below the minimum workers required by app configurations. For example, if an app has `workers = 3`, you cannot scale below 3 dirty workers.

```
WARNING: SIGTTOU: Cannot decrease below 3 workers (required by app specs)
```

Closes #3489

### DIFF
--- a/tests/docker/dirty_ttin_ttou/gunicorn_conf.py
+++ b/tests/docker/dirty_ttin_ttou/gunicorn_conf.py
@@ -1,0 +1,23 @@
+#
+# This file is part of gunicorn released under the MIT license.
+# See the NOTICE for more information.
+
+"""Gunicorn configuration for TTIN/TTOU testing."""
+
+bind = "0.0.0.0:8000"
+workers = 2
+worker_class = "gthread"
+threads = 2
+
+# Dirty arbiter config
+dirty_apps = [
+    "app:UnlimitedTask",
+    "app:LimitedTask",  # Has workers=2 attribute
+]
+dirty_workers = 3
+dirty_timeout = 30
+
+# Logging
+loglevel = "debug"
+accesslog = "-"
+errorlog = "-"
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ gunicorn myapp:app --worker-class asgi
 ## Features
 
 - WSGI support for Django, Flask, Pyramid, and any WSGI framework
-- **ASGI support** (beta) for FastAPI, Starlette, Quart
+- **ASGI support** for FastAPI, Starlette, Quart
 - **HTTP/2 support** (beta) with multiplexed streams
-- **Dirty Arbiters** for heavy workloads (ML models, long-running tasks)
+- **Dirty Arbiters** (beta) for heavy workloads (ML models, long-running tasks)
 - uWSGI binary protocol for nginx integration
 - Multiple worker types: sync, gthread, gevent, eventlet, asgi
 - Graceful worker process management
--- a/docs/content/2026-news.md
+++ b/docs/content/2026-news.md
@@ -18,6 +18,12 @@
   - Supports all Python types: str, bytes, int, float, bool, None, list, dict
   - Better performance for large payloads
 
+- **Dirty TTIN/TTOU Signals**: Add dynamic worker scaling for dirty arbiters
+  ([PR #3504](https://github.com/benoitc/gunicorn/pull/3504))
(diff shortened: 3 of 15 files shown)


tests/docker/dirty_ttin_ttou/gunicorn_conf.py
```diff
@@ -0,0 +1,23 @@
+#
+# This file is part of gunicorn released under the MIT license.
+# See the NOTICE for more information.
+
+"""Gunicorn configuration for TTIN/TTOU testing."""
+
+bind = "0.0.0.0:8000"
```
jayaddison, line 7: @benoitc this is nitpicky, but I wonder whether this could/should be overridden by adding `GUNICORN_BIND` in the [`dirty_ttin_ttou/docker-compose.yml`](https://github.com/benoitc/gunicorn/pull/3504/changes#diff-0f5f4fd2133b74bd1315050e2fc7148724ff7665de6641b4abfeca8a4f337c15) file instead?

(my thinking is that sysops/devops people are more likely to read that file early in the deployment/audit process -- and more likely when/during inspecting other compose files -- while individual conf files may take more time to discover or remember about -- and so, I'm thinking that the compose file might be a preferable location so that the network/port config is easier to learn about and compare)

jayaddison: Self-quoting:
> @benoitc this is nitpicky, but I wonder whether this could/should be overridden by adding `GUNICORN_BIND` in the [dirty_ttin_ttou/docker-compose.yml](https://github.com/benoitc/gunicorn/pull/3504/changes#diff-0f5f4fd2133b74bd1315050e2fc7148724ff7665de6641b4abfeca8a4f337c15) file instead?

In retrospec: no: for the Docker test case configurations (e.g. `tests/docker/*/gunicorn_conf.py`), I think it's somewhat reasonable for the server code to bind to `0.0.0.0` by default.

_Perhaps_ if each container had a single IP+port for publishing on, then we could bind solely to that in each configuration (either statically or dynamically) -- but I'm not aware of an easy way to specify-and-or-retrieve that for Docker containers externally, other than hardcoding values for each container in the `docker-compose.yml` files, and that'd be a maintenance annoyance.

So: I'll close this as resolved.

A similar question _might_ apply to whether all of the example code within the `examples` directory should consistently inspect `GUNICORN_BIND`, but that's a separate question.

PR #3521, opened 2026-02-20
fix(http): decrement LengthReader.length by actual bytes read, not requested size

bysiber:
`LengthReader.read()` decrements `self.length` by the requested `size` rather than the actual number of bytes returned:

```python
ret, rest = buf[:size], buf[size:]
self.unreader.unread(rest)
self.length -= size    # should be len(ret)
return ret
```

When the underlying reader returns fewer bytes than requested (slow client, premature connection close), `ret` is shorter than `size`, but `self.length` is still reduced by the full amount. This causes `self.length` to undercount the remaining bytes in the request body.

On keepalive connections (gthread and async workers), this can cause `parser.finish_body()` to stop draining the body too early, leaving unconsumed bytes on the socket that bleed into the next HTTP request's parsing.

### DIFF
--- a/gunicorn/http/body.py
+++ b/gunicorn/http/body.py
@@ -138,7 +138,7 @@ class LengthReader:
         buf = buf.getvalue()
         ret, rest = buf[:size], buf[size:]
         self.unreader.unread(rest)
-        self.length -= size
+        self.length -= len(ret)
         return ret
 
 


benoitc:
Hrm  this doesn't make sense. The only way len(ret) < size is when unreader.read() returns empty, which happens when the connection is closed. How can I reproduce?

bysiber:
You're right that in the normal path `unreader.read()` returns the full slice. The scenario I had in mind was when the underlying socket returns short reads due to TCP framing, but looking at it again — `unreader.chunk()` is the layer that does the actual `recv()`, and `read()` slices from the buffered chunk, so `buf[:size]` might indeed always have `len(ret) == size` when there are enough bytes available.

The only way `len(ret) < size` is if the chunk is shorter than requested, and that should only happen close to EOF. In that case the discrepancy between `size` and `len(ret)` makes `self.length` go negative faster, but `read()` already guards on `self.length <= 0` so it just terminates the body read early, which is the correct behavior for a truncated request anyway.

So you're right, the practical impact is limited to already-corrupted connections. If you'd rather close this as wontfix I'm fine with that.

benoitc:
Ok thanks for the feedback. I am closing it at won't fix then.

PR #3522, opened 2026-02-20
fix: prevent UnboundLocalError in sd_notify when socket creation fails

bysiber:
If `socket.socket()` raises an exception in `sd_notify()` (e.g., resource exhaustion, permission denied), the `sock` variable is never assigned. The `finally` block unconditionally calls `sock.close()`, which raises `UnboundLocalError` — masking the original exception and potentially crashing the arbiter during startup or shutdown.

This initializes `sock = None` before the try block and guards the close call.


benoitc:
Merged. Thanks!

PR #3523, opened 2026-02-20
fix: ensure pidfile fd is closed on rename failure

bysiber:
In `Pidfile.create()`, if `os.rename()` fails (cross-device, permissions, etc.), the file descriptor from `tempfile.mkstemp()` is never closed — `os.close(fd)` is placed after the rename without any try/finally protection.

On repeated failures during reload cycles, this accumulates leaked file descriptors. Wrapping the write+rename in try/finally ensures the fd is always closed.


benoitc:
Merged. Thanks!

PR #3546, opened 2026-03-18
workers: drain unread request body before close to prevent TCP RST

### LINKED ISSUE #3334: GET request with empty transfer-encoding: chunked body causes Connection reset by peer on large responses
If I make a large empty file:
```bash
mkfile -n 1g temp_1GB_file
```
A basic gunicorn+Flask server to serve it in test.py:
```python
from flask import Flask

app = Flask(__name__, static_folder='.')

@app.route('/<path:path>')
def static_proxy(path):
    return app.send_static_file(path)
```
Run using:
```bash
gunicorn --workers=2 test:app
```
And then some code to (slowly) GET the large file through the webserver, but in a way that forces the request to have `transfer-encoding: chunked`:
```python
import urllib3
import time

def data():
    time.sleep(4)
    if False:
        yield b''

resp = urllib3.request("GET", "http://localhost:8000/temp_1GB_file", 
    body=data(),
    preload_content=False,
)
for chunk in resp.stream(65536):
    time.sleep(0.0005)
```
Then _sometimes_ I get the error:
```
urllib3.exceptions.ProtocolError: ("Connection broken: ConnectionResetError(54, 'Connection reset by peer')", ConnectionResetError(54, 'Connection reset by peer'))
```
Interestingly:
- if I change the request to a `POST`, the error doesn't occur
- or if I make a GET request but _without_ `transfer-encoding: chunked` (e.g. via `body=b''`) the error doesn't occur
- or if I remove the sleeps in the request code, the error doesn't occur
- or if I use Flask's built-in webserver via `flask --app test run --port 8080` instead of gunicorn, the error doesn't occur

TimRots:
When keep-alive is disabled and a WSGI app does not consume the full request body, closing the socket with unread data in the receive buffer triggers the kernel to send a TCP RST instead of a graceful FIN. This prematurely aborts the connection, often causing clients to encounter a ConnectionResetError before they can finish reading the response.

This change unifies the close path with the keep-alive path by calling parser.finish_body() to drain the remaining request data in the finally block of sync, base_async, and gthread workers.

To prevent worker exhaustion from slow-reading or malicious clients, a 1-second timeout is enforced during this drain. This ensures the worker moves to the next task while still providing a window for a graceful TCP shutdown.

Fixes #3334

### DIFF
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -32,6 +32,7 @@ class AsyncWorker(base.Worker):
 
     def handle(self, listener, client, addr):
         req = None
+        parser = None
         try:
             # Complete the handshake to ensure ALPN negotiation is done
             # (needed if do_handshake_on_connect is False)
@@ -98,6 +99,14 @@ class AsyncWorker(base.Worker):
         except BaseException as e:
             self.handle_error(req, client, addr, e)
         finally:
+            # Drain unread request body to prevent RST on close.
+            # See: https://github.com/benoitc/gunicorn/issues/3334
+            if parser:
+                try:
+                    client.settimeout(1)
+                    parser.finish_body()
+                except Exception:
+                    pass
             util.close(client)
 
     def handle_http2(self, listener, client, addr):
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -443,6 +443,7 @@ class ThreadWorker(base.Worker):
     def handle(self, conn):
         """Handle a request on a connection. Runs in a worker thread."""
         req = None
+        drain = False
         try:
             # For new connections (not yet initialized), wait for data with timeout
             # to prevent slow clients from blocking thread pool slots indefinitely.
@@ -468,6 +469,7 @@ class ThreadWorker(base.Worker):
             if conn.is_http2:
                 return self.handle_http2(conn)
 
+            drain = True
             req = next(conn.parser)
             if not req:
                 return False
@@ -478,6 +480,7 @@ class ThreadWorker(base.Worker):
                 # Discard any unread request body before keepalive
                 # to prevent socket appearing readable due to leftover bytes
                 conn.parser.finish_body()
+                drain = False
                 return True
         except http.errors.NoMoreData as e:
             self.log.debug("Ignored premature client disconnection. %s", e)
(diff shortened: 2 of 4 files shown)


benoitc:
I think kit would be better to force shutdown instead. Draining sockets likes this is the door to ddos . How do you reproduce the issue?

TimRots:
> I think kit would be better to force shutdown instead. Draining sockets likes this is the door to ddos .

Thanks for checking, I'm interested to see how this works out.

>  How do you reproduce the issue?

This was my reproducer:

* gunicorn sync worker, WSGI app returns 6.5MB response without reading request body
```
    start_response('200 OK', [])
    return [b'-' * 6553600]
```

* urllib3 POST with streaming body (1-byte, 1s delayed), reads response slowly
```
def data():
    time.sleep(1)
    yield b'-'

try:
    resp = urllib3.request("POST", "http://localhost:8000/",
        body=data(), preload_content=False)
    for chunk in resp.stream(65536):
        time.sleep(0.001)
    print("SUCCESS - no ConnectionResetError")
except Exception as e:
    print(f"FAILED - {type(e).__name__}: {e}")
```

* `tcpdump -i lo -nn port 8000` on loopback
```
  Client (:41934)          Server (:8000)
    |                           |
    |--- SYN ------------------>|  [S]  seq 3250166847
    |<-- SYN,ACK ---------------|  [S.] seq 3586771697 ack 3250166848
    |--- ACK ------------------>|  [.]  ack 1
    |                           |
    |--- POST headers (130B) -->|  [P.] seq 1:131
    |<-- ACK -------------------|  [.]  ack 131
    |                           |
    |<-- HTTP 200 (121B hdr) ---|  [P.] seq 1:122
    |--- ACK ------------------>|  [.]  ack 122
    |                           |
    |<== 6.5MB response data ===>|  (many data segments, ~80 packets)
    |                           |
    |--- chunked body (6B) ---->|  [P.] seq 131:137  (after 1s delay)
    |--- chunked body (5B) ---->|  [P.] seq 137:142  (final chunk)
    |                           |
    |<-- data continues --------|  (server pushing remaining response)
    |--- ACK (win 0) ---------->|  [.]  ack 2456698 win 0  (client backpressure)
    |                           |
    |<-- RST -------------------|  [R.] seq 2456698 ack 142  <-- KERNEL ABORTS
    X  ConnectionResetError     |     (unread body data in recv buffer)
```

Post patching a FFA pattern is observed and the socket `close()` was handled gracefully for my use case:
```
22:28:48.215170 Server > Client: Flags [FP.], seq 6517827:6553737, length 35910  FIN!
22:28:48.227695 Client > Server: Flags [F.],  seq 142, ack 6553738               FIN!
22:28:48.227736 Server > Client: Flags [.],   ack 143                            ACK
```

TimRots:
> I think kit would be better to force shutdown instead. Draining sockets likes this is the door to ddos . How do you reproduce the issue?

Hi there. Did you have any time to take another look at this issue?

TimRots:
I swapped the backend to Go. C'est dommage.

benoitc:
ce qui est domage c'est de ne pas tester soit même les solutions proposées.  Merci pour le ton passif aggresif. Et bonne chance.

nicwolff:
> What is a shame is not to test or even the proposed solutions. Thank you for the aggressive passive tone. And good luck.

😢

PR #3550, opened 2026-03-23
fix: add __iter__ and __next__ to FileWrapper for PEP 3333 compliance

### LINKED ISSUE #3396: FileWrapper should be iterable
If I understand the WSGI specification (PEP 3333) the file wrapper callable should return an iterable object - see https://peps.python.org/pep-3333/#optional-platform-specific-file-handling . I see GUnicorn can handle file wrapper another way (respiter discrimination), but when trying to use a WSGI app that depends on the iterable interface, the app breaks. Can we please add __iter__() to FileWrapper and fix this?

r266-tech:
## Summary

Fixes #3396

The WSGI spec ([PEP 3333](https://peps.python.org/pep-3333/#optional-platform-specific-file-handling)) requires that `wsgi.file_wrapper` return an **iterable** object. Currently, Gunicorn's `FileWrapper` only implements `__getitem__`, which makes it iterable via old-style fallback but breaks code that explicitly relies on the iterator protocol (e.g., calling `iter()`, using `next()`, or passing to builtins expecting proper iterators).

## Changes

- Added `__iter__` returning `self`
- Added `__next__` reading the next chunk and raising `StopIteration` when done
- Added test in `tests/test_http.py` verifying the iterator protocol

## Backward Compatibility

This is fully backward compatible:
- Existing `__getitem__`-based iteration continues to work
- `for` loops, `list()`, and other iteration patterns all work
- The `__next__` logic mirrors the existing `__getitem__` logic exactly

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -9,7 +9,7 @@ from unittest import mock
 
 from gunicorn import util
 from gunicorn.http.body import Body, LengthReader, EOFReader
-from gunicorn.http.wsgi import Response
+from gunicorn.http.wsgi import FileWrapper, Response
 from gunicorn.http.unreader import Unreader, IterUnreader, SocketUnreader
 from gunicorn.http.errors import InvalidHeader, InvalidHeaderName, InvalidHTTPVersion
 from gunicorn.http.message import TOKEN_RE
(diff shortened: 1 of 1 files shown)


benoitc:
Thanks for the contribution!

Issue #3220, opened 2024-05-30
Workers that are not supposed to exist exit with code 255


louisabraham:
I launched a webserver with 1 worker and 1 thread (default).

Here is my log when making 1 request.

```
[2024-05-30 12:55:12 +0200] [49155] [INFO] Starting gunicorn 22.0.0
[2024-05-30 12:55:12 +0200] [49155] [DEBUG] Arbiter booted
[2024-05-30 12:55:12 +0200] [49155] [INFO] Listening at: http://0.0.0.0:8080 (49155)
[2024-05-30 12:55:12 +0200] [49155] [INFO] Using worker: aiohttp.GunicornWebWorker
[2024-05-30 12:55:12 +0200] [49159] [INFO] Booting worker with pid: 49159
[2024-05-30 12:55:12 +0200] [49155] [DEBUG] 1 workers
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49241) exited with code 255
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49241) exited with code 255.
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49242) exited with code 255
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49242) exited with code 255.
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49243) exited with code 255
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49243) exited with code 255.
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49244) exited with code 255
[2024-05-30 12:55:19 +0200] [49159] [ERROR] Worker (pid:49244) exited with code 255.
```

The request is handled fine. I traced `os.getpid()` in my app and it's 49159 all the way.
What are those additional workers with pid 49241, 49242, 49243, 49244 and why are they causing errors?

Note: I'm using `aiohttp.GunicornWebWorker`

louisabraham:
thank you! 
actually I have trouble to reproduce, closing this for now

louisabraham:
Hello, do you know when/if this branch will be merged?

PR #3566, opened 2026-03-31
Fix false error logs for non-worker child processes in reap_workers

joho54:
## Summary

When gunicorn runs as PID 1 in a container (common in Docker/ECS/Kubernetes without `tini` or `dumb-init`), `reap_workers()` emits false ERROR logs for non-worker child processes reaped via `waitpid(-1)`.

This PR moves the `WORKERS` membership check **before** exit status logging, so that non-worker processes are reaped silently at DEBUG level while real worker exits continue to be reported as errors.

## Problem

`reap_workers()` calls `os.waitpid(-1)`, which reaps **any** terminated child process. When gunicorn is PID 1, orphaned child processes are reparented to it by the kernel. Upon termination, these non-worker processes are reaped and incorrectly logged as:

```
[ERROR] Worker (pid:30740) exited with code 23
```

This triggers false alerts in error tracking systems (e.g. Sentry).

### Current code flow

```
waitpid(-1) -> get exit code -> log ERROR -> check if worker -> skip if not worker
```

The error log fires **before** the worker membership check, so every non-worker child with a non-zero exit code produces a spurious error.

## Evidence from production

**Environment:** AWS ECS Fargate, gunicorn 21.2.0, Python 3.11, Django backend

**Observed:**
- Sentry captured `Worker (pid:30740) exited with code 23` from `gunicorn.error` logger
- CloudWatch log analysis showed only 3 active workers in the container: pid 40, 41, 42 (matching `--workers 3`)
- **Zero** application logs from pid 30740 — it was never a gunicorn worker
- **Zero** `Booting worker` logs for pid 30740
- `max_requests` was not configured, ruling out worker recycling
- All 3 real workers continued processing requests normally before and after the event
- The error was a single occurrence over 7 days — a non-worker child process reaped by PID 1

## Fix

Move the `WORKERS.pop(wpid)` check to the top of the else branch:

```python
worker = self.WORKERS.pop(wpid, None)
if not worker:
    self.log.debug(
        "Reaped unknown child process (pid:%s, status:%s). "
        "Not a gunicorn worker.", wpid, status)
    continue

# Only report exit status for actual workers
exitcode = ...
```

**What stays the same:**
- `waitpid(-1)` is preserved — PID 1 must reap all children to prevent zombies
- Worker error reporting, `HaltServer` on boot/load errors, signal logging — all unchanged
- `child_exit` hook only fires for real workers (same as before)

**What changes:**
- Non-worker child processes are logged at DEBUG instead of ERROR
- No false alerts in monitoring systems

## Tests added

- `test_reap_non_worker_child_no_error_log` — non-worker exit with error code produces no ERROR log
- `test_reap_non_worker_child_with_signal_no_error_log` — non-worker killed by SIGKILL produces no OOM error
- `test_reap_mixed_worker_and_non_worker` — mixed scenario: non-worker is silent, real worker is reported

All existing tests pass (1671 passed, 214 skipped).

## Design note: why DEBUG instead of suppressing entirely

A question that may come up: is it safe to silently reap unknown child processes?

First, `waitpid` does not kill anything — it reaps an **already terminated** process from the kernel's process table. The three options are:

| Approach | Result |
|----------|--------|
| Don't reap | Zombie accumulates in process table (worse) |
| Reap + ERROR log | Zombie prevented, but false alert (current behavior) |
| Reap + DEBUG log | Zombie prevented, no false alert, still traceable (this PR) |

DEBUG was chosen over complete suppression so that operators can still trace unexpected child processes when needed (e.g. `--log-level debug`). This preserves observability without generating noise at the default log level.

The reaping itself is not optional — as PID 1, gunicorn **must** call `waitpid` for all children, whether it spawned them or not. This is standard POSIX behavior: the kernel reparents orphaned processes to init (PID 1), and init is responsible for reaping them.

## Related issues

- #3220 — Workers that are not supposed to exist exit with code 255
- #1585 — Proposed checking all worker pids instead of every child (closed, not merged)

This PR takes a simpler approach than #1585: rather than changing `waitpid(-1)` to per-pid `waitpid`, it keeps the reaping behavior intact and only fixes the log level for non-worker processes.

### DIFF
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -613,9 +613,24 @@ class Arbiter:
                     # and should not be reported as worker errors.
                     worker = self.WORKERS.pop(wpid, None)
                     if not worker:
-                        self.log.debug(
-                            "Reaped unknown child process (pid:%s, "
-                            "status:%s). Not a gunicorn worker.", wpid, status)
+                        if os.WIFEXITED(status):
+                            self.log.debug(
+                                "Reaped non-worker child (pid:%s) "
+                                "exited with code %s.", wpid,
+                                os.WEXITSTATUS(status))
+                        elif os.WIFSIGNALED(status):
+                            sig = os.WTERMSIG(status)
+                            try:
+                                sig_name = signal.Signals(sig).name
+                            except ValueError:
+                                sig_name = "signal {}".format(sig)
+                            self.log.debug(
+                                "Reaped non-worker child (pid:%s) "
+                                "received %s.", wpid, sig_name)
+                        else:
+                            self.log.debug(
+                                "Reaped non-worker child (pid:%s, "
+                                "status:%s).", wpid, status)
                         continue
 
                     # A worker was terminated. If the termination reason was
--- a/gunicorn/arbiter.py
+++ b/gunicorn/arbiter.py
@@ -606,6 +606,18 @@ class Arbiter:
                 if self.reexec_pid == wpid:
                     self.reexec_pid = 0
                 else:
+                    # Check if this is a known worker process first.
+                    # When running as PID 1 (e.g. in containers), the
+                    # arbiter inherits orphaned child processes and reaps
+                    # them via waitpid(-1). These are not gunicorn workers
+                    # and should not be reported as worker errors.
+                    worker = self.WORKERS.pop(wpid, None)
+                    if not worker:
+                        self.log.debug(
+                            "Reaped unknown child process (pid:%s, "
+                            "status:%s). Not a gunicorn worker.", wpid, status)
+                        continue
+
                     # A worker was terminated. If the termination reason was
                     # that it could not boot, we'll shut it down to avoid
                     # infinite start/stop cycles.
@@ -643,9 +655,6 @@ class Arbiter:
                         reason = "App failed to load."
                         raise HaltServer(reason, self.APP_LOAD_ERROR)
 
-                    worker = self.WORKERS.pop(wpid, None)
-                    if not worker:
-                        continue
                     worker.tmp.close()
                     self.cfg.child_exit(self, worker)
         except OSError as e:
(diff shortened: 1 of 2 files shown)


pajod:
I don't disagree, but I believe with very slightly more complexity, a more useful logging is possible.

> Non-worker child processes are logged at DEBUG instead of ERROR

Yes!

> No false alerts in monitoring systems

Yes!

> Unambigously explain the situation "Not a gunicorn worker"

Yes!

> skip os.WIFEXITED / os.WIFSIGNALED details from the (demoted to DEBUG level) log line

No! If someone bothered to put log level to DEBUG, then they clearly were interested. Likely the oprhaned child process was reaped by Gunicorn after something went wrong. Once Gunicorn clears it, the status is *gone* and may not be reported at a higher level. So even after demoting the log level for the following messages and clarifying that the messages are *not* about a Gunicorn worker, those messages should appear. Only the special treatment of Gunicorn-worker-specific exit codes should be skipped.

joho54:
Thanks for the feedback! I've added WIFEXITED/WIFSIGNALED logging for non-worker children reaping, while only skipping worker specific handling.  Please check [3e705a6](https://github.com/benoitc/gunicorn/pull/3566/commits/3e705a6f201ade66f4b716e32598b412daa8f39d)

benoitc:
Thanks for this, and for the production write up. The evidence you gathered is
what made it obvious the pid was never a worker, and moving the ownership check
ahead of the exit status logging is the right fix.

Two things came out of it beyond the log noise. The same ordering meant a
non-worker exiting with code 3 or 4 reached the HaltServer raises and shut the
whole server down. And the dirty arbiter is a third kind of child that is not in
WORKERS, so it needed its own branch rather than being treated as an unknown
child.

Merged in #3689 with those two additions and tests. Closing in favour of that one.

Issue #1299, opened 2016-06-25
Is there a way to set a maximum memory limit per worker


BernhardPosselt:
Hi, I'm parsing user uploaded tar.gz archives which contain XML files. I've got checks in place for maximum XML recursion limit but turns out adding and testing code to prevent a zip bomb is a lot harder than expected. 

My current solution is to start a sub process which launches a python script using ulimit which in turn extracts the xml file. It would be nice if I could just set a maximum memory limit per worker until it gets killed (e.g. 1GB)

I've looked into the available settings but could not find a solution.

diwu1989:
http://stackoverflow.com/questions/13622706/how-to-protect-myself-from-a-gzip-or-bzip2-bomb

benoitc:
@BernhardPosselt  you can limit the resources used by a worker using the [post_worker_init](http://docs.gunicorn.org/en/stable/settings.html?highlight=max_requests#post-worker-init) hook using some code like above. 

Such option can probably be added to gunicorn also, I need to think about it :)

BernhardPosselt:
@diwu1989 right, tried that already and could not get it to work, read through all the SO posts already and settled on something else which was hard to find out :) https://github.com/nextcloud/appstore/pull/56/files#diff-448835e9081cff9955382b5a4efce875R102

I hope my approach works, however it seems like such limits may make sense in general (or at least we should document the post_worker_init approach)

benoitc:
@BernhardPosselt  would you want to do that change?

BernhardPosselt:
Havent set up a running instance using Gunicorn yet and worked around the issue by reading the file in chunks. Since I didn't get the context manager stuff to work at all and I'm busy with other stuff I won't work on that issue :)

benoitc:
closing the activity. no activity since awhile. Feel free to open a new issue if needed.

Dobiasd:
Ok, I learned, it's probably not Gunicorn killing the worker (signal 9), it's `cgroups` doing so, and Gunicorn is just noticing (and logging) it. :bulb:

Thus, I implemented the following custom solution (`gunicorn.conf.py`) for my special use case:

```python3
#!/usr/bin/env python3
"""
For single-worker Gunicorn applications with cgroups-managed memory limit only! (E.g., in Kubernetes)

Kills the worker when the memory usage gets close to the limit.

[...]

Issue #124, opened 2010-11-17
 workers with high memory usage


ashokrkm:
i am having some workers suddenly using up over 1.5GB of RAM in my app. is it possible for gunicorn to monitor this and kill workers that use memory over something specified in the configuration, just like max_requests?

or is it possible to send some kind of signal or specify someway to gunicorn to kill that high memory using worker?

PR #3582, opened 2026-04-05
Add --max-worker-memory to recycle workers by RSS

### LINKED ISSUE #3581: Add --max-worker-memory option to recycle workers by RSS usage
## Summary

Add a `--max-worker-memory` configuration option that gracefully restarts workers when their resident set size (RSS) exceeds a configurable threshold in MB.

## Motivation

While `--max-requests` helps limit damage from memory leaks by recycling workers after a fixed number of requests, it cannot catch cases where a single request (or a small number of requests) causes disproportionate memory growth. A memory-based limit provides a direct safety net regardless of request patterns.

This has been discussed previously in #124 and #1299.

## Implementation

PR #3580 implements this feature:

- New `MaxWorkerMemory` config setting (int, default `0` = disabled, units: MB)
- `_get_rss()` reads `/proc/self/status` VmRSS on Linux, falls back to `resource.getrusage` on other platforms (stdlib only, no new dependencies)
- Memory check wired into all worker types (sync, gthread, gevent, eventlet, tornado, ASGI)
- Check is skipped entirely when the feature is disabled (zero overhead for existing deployments)

## Usage

```console
$ gunicorn app:app --max-worker-memory 512 --workers 4
```

Workers exceeding 512 MB RSS will be gracefully restarted after completing their current request.

hraju115:
## Summary

Adds a `--max-worker-memory` config option that gracefully restarts workers when their RSS exceeds a configurable threshold (in MB). This provides a hard safety net against memory leaks that `max_requests` alone cannot catch.

Closes #3581. Motivated by discussion in #124 and #1299.

## Changes

- **Config**: `MaxWorkerMemory` setting in `gunicorn/config.py` (int, default `0` = disabled, units: MB)
- **Core**: `_check_memory_usage()` and `_get_rss()` methods on `Worker` base class in `gunicorn/workers/base.py`
  - `_get_rss()` reads `/proc/self/status` `VmRSS` on Linux, falls back to `resource.getrusage` on other platforms (note: fallback measures peak RSS, documented in docstring)
  - Memory check skips entirely when feature is disabled (no `/proc` reads on the default hot path)
- **All worker types**: memory check wired into sync, gthread, async (gevent/eventlet), tornado, and ASGI protocol handlers — placed alongside existing `max_requests` checks
- **Docs**: settings reference regenerated via `build_settings_doc.py`
- **Tests**: 12 unit tests covering config parsing, `_get_rss()` (including `/proc` fallback), `_check_memory_usage()` logic, and tornado worker integration
- **Existing test mocks**: added `max_worker_memory` attribute to mock config objects in `test_handlers.py`, `test_server.py`, `test_asgi_protocol_compat.py`, `test_http2_alpn.py`

## Test plan

- [x] Full test suite passes: 1850 passed, 216 skipped, 0 failures
- [x] New tests in `tests/test_max_worker_memory.py` all pass
- [x] Manual validation: ran gunicorn with `--max-worker-memory 50 --workers 2` against a memory-leaking WSGI app; workers recycled correctly when RSS exceeded 50 MB

### Pre-existing test flakiness (not introduced by this PR)

When `tests/test_http2_alpn.py` runs before `tests/ctl/test_server.py` in the same pytest session, 9 `test_server` integration tests fail with socket connection errors. This is caused by eventlet's monkey-patching of the socket module during `TestEventletWorkerAlpn` — it corrupts socket state for subsequent tests. Verified by reproducing the same failures on unmodified `master` with `git stash && python -m pytest tests/test_http2_alpn.py tests/ctl/test_server.py`. The tests pass when run in isolation.

### DIFF
--- a/docs/content/reference/settings.md
+++ b/docs/content/reference/settings.md
@@ -1868,6 +1868,25 @@ restarts to avoid all workers restarting at the same time.
 
 !!! info "Added in 19.2"
 
+### `max_worker_memory`
+
+**Command line:** `--max-worker-memory INT`
+
+**Default:** `0`
+
+The maximum resident memory (in MB) a worker may use before being
+recycled.
+
+When a worker's RSS exceeds this value, it will be gracefully
+restarted after the current request completes. This provides a
+hard safety net against memory leaks that ``max_requests`` alone
+cannot catch.
+
+If this is set to zero (the default) then memory-based worker
+recycling is disabled.
+
+Motivated by discussion in GitHub issues #124 and #1299.
+
 ### `timeout`
 
 **Command line:** `-t INT`, `--timeout INT`
--- a/gunicorn/asgi/protocol.py
+++ b/gunicorn/asgi/protocol.py
@@ -747,6 +747,7 @@ class ASGIProtocol(asyncio.Protocol):
                     self.log.info("Autorestarting worker after current request.")
                     self.worker.alive = False
                     keepalive = False
+                self.worker._check_memory_usage()
 
                 if not keepalive or not self.worker.alive:
                     break
@@ -815,6 +816,7 @@ class ASGIProtocol(asyncio.Protocol):
                 self.log.info("Autorestarting worker after current request.")
                 self.worker.alive = False
                 keepalive = False
+            self.worker._check_memory_usage()
 
             if not keepalive or not self.worker.alive:
                 break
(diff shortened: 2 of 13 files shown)


benoitc:
Thank you for your PR. I will close it though. Memory limits are hard to handle. and can  be temporary. I think it would better to keep it on the observability side  rather than enforcing it there. 

A better operational path for such issue:

* trace your app to handle memory leaks
* observability stack: observe memory usage. Enevntually on thrshold, kill the worker. It's easily done today with scripting
* if you really want it inline, add a hook

PR #3588, opened 2026-04-13
fix: add header validation to early_hints callback

### LINKED ISSUE #3585: wsgi.early_hints callback has no header validation (response splitting risk)
## Summary
`gunicorn/http/wsgi.py:140-167` — the `wsgi.early_hints` callback constructs 103 Early Hints responses with zero header validation, while the main response path (`process_headers`) validates against `TOKEN_RE` and `HEADER_VALUE_RE`.

A WSGI app passing unsanitized data to early_hints could enable HTTP response splitting via CRLF injection.

## Suggested Fix
Apply the same `TOKEN_RE`/`HEADER_VALUE_RE` validation used in `process_headers` to the early_hints callback.

eddieran:
## Summary

Fixes #3585 — the `wsgi.early_hints` callback constructs 103 Early Hints responses with zero header validation, while `Response.process_headers` already validates header names against `TOKEN_RE` and header values against `HEADER_VALUE_RE`. This gap means a WSGI app passing unsanitized user input to `wsgi.early_hints` could enable HTTP response splitting via CRLF injection.

## What this does

Adds the same `TOKEN_RE.fullmatch(name)` / `HEADER_VALUE_RE.fullmatch(value)` checks (and the `value.strip(" \t")` normalization) from `process_headers` into `send_early_hints`. Raises the same `InvalidHeaderName` / `InvalidHeader` exceptions on failure.

## Addressing maintainer concerns

> Frameworks should do it. Maybe we should ensure to not do the work twice.

Agreed — frameworks *should* validate. But gunicorn already validates in `process_headers` for normal responses, so this PR simply brings `early_hints` to parity. It's defense-in-depth: if the framework misses it, the server catches it. The regex checks are fast and only run on the (typically small) early hints header list, so the cost is negligible.

## Test plan

- [x] Added tests for CRLF injection in header values → `InvalidHeader` raised, nothing written to socket
- [x] Added tests for CRLF injection in header names → `InvalidHeaderName` raised
- [x] Added tests for invalid token characters in header names → `InvalidHeaderName` raised
- [x] Verified valid headers still pass through correctly
- [x] All 11 existing + new early hints tests pass

### DIFF
--- a/gunicorn/http/wsgi.py
+++ b/gunicorn/http/wsgi.py
@@ -143,6 +143,10 @@ def _make_early_hints_callback(req, sock, resp):
         Args:
             headers: List of (name, value) header tuples, typically Link headers
                      Example: [('Link', '</style.css>; rel=preload; as=style')]
+
+        Raises:
+            InvalidHeaderName: If a header name is not a valid HTTP token.
+            InvalidHeader: If a header value contains invalid characters.
         """
         # Don't send after response has started - would break framing
         if resp.headers_sent:
@@ -159,6 +163,16 @@ def _make_early_hints_callback(req, sock, resp):
                 name = name.decode('latin-1')
             if isinstance(value, bytes):
                 value = value.decode('latin-1')
+
+            # Validate header name and value using the same checks as
+            # Response.process_headers — defense-in-depth against
+            # HTTP response splitting via CRLF injection.
+            if not TOKEN_RE.fullmatch(name):
+                raise InvalidHeaderName('%r' % name)
+            if not HEADER_VALUE_RE.fullmatch(value):
+                raise InvalidHeader('%r' % value)
+
+            value = value.strip(" \t")
             response += f"{name}: {value}\r\n".encode('latin-1')
         response += b"\r\n"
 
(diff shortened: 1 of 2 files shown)


gunicorn/http/wsgi.py
```diff
@@ -159,6 +163,16 @@ def send_early_hints(headers):
...
             if isinstance(value, bytes):
                 value = value.decode('latin-1')
+
+            # Validate header name and value using the same checks as
+            # Response.process_headers — defense-in-depth against
+            # HTTP response splitting via CRLF injection.
+            if not TOKEN_RE.fullmatch(name):
+                raise InvalidHeaderName('%r' % name)
+            if not HEADER_VALUE_RE.fullmatch(value):
+                raise InvalidHeader('%r' % value)
```
pajod: Until error handling receives some major cleanup (preferably [class inheritance instead of massive elif](https://github.com/benoitc/gunicorn/blob/9aa54703f4950818aed538dbee9578e868375cc9/gunicorn/workers/base.py#L216-L255)) that exception class should be instantiated with just the name, not the value. Otherwise any way of triggering the exception on purpose can be used move sensitive information between security boundaries. Browsers are picky about what headers can be accessed cross-site, and proxies sitting between Gunicorn and the client might pick which ones to forward. Best not to interfere with that.

eddieran:
Addressed, @pajod — thanks for catching the cross-boundary leak. Changed `raise InvalidHeader('%r' % value)` → `raise InvalidHeader('%r' % name)` so the value never appears in the exception. Agreed that the existing `process_headers` on line 350 has the same pattern, but since you flagged that it's pending the broader error-handling cleanup, I left it alone in this PR.

PR #3609, opened 2026-04-26
Honor forwarded_allow_ips in fast parser

script3r:
## Summary
- make the fast HTTP parser honor `forwarded_allow_ips` before accepting underscore `forwarder_headers`
- align fast-parser behavior with the existing Python parser
- add a regression test covering trusted and untrusted peers under both parser modes

## Testing
- `uv run --extra testing --extra fast pytest tests/test_http_forwarder_headers.py tests/test_valid_requests.py tests/test_invalid_requests.py`

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -206,6 +206,16 @@ class Message:
     def parse(self, unreader):
         raise NotImplementedError()
 
+    def _source_is_trusted_proxy(self):
+        return (
+            not isinstance(self.peer_addr, tuple)
+            or _ip_in_allow_list(
+                self.peer_addr[0],
+                self.cfg.forwarded_allow_ips,
+                self.cfg.forwarded_allow_networks(),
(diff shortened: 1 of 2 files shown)


benoitc:
Thanks for catching this. The fast parser accepting underscore forwarder headers
without checking `forwarded_allow_ips` was a real gap.

It was fixed on master in e90b1c2c, which landed after you opened this. Both
parsers now go through `_peer_trusted_for_forwarded()`, which returns empty
policy for an untrusted peer, and the parity tests you proposed exist in
`tests/test_header_policy_parity.py` as
`test_untrusted_peer_underscore_header_rejected` and
`test_trusted_peer_underscore_header_accepted`, both parametrized over the python
and fast parsers. I ran them with gunicorn_h1c installed to confirm the fast
variants pass.

Closing as already fixed. Sorry it sat here long enough for that to happen.

PR #3615, opened 2026-05-03
fix: tighten keepalive gate and scope finish_body byte cap

benoitc:
Stacked on #3614. Two regressions surfaced when reviewing that PR:

- ASGI keepalive gate keyed on `receiver._closed`, which is now also set on receive timeout. Treating timeout as a framed-complete message re-opens the smuggling vector. Gate now keys on `_complete` only; transport-disconnect cases are already handled by the existing `self._closed` / `keepalive=False` checks.
- `Parser.finish_body`'s 64 KiB byte cap fired on every call, including the default invocation from `__next__`. Non-gthread workers (base_async, sync) ignore the return value and were silently desyncing on long bodies. Cap now applies only when an explicit deadline is given.

### DIFF
--- a/gunicorn/asgi/protocol.py
+++ b/gunicorn/asgi/protocol.py
@@ -782,15 +782,13 @@ class ASGIProtocol(asyncio.Protocol):
                     break
 
                 # Refuse keepalive if the previous request body was not fully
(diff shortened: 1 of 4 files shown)


benoitc:
Folding into #3614 directly per review.

PR #3643, opened 2026-06-05
chore: minimise dependency on the Python package named 'packaging'

jayaddison:
The Python package named `packaging` is only required as a runtime dependency by the `gevent` worker of `gunicorn`; in other words, we do not need to install it as a dependency for installations of other worker types.

This changeset relocates the `packaging` requirement from the default dependencies for `gunicorn` to the `gevent` extras set (and also the `requirements_test.txt` requirements file, so that it is available when the unit tests run).

Resolves #3608 (discussion thread).

PR #3644, opened 2026-06-08
Cleanup: remove unused socket variable

jayaddison:
Initially, remove an unused `sock` variable reported in a CI failure; ref: https://github.com/benoitc/gunicorn/pull/3643#issuecomment-4653712487

..then notice that that tuple element isn't used in the other callsites, so it can be omitted from the helper function return values.


jayaddison:
This `sock` variable _is_ now used elsewhere in `test_http.py`.  Closing this as not-for-merge.

PR #3663, opened 2026-07-07
feat: opt-in prior-knowledge h2c support for the gthread worker

### LINKED ISSUE #3489: h2c support

fluffy-dev:
Closes #3489

Makes #3489 concrete, as asked in the thread - small enough to judge the actual complexity.

HTTP/2 currently only engages over TLS via ALPN. The h2 machinery itself doesn't care about the transport, so this adds the missing cleartext way in:

- New `--http2-prior-knowledge` flag, **off by default**. Only active when `h2` is in `--http-protocols` and no TLS is configured. With TLS, ALPN keeps doing its thing, untouched.
- In the gthread worker, the preface is only sniffed for peers trusted per `forwarded_allow_ips` (i.e. the TLS-terminating proxy; unix socket peers count as trusted, same semantics as the forwarded-header policy). A trusted peer whose first bytes match the HTTP/2 connection preface (`PRI * HTTP/2.0...`) goes down the existing `handle_http2` path. Anything else from a trusted peer (an h1 request, a malformed or stalled preface) gets a 400 - on a prior-knowledge port the proxy is expected to speak h2, and a silent downgrade would only hide misconfiguration. Peers outside the allow list are served HTTP/1.1 byte-for-byte as if the flag were off, no sniff and no preface deadline.
- Prior-knowledge only (RFC 9113 §3.4). The old `Upgrade: h2c` dance was deprecated by RFC 9113, so it's deliberately not implemented - which also sidesteps most of the historical security concern; the docs mark this as trusted-proxy-only either way.

Target use case is the one from the thread: platforms and meshes that terminate TLS and talk h2c upstream (Cloud Run, fly.io, Istio sidecars - @frittentheke's case). Without this, that hop silently drops to HTTP/1.1.

Verified end-to-end: `curl --http2-prior-knowledge` gets HTTP/2 with the flag on, a plain h1 curl from a trusted peer gets a 400 with a warning in the log, and with the flag off a preface is rejected exactly as today. 14 tests in `tests/test_http2_h2c.py` covering the trusted 400 paths, untrusted passthrough, `*` allow list, and unix socket peers; full suite passes.

Scope is gthread only for now - if the approach is acceptable the async workers can follow the same pattern in a follow-up.

@benoitc the docs currently say h2c is intentionally unsupported, so this stays a draft until you weigh in. If opt-in prior-knowledge is still not something you want in core, happy to close.

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -37,6 +37,14 @@ _DEFER = object()
 # the main poller to prevent thread pool exhaustion from slow clients.
 DEFAULT_WORKER_DATA_TIMEOUT = 5.0
 
+# HTTP/2 connection preface sent by clients using prior-knowledge h2c
+# (RFC 9113 section 3.4).
+H2C_PREFACE = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n"
+
+# How long (in seconds) to wait for the full connection preface once the
+# first bytes of a candidate h2c connection have arrived.
+H2C_PREFACE_TIMEOUT = 1.0
+
 
 class TConn:
 
@@ -82,8 +90,61 @@ class TConn:
                     self.parser.initiate_connection()
                     return
 
+            elif (
+                self.cfg.h2c
+                and "h2" in self.cfg.http_protocols
+                and getattr(self.cfg, "protocol", "http") == "http"
+            ):
+                # HTTP/2 cleartext (h2c) with prior knowledge (RFC 9113
+                # section 3.4): the client opens a plain TCP connection and
+                # immediately sends the HTTP/2 connection preface.
+                matched, buf = self._read_h2c_preface()
+                if matched:
+                    self.is_http2 = True
+                    self.parser = http.get_parser(
+                        self.cfg, self.sock, self.client, http2_connection=True
+                    )
+                    self.parser.initiate_connection()
+                    # The consumed preface bytes still need to reach the
+                    # HTTP/2 state machine. The preface alone cannot
+                    # complete a request, so nothing is dropped here.
+                    self.parser.receive_data(buf)
+                    return
+                if buf:
+                    # Not an HTTP/2 preface: serve as HTTP/1.x, returning
+                    # the consumed bytes to the parser.
+                    self.parser = http.get_parser(self.cfg, self.sock, self.client)
+                    self.parser.unreader.unread(buf)
+                    return
+
             # initialize the HTTP/1.x parser
             self.parser = http.get_parser(self.cfg, self.sock, self.client)
+
+    def _read_h2c_preface(self):
+        """Read up to the length of the HTTP/2 connection preface.
+
+        Consumes bytes from the socket until the preface is complete, a
+        byte diverges from it, the peer closes, or the preface timeout
+        expires. Returns a tuple ``(matched, consumed_bytes)``; on a
+        non-match the consumed bytes must be handed back to the HTTP/1.x
+        parser by the caller.
+        """
+        buf = b""
+        self.sock.settimeout(H2C_PREFACE_TIMEOUT)
+        try:
+            while len(buf) < len(H2C_PREFACE):
+                try:
+                    chunk = self.sock.recv(len(H2C_PREFACE) - len(buf))
+                except TimeoutError:
+                    return False, buf
+                if not chunk:
+                    return False, buf
+                buf += chunk
+                if not H2C_PREFACE.startswith(buf):
+                    return False, buf
+            return True, buf
+        finally:
+            self.sock.settimeout(None)
 
     def set_timeout(self):
         # Use monotonic clock for reliability (time.time() can jump due to NTP)
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -27,6 +27,8 @@ from .. import http
 from .. import util
 from .. import sock
 from ..http import wsgi
+from ..http.errors import InvalidH2CPreface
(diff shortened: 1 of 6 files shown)


frittentheke:
Thanks @fluffy-dev for putting this together!

1. You might want to link this PR to the issue via https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue 
2. Why do you consider this one still a draft?

gunicorn/workers/gthread.py
```diff
@@ -82,9 +90,62 @@ def init(self):
...
+                    # HTTP/2 state machine. The preface alone cannot
+                    # complete a request, so nothing is dropped here.
+                    self.parser.receive_data(buf)
+                    return
+                if buf:
+                    # Not an HTTP/2 preface: serve as HTTP/1.x, returning
+                    # the consumed bytes to the parser.
+                    self.parser = http.get_parser(self.cfg, self.sock, self.client)
+                    self.parser.unreader.unread(buf)
+                    return
```
pajod: Is this merely the most compact method of ensuring error handling is *somewhat* sane in that branch?

.. Or do you know that someone (some software) would actually do that and/or would like to actually support the (deprecated) case where "prior knowledge" was configured on the Gunicorn side - yet the proxy tries (and fails) to negotiate via 101 anyway?

(If it is the former, then making `--http2-prior-knowledge` deliberately not parse h2c unless `_peer_trusted_for_forwarded()` *and* unconditionally responding 400 for http/1.x (plus any malformed/refused prefaces) would limit unforeseen interaction with other features/timeouts, while keeping the limited case that is actually wanted intact.)

fluffy-dev: Yeah, it was the former. Nobody I know of actually configures prior knowledge on the gunicorn side and then has the proxy try a 101 anyway, the fallback was just the cheapest way to keep that branch sane.

Your version is better, so that's what it does now. The preface only gets sniffed when the peer passes the `forwarded_allow_ips` check (same semantics as `_peer_trusted_for_forwarded()`, unix sockets trusted), and a trusted peer that sends anything other than a valid preface gets a plain 400 instead of being downgraded. Everyone outside the allow list lands on the old h1 path untouched, no sniff and no preface deadline, so the timeout interaction goes away too. I also renamed the flag to `--http2-prior-knowledge` while at it, since that's all it ever did.

The 101 dance stays unimplemented. Added tests for the 400 paths, the untrusted passthrough, `*` in the allow list, and unix socket peers.

fluffy-dev:
Done, added `Closes #3489` to the description.

It's a draft because the docs currently say h2c is unsupported on purpose, so I didn't want to push it as ready before benoitc says whether he wants this in core at all. Code-wise it's complete and tested, I'll flip it to ready if the direction gets a nod.

gunicorn/workers/gthread.py
```diff
@@ -82,9 +92,78 @@ def init(self):
...
+                # A trusted peer on a prior-knowledge port must speak
+                # HTTP/2. Anything else (an HTTP/1.x request, a malformed
+                # or stalled preface) is rejected with a 400 rather than
+                # silently downgraded.
+                raise InvalidH2CPreface(buf)
+
             # initialize the HTTP/1.x parser
             self.parser = http.get_parser(self.cfg, self.sock, self.client)
 
+    def _peer_trusted_for_h2c(self):
```
jeffsawatzky, line 127: I'm no expert on gunicorn, but some of this seems like it would be useful in other workers as well, such as the ASGI worker (where http2 would be really useful). Would it make sense to move some of it into the base worker class?

fluffy-dev: Right, most of the pieces don't care about the worker at all - the preface constant, the trust check, the "is h2c allowed here" predicate. Only the sniff-then-hand-off is gthread specific. I kept it all in gthread on purpose so the diff stays small while the direction is still undecided (the docs said h2c was unsupported on purpose, so this may get rejected outright). If it gets a nod, the plan is to pull the shared bits into a common helper and wire up the async workers in a follow-up. Agreed asgi is where h2 pays off most.

fluffy-dev:
Marking this ready: pajod's points are addressed (trust-gated sniffing, 400 instead of a silent downgrade, flag renamed to `--http2-prior-knowledge`), the suite is green, and being a draft was keeping CI switched off. @benoitc if opt-in prior-knowledge h2c is still something you don't want in core, just say so and I'll close - but if the direction is fine, this should be reviewable as is.

benoitc:
superseded and merged in  #3706 . Thanks a lot for the PR!

PR #3664, opened 2026-07-08
asgi: decrement nr_conns on server-initiated connection close (#3661)

### LINKED ISSUE #3661: ASGI worker: nr_conns never decremented on server-initiated close — every graceful stop waits out the full graceful_timeout
**Environment**

- gunicorn 26.0.0 (the code path is unchanged in current master)
- CPython 3.12, `--worker-class asgi`, plain HTTP/1.x
- `gunicorn_h1c` not installed (pure-Python callback parser path)

**Symptom**

Every worker stop, both SIGTERM and the `--max-requests` autorestart, blocks for the full `graceful_timeout` and then logs `Forcing close of N connections`, where N keeps growing with worker age. It happens even when the worker is completely idle.

Production example (3 workers, `--max-requests 1000 --max-requests-jitter 500`, behind an nginx that proxies HTTP/1.0 close-per-request), captured at a zero-traffic hour:

```
06:07:18 [132970] [INFO] Worker shutting down...
06:07:18 [132970] [INFO] Waiting for 1426 connections to finish...
06:07:28 [146478] [INFO] Waiting for 347 connections to finish...
06:07:28 [134827] [INFO] Waiting for 1350 connections to finish...
06:09:13 [132970] [WARNING] Forcing close of 1426 connections     <- 115s later, same number
06:09:13 [132970] [INFO] Worker exiting (pid: 132970)
```

The counts match each worker's requests served since boot (1426 and 1350 are inside the 1000 to 1500 max-requests band; the worker logging 1426 was autorestarting for having reached it). They are not real connections: the count never moves during the whole graceful window.

**Cause**

In `gunicorn/asgi/protocol.py`:

- `connection_made()` does `self.worker.nr_conns += 1`.
- A server-initiated close goes through `_close_transport()`, which calls `transport.close()` and sets `self._closed = True`. It does not decrement.
- When asyncio then delivers `connection_lost()`, the idempotence guard returns before the decrement:

```python
def connection_lost(self, exc):
    # Guard against multiple calls (idempotent)
    if self._closed:
        return

    self._closed = True
    self.worker.nr_conns -= 1
```

So every connection the server closes first leaks one count permanently. That covers any `Connection: close` response (for example HTTP/1.0 reverse proxies), every keepalive timeout (`_keepalive_timeout()` calls `_close_transport()`), and error aborts. Client-initiated closes are balanced because `connection_lost()` runs first, which is presumably why client-driven test suites do not catch it.

`ASGIWorker._shutdown()` (`gunicorn/workers/gasgi.py`) then waits on that counter:

```python
if self.nr_conns > 0 and not self._quick_shutdown:
    graceful_timeout = self.cfg.graceful_timeout
    self.log.info("Waiting for %d connections to finish...", self.nr_conns)
    deadline = self.loop.time() + graceful_timeout
    while self.nr_conns > 0 and self.loop.time() < deadline:
        ...
```

Since the counter can never reach zero, every graceful stop runs to the deadline. Consequences: deploys that SIGTERM workers always take the full `graceful_timeout`, and each `--max-requests` recycle keeps a worker out of rotation (not accepting) for that whole window, which can starve the listen socket when several workers recycle close together.

**Minimal repro (protocol level, no sockets)**

```python
import asyncio
import logging
import types

from gunicorn.asgi.protocol import ASGIProtocol
from gunicorn.config import Config


class FakeTransport:
    def get_extra_info(self, name, default=None):
        return default  # no ssl_object: plain HTTP/1.x path

    def set_write_buffer_limits(self, high=None, low=None):
        pass

    def can_write_eof(self):
        return False

    def close(self):
        pass

    def is_closing(self):
        return True


def make_worker():
    return types.SimpleNamespace(
        cfg=Config(),
        loop=asyncio.get_running_loop(),
        nr_conns=0,
        alive=True,
        log=logging.getLogger("repro"),
        asgi=None,
    )


async def main():
    # Server closes first (Connection: close response, keepalive timeout, ...)
    worker = make_worker()
    proto = ASGIProtocol(worker)
    proto.connection_made(FakeTransport())
    proto._close_transport()      # sets _closed = True
    proto.connection_lost(None)   # guard returns early: no decrement
    print("server-initiated close, nr_conns =", worker.nr_conns)  # 1  (leak)

    # Client closes first: balanced
    worker = make_worker()
    proto = ASGIProtocol(worker)
    proto.connection_made(FakeTransport())
    proto.connection_lost(None)
    proto._close_transport()
    print("client-initiated close, nr_conns =", worker.nr_conns)  # 0


asyncio.run(main())
```

Output on 26.0.0:

```
server-initiated close, nr_conns = 1
client-initiated close, nr_conns = 0
```

**End-to-end repro**

```python
# app.py
async def app(scope, receive, send):
    if scope["type"] != "http":
        return
    await send({"type": "http.response.start", "status": 200, "headers": []})
    await send({"type": "http.response.body", "body": b"ok"})
```

```console
$ gunicorn app:app --worker-class asgi --workers 1 --bind 127.0.0.1:8123 --graceful-timeout 115 &
$ for i in 1 2 3; do curl -s --http1.0 http://127.0.0.1:8123/ >/dev/null; done
$ kill -TERM %1
```

The worker logs `Waiting for 3 connections to finish...` with no client connected and exits only when the 115s deadline expires. With the decrement fixed (we currently carry the wrapper below in our `gunicorn.conf.py`), the same sequence exits in about 2 seconds.

**Workaround / suggested fix**

Perform the decrement wherever `_closed` flips, so each connection is decremented exactly once whichever side closes first. As a config-file workaround:

```python
from gunicorn.asgi.protocol import ASGIProtocol

_orig = ASGIProtocol._close_transport

def _close_transport(self):
    was_closed = self._closed
    _orig(self)
    if self._closed and not was_closed:
        self.worker.nr_conns -= 1

ASGIProtocol._close_transport = _close_transport
```

A proper fix could decrement in `_close_transport()` itself, or centralize the close accounting in a single helper used by both paths. Similar accounting bug fixed long ago in the gthread worker: #1387.

Happy to open a PR.

apoorvdarshan:
## Summary

Fixes #3661. On the ASGI worker, `nr_conns` is never decremented for **server-initiated** connection closes, so it grows with worker age and every graceful stop (`SIGTERM`, `--max-requests` autorestart) blocks for the full `graceful_timeout` — even on an idle worker. This hits any setup with `Connection: close` / HTTP/1.0-per-request (e.g. nginx proxying to the worker).

## Root cause

`ASGIProtocol` used a single `_closed` flag for two different jobs:

- `_close_transport()` sets `_closed = True` to mark the transport closed;
- `connection_lost()` used `if self._closed: return` as its idempotency guard.

A **client**-initiated close calls `connection_lost()` directly (`_closed` is `False`) → the count is decremented. A **server**-initiated close runs `_close_transport()` first, setting `_closed = True`; when asyncio then calls `connection_lost()`, the guard returns early and skips `self.worker.nr_conns -= 1` (and the rest of the cleanup).

## Fix

Introduce a dedicated `_conn_lost_handled` flag for `connection_lost()`'s idempotency guard, independent of the transport's `_closed` state. The cleanup (decrementing `nr_conns`, disconnect signalling, task cancellation) now runs exactly once regardless of who initiated the close.

## Tests

- Added `test_server_initiated_close_still_decrements_nr_conns`: `_close_transport()` then `connection_lost()` must leave `nr_conns == 0`. It fails on `master` (stays `1`) and passes here.
- `tests/test_asgi_disconnect.py` (14) and `tests/test_asgi_worker.py` (32) pass; the full ASGI suite is green (628 passed, 121 skipped); `pycodestyle` clean.

Happy to add a changelog entry if you'd like one.

*Disclosure: prepared with AI assistance; reviewed and verified locally.*

### DIFF
--- a/gunicorn/asgi/protocol.py
+++ b/gunicorn/asgi/protocol.py
@@ -362,6 +362,11 @@ class ASGIProtocol(asyncio.Protocol):
 
         # Connection state
         self._closed = False
+        # Guards the connection_lost() cleanup so it runs exactly once. Kept
+        # separate from ``_closed`` because a server-initiated close sets
+        # ``_closed`` early (in _close_transport), and connection_lost() must
+        # still run its cleanup (e.g. decrement nr_conns) afterwards.
+        self._conn_lost_handled = False
         self._body_receiver = None  # Set per-request for disconnect signaling
 
         # Response buffering for write batching
@@ -679,10 +684,14 @@ class ASGIProtocol(asyncio.Protocol):
 
         See: https://github.com/benoitc/gunicorn/issues/3484
         """
-        # Guard against multiple calls (idempotent)
-        if self._closed:
+        # Guard against multiple calls (idempotent). Use a dedicated flag rather
+        # than ``_closed``: a server-initiated close sets ``_closed`` in
+        # _close_transport() before the transport reports the loss, and the
+        # cleanup below (notably decrementing nr_conns) must still run.
+        if self._conn_lost_handled:
             return
 
+        self._conn_lost_handled = True
         self._closed = True
         self.worker.nr_conns -= 1
 
--- a/tests/test_asgi_disconnect.py
+++ b/tests/test_asgi_disconnect.py
@@ -103,6 +103,28 @@ class TestASGIGracefulDisconnect:
         assert mock_worker.nr_conns == 1  # Should not decrement again
         # Closed flag is still set
 
+    def test_server_initiated_close_still_decrements_nr_conns(self, mock_worker):
+        """A server-initiated close must not leak the connection count (#3661).
+
+        _close_transport() sets ``_closed`` before asyncio reports the transport
+        loss, so connection_lost() must still run its cleanup and decrement
+        nr_conns. Otherwise every graceful stop waits out the full timeout.
+        """
+        protocol = ASGIProtocol(mock_worker)
+        protocol.reader = mock.Mock()
+        protocol.transport = mock.Mock()
+        mock_worker.nr_conns = 1
+
+        # Server closes the connection (e.g. `Connection: close`, keepalive
+        # timeout), which sets ``_closed`` early.
+        protocol._close_transport()
+        assert protocol._closed is True
+        assert mock_worker.nr_conns == 1
+
+        # asyncio then reports the transport loss.
+        protocol.connection_lost(None)
+        assert mock_worker.nr_conns == 0
+
     def test_disconnect_does_not_cancel_immediately(self, mock_worker):
         """Test that connection_lost doesn't cancel task immediately."""
         protocol = ASGIProtocol(mock_worker)


benoitc:
Reopening to trigger CI, which never fired on this PR. Thanks for the fix, the approach here is the right one.

benoitc:
Thanks, your analysis and fix were the right ones. Superseded by #3686, which carries your commit over unchanged and still under your authorship, and adds coverage for three gaps: the keepalive close path leaking as well, the cleanup steps the early return also skipped (keepalive timer, reader EOF, body receiver disconnect signal), and the reported symptom itself, that `_shutdown()` burns the full `graceful_timeout`.

Closing here in favour of #3686.

PR #3667, opened 2026-07-21
reloader: treat empty dirname as cwd for inotify watches

### LINKED ISSUE #3377: Reload extra files in the current directory with inotify
It seems that adding an extra watch file in the current directory, like `.env`, will cause inotify to crash because the `os.path.dirname(".env") == ""`. Using `./.env` works though (dirname is then `.`) or an absolute path works too.

https://github.com/benoitc/gunicorn/blob/a86ea1e4e6c271d1cd1823c7e14490123f9238fe/gunicorn/reloader.py#L84-L91

Curious what you think? If this is something that should be allowed, then I wonder if you could just do something like this to protect against it:

```python
if not dirname:
    dirname = "."
```

Thanks!

alexchen-sys:
#### Description
Watching an extra file in the current directory (for example `.env`) makes `os.path.dirname(filename)` return an empty string. Passing that to inotify crashes.

Normalize empty dirname to `.` so cwd-relative extra files work, matching absolute/`./.env` paths that already work.

Closes #3377

#### Checklist
- [x] `pytest tests/test_reload.py`

### DIFF
--- a/gunicorn/reloader.py
+++ b/gunicorn/reloader.py
@@ -81,7 +81,7 @@ if has_inotify:
         def add_extra_file(self, filename):
             super().add_extra_file(filename)
 
-            dirname = os.path.dirname(filename)
+            dirname = os.path.dirname(filename) or '.'
             if dirname in self._dirs:
                 return
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ testing = [
     "pytest-asyncio",
     "uvloop>=0.19.0",
     "httpx[http2]",
+    "inotify>=0.2.10; sys_platform == 'linux'",
 ]
 
 [project.scripts]
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -6,3 +6,4 @@ pytest-asyncio
 gunicorn_h1c>=0.6.5
 h2>=4.1.0
 uvloop>=0.19.0
+inotify>=0.2.10; sys_platform == 'linux'
--- a/tests/test_reload.py
+++ b/tests/test_reload.py
@@ -4,6 +4,8 @@
 
 import unittest.mock as mock
 
+import pytest
+
 from gunicorn.app.base import Application
 from gunicorn.workers.base import Worker
 from gunicorn.reloader import reloader_engines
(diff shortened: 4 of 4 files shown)


yangfan-yf-yf (changes requested):
The production change is appropriately narrow, but the regression does not reliably exercise it. `tests/test_reload.py` returns successfully whenever `has_inotify` is false, while the documented `testing` extra does not include the `inotify` package. In those environments this test is a silent no-op, so the affected branch receives no regression coverage.

Please make the test establish the inotify path explicitly (or make the test dependency and skip condition explicit) so the assertion is guaranteed to run in the supported test environment. `real_add_watch` is also assigned but unused and should be removed.

alexchen-sys:
Thanks for the careful review — both points are addressed.

The test now starts with `pytest.importorskip("inotify", reason="inotify reloader requires the inotify package")`, so the assertion runs wherever the dependency is installed, and on the `has_inotify=False` lane it reports a visible `pytest.skip` instead of a silent pass. The unused `real_add_watch` is removed.

To make that dependency present, `inotify>=0.2.10; sys_platform == 'linux'` is added to both `requirements_test.txt` and the `testing` extra, so Linux CI installs it while macOS/FreeBSD skip it. On `7823ebf` `pytest tests/test_reload.py --no-cov` gives 3 passed with it installed, 2 passed + 1 skipped without. Squashed into one commit.

yangfan-yf-yf (approved):
The follow-up addresses the requested test reliability issues. The test dependency is now platform-guarded in both `requirements_test.txt` and the `testing` extra, so Linux test environments install it while other platforms receive an explicit skip. The unused watcher assignment is gone, and the fake watcher asserts that a cwd-relative extra file is registered against `.`.

LGTM.

benoitc:
merged. Thanks!

benoitc:
Thanks! Validated on Linux: on master `add_extra_file('.env')` raises `InotifyError ENOENT`, and with this change the directory is watched as `.` and the reload callback fires. Other path shapes (`sub/a.json`, absolute, `./rel.txt`) are unchanged.

PR #3673, opened 2026-07-29
fix: catch BaseException in gthread worker

ipr-ams:
All exceptions in Python must inherit from `BaseException` and since connections should always be closed it makes sense to catch `BaseException` rather than `Exception`.

### DIFF
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -439,7 +439,7 @@ class ThreadWorker(base.Worker):
             else:
                 self.nr_conns -= 1
                 conn.close(graceful=True)
-        except Exception:
+        except BaseException:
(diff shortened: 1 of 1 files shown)


yangfan-yf-yf (changes requested):
Catching `BaseException` here also catches control-flow exceptions such as `KeyboardInterrupt` and `SystemExit`. Because this handler closes the connection and then returns normally, those exceptions no longer reach the worker lifecycle, which can prevent the requested interruption or process exit from taking effect.

Please either retain the existing `Exception` boundary or perform the connection cleanup and re-raise `BaseException`. A focused regression should verify that a future raising `KeyboardInterrupt` (and ideally `SystemExit`) still escapes `finish_request` after cleanup.

PR #3676, opened 2026-08-03
fix: send zero-valued StatsD metrics

jgorostegui:
Discussed in #3675.

`Statsd.log()` checks metric values for truthiness, so a valid zero is treated like a missing value. The backlog metric makes this visible on an idle server.

This changes the check to `value is not None` and adds regression tests for zero-valued gauges, counters, histograms and timers. `None` stays ignored and positive values behave as before.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -97,6 +97,7 @@ Jonas Haag <jonas@lophus.org>
 Jonas Nockert <jonasnockert@gmail.com>
 Jorge Niedbalski <jorge@nimbic.com>
 Jorge Niedbalski R <niedbalski@gmail.com>
+Josu Gorostegui <gorosjosu@gmail.com>
 Justin Quick <justquick@gmail.com>
 keakon <keakon@gmail.com>
 Keegan Carruthers-Smith <keegan.csmith@gmail.com>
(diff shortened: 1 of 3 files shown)


yangfan-yf-yf (approved):
The old truthiness guard treated a valid zero as missing. Checking specifically for `None` preserves the existing checks for metric and type while allowing zero-valued gauges, counters, histograms, and timers to reach their existing formatters.

The regression coverage exercises each metric type at zero and keeps the no-output behavior for `None`.

LGTM.

benoitc:
merged. Thanks!

benoitc:
Thanks! Confirmed master drops every zero value across gauge, counter, histogram and timer (both `0` and `0.0`), and that `None`, an empty metric name and a `None` type stay dropped.

jgorostegui:
> Thanks! Confirmed master drops every zero value across gauge, counter, histogram and timer (both `0` and `0.0`), and that `None`, an empty metric name and a `None` type stay dropped.

Many thanks, @benoitc. Just for context, we use `gunicorn.backlog` in an autoscaling loop. A reported zero means the queue is empty; no report means we don't know. Gunicorn used to collapse those two states, so safe scale-in needed a downstream workaround. Merging this fixes that at the source ^^

Thanks again.

PR #3678, opened 2026-08-06
Split log arguments from log message in arbiter

ds-cbo:
Hi, in our logging application we group messages based on the log message, but because the `wpid` is directly formatted in the string here and it keeps changing, we cannot group them anymore. This fixes that. In general it also feels more correct to do it this way and it's done more often in this code base, for example lines 636/637 below here as well.


benoitc:
Thanks! Checked the rendered output for SIGKILL, SIGTERM and SIGSEGV against both the default logger and the statsd logger: messages and levels are identical, including the out-of-memory suffix.

PR #3683, opened 2026-08-13
fix(example): replace bare except with except Exception in gevent websocket example

TrueFurina:


benoitc:
Thanks! The example is excluded from collection via `norecursedirs`, and `b64decode` raises `binascii.Error`, so the diagnostic still prints for the case it was meant for.

PR #3703, opened 2026-08-19
docs: the control socket is not created in the current working directory

cormier:
The markdown docs states that the control socket is created in the current working directory, but the control_socket Setting docs states that the control socket is created in XDG_RUNTIME_DIR or the home directory if XDG_RUNTIME_DIR is unset or not a directory.

This PR fixes this issue by making the markdown docs consistent with the settings documentation


benoitc:
thanks! merged.

PR #3714, opened 2026-08-25
uwsgi: fix response handling

methane:
Strip Transfer-Encoding headers when an app is served through the uWSGI protocol, so chunk markers are not forwarded as part of the application payload. This keeps the body framing consistent with the raw-byte uWSGI response semantics and avoids leaking chunk encoding to clients behind nginx.

Also update the close-response logic and add regression coverage for the uWSGI integration and behavior.

fix: https://github.com/benoitc/gunicorn/discussions/3697


benoitc:
Thanks @methane, merged.

PR #3715, opened 2026-08-29
Stop writing a 500 on top of a response that already started

### LINKED ISSUE #3410: Gunicorn injects "HTTP/1.1 500 Internal Server Error" into body when worker is terminated while streaming response
Hi,

one of users tried to download a file via service based on Flask, gunicorn and sync workers. He was using a very slow connection and file was pretty big, so connection was interrupted after 60 seconds due to worker timeout. Endpoint was supporting Range header, so wget made a new request for a missing part of the file and download was finished.

Unfortunately, the file appeared to be incorrect. After inspection, we noticed that ISE 500 headers and body were mixed into content:

Image

This situation can be easily reproduced using that code:

```python
import time
from flask import Flask, Response

app = Flask(__name__)

def stream_endlessly():
    while True:
        yield b'a'*16 + b'\n'
        time.sleep(1)

@app.route('/')
def index():
    return Response(stream_endlessly(), mimetype='text/plain')
```

If we spawn it using `gunicorn app:app -t 10` and make a request `curl http://127.0.0.1:8000 --http1.0 -vvv` (using HTTP 1.0 to avoid chunked encoding), we will see the following output:

```
$ curl http://127.0.0.1:8000 --http1.0 -vvv
*   Trying 127.0.0.1:8000...
* Connected to 127.0.0.1 (127.0.0.1) port 8000
> GET / HTTP/1.0
> Host: 127.0.0.1:8000
> User-Agent: curl/8.5.0
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Server: gunicorn
< Date: Tue, 12 Aug 2025 18:51:19 GMT
< Connection: close
< Content-Type: text/plain; charset=utf-8
< 
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
aaaaaaaaaaaaaaaa
HTTP/1.1 500 Internal Server Error
Connection: close
Content-Type: text/html
Content-Length: 141

<html>
  <head>
    <title>Internal Server Error</title>
  </head>
  <body>
    <h1><p>Internal Server Error</p></h1>
    
  </body>
</html>
* Closing connection
```

The root cause is well described by the stack trace:

```
[2025-08-12 20:51:35 +0200] [46057] [CRITICAL] WORKER TIMEOUT (pid:46085)
[2025-08-12 20:51:35 +0200] [46085] [ERROR] Error handling request /
  File "/home/psrok1/gunicorn-bug-20250812/.venv/lib/python3.12/site-packages/gunicorn/workers/sync.py", line 134, in handle
    self.handle_request(listener, req, client, addr)
  File "/home/psrok1/gunicorn-bug-20250812/.venv/lib/python3.12/site-packages/gunicorn/workers/sync.py", line 182, in handle_request
    for item in respiter:
  File "/home/psrok1/gunicorn-bug-20250812/.venv/lib/python3.12/site-packages/werkzeug/wsgi.py", line 256, in __next__
    return self._next()
           ^^^^^^^^^^^^
  File "/home/psrok1/gunicorn-bug-20250812/.venv/lib/python3.12/site-packages/werkzeug/wrappers/response.py", line 32, in _iter_encoded
    for item in iterable:
  File "/home/psrok1/gunicorn-bug-20250812/app.py", line 9, in stream_endlessly
    time.sleep(1)
  File "/home/psrok1/gunicorn-bug-20250812/.venv/lib/python3.12/site-packages/gunicorn/workers/base.py", line 204, in handle_abort
    sys.exit(1)
SystemExit: 1
[2025-08-12 20:51:35 +0200] [46085] [INFO] Worker exiting (pid: 46085)
```

When regular Exceptions are handled, `handle_request` checks if headers were already sent (https://github.com/benoitc/gunicorn/blob/a86ea1e4e6c271d1cd1823c7e14490123f9238fe/gunicorn/workers/sync.py#L194). If yes, it means that response is partially sent and best we can do is to close the connection. Unfortunately, BaseExceptions (including SystemExit) are not handled, so we land in this part: https://github.com/benoitc/gunicorn/blob/a86ea1e4e6c271d1cd1823c7e14490123f9238fe/gunicorn/workers/sync.py#L157, which finally writes additional headers and body into the socket (https://github.com/benoitc/gunicorn/blob/a86ea1e4e6c271d1cd1823c7e14490123f9238fe/gunicorn/util.py#L309).

I think that Exception handler in `handle_request` should be extended to handle BaseException as well.

gyanu2507:
Worker timeout fires SIGABRT → `sys.exit(1)`. That's a BaseException, so the `except Exception` guard that checks `headers_sent` never ran. `handle_error` then wrote a 500 onto the same socket, which is how wget ended up with ISE HTML concatenated onto a partial download.

Once headers are out we just close, same as a regular exception.

Fixes #3410

### DIFF
--- a/docs/content/news.md
+++ b/docs/content/news.md
@@ -1,6 +1,15 @@
 <span id="news"></span>
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- Worker timeout used to dump a 500 response onto a body that had already
+  started, because `sys.exit()` from the abort handler is a `BaseException`.
+  Once headers are out we just close the connection.
+  ([#3410](https://github.com/benoitc/gunicorn/issues/3410)).
+
 ## 26.2.0 - 2026-08-24
 
 ### New Features
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -306,7 +306,7 @@ class AsyncWorker(base.Worker):
             # If the original exception was a socket.error we delegate
             # handling it to the caller (where handle() might ignore it)
             util.reraise(*sys.exc_info())
-        except BaseException:
+        except (Exception, SystemExit):
             if resp and resp.headers_sent:
                 # Timeout abort is SystemExit. If we already started the
                 # response, writing a 500 on the same socket corrupts it.
--- a/gunicorn/workers/base_async.py
+++ b/gunicorn/workers/base_async.py
@@ -306,10 +306,10 @@ class AsyncWorker(base.Worker):
             # If the original exception was a socket.error we delegate
             # handling it to the caller (where handle() might ignore it)
             util.reraise(*sys.exc_info())
-        except Exception:
+        except BaseException:
             if resp and resp.headers_sent:
-                # If the requests have already been sent, we should close the
-                # connection to indicate the error.
+                # Timeout abort is SystemExit. If we already started the
+                # response, writing a 500 on the same socket corrupts it.
                 self.log.exception("Error handling request")
                 try:
                     sock.shutdown(socket.SHUT_RDWR)
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -730,7 +730,7 @@ class ThreadWorker(base.Worker):
         except OSError:
             # pass to next try-except level
             util.reraise(*sys.exc_info())
-        except BaseException:
+        except (Exception, SystemExit):
             if resp and resp.headers_sent:
                 # Timeout abort is SystemExit. If we already started the
                 # response, writing a 500 on the same socket corrupts it.
--- a/gunicorn/workers/gthread.py
+++ b/gunicorn/workers/gthread.py
@@ -730,10 +730,10 @@ class ThreadWorker(base.Worker):
         except OSError:
             # pass to next try-except level
             util.reraise(*sys.exc_info())
-        except Exception:
+        except BaseException:
             if resp and resp.headers_sent:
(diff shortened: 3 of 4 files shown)


benoitc:
Thanks for tracking this down, the fix is right: once headers are out we should close, not write a 500.

One thing on the catch itself: `except BaseException` is too broad. The abort handler does `sys.exit(1)`, so the only thing reaching here beyond a normal `Exception` is `SystemExit`. Could you narrow it to:

```python
except (Exception, SystemExit):
```

in all three workers possibly.

benoitc (changes requested):
lPlease look at my comment about the exception. Once handled , the code is OK to merge.

gyanu2507:
Narrowed the headers-sent catch in sync, gthread, and the async worker to `except (Exception, SystemExit)`.

benoitc:
Thanks @gyanu2507, merged.

PR #3716, opened 2026-08-29
http2: bound the request body held per stream

benoitc:
HTTP/2 DATA frames were kept in two buffers (`_body_chunks` and a `request_body` BytesIO), copied twice more at dispatch, and flow-control credit was returned as each frame arrived. A client that never sent END_STREAM could grow a worker's memory without the application ever being called. Only listeners with `h2` in `http_protocols` were affected.

Changes:
- `HTTP2Stream` keeps the body once (`_body_chunks`, `body_size`) and hands it over without copying.
- New `http2_max_request_body_size` / `--http2-max-request-body-size`, default 100MB, 0 for unlimited.
- Both connection classes check the limit before buffering. Over it, the stream is dropped, answered with `413` + `END_STREAM`, then `RST_STREAM(NO_ERROR)` (RFC 9113 8.1). Later frames on that stream are ignored and the connection stays open.
- Docs and changelog updated.

### DIFF
--- a/docs/content/2026-news.md
+++ b/docs/content/2026-news.md
@@ -1,6 +1,19 @@
 <span id="news-2026"></span>
 # Changelog - 2026
 
+## Unreleased
+
+### Bug Fixes
+
+- **HTTP/2 request bodies were buffered without bound**: a stream's DATA
+  frames were kept in two buffers, copied twice more at dispatch, and flow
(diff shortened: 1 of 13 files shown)


benoitc:
Superseded: the body is streamed to the application instead of being capped.

PR #3718, opened 2026-08-29
Skip filesystem checks for abstract namespace unix sockets

### LINKED ISSUE #2699: Gunicorn fail to bind abstract unix domain socket
Abstract Unix domain sockets are not related to the file system and their path starts with \0.
When trying to bind gunicorn to such socket we get the following exception:
```console
~/gunicorn$ cat gunicorn.conf.py 
bind = "unix:\0/foo/bar"
~/gunicorn$ /usr/bin/python3 /home/ubuntu/.local/bin/gunicorn -w 2 _--config gunicorn.conf.py_ echoapp:app
[2021-12-09 21:19:29 +0000] [1310108] [INFO] Starting gunicorn 20.1.0
  File "/home/ubuntu/.local/bin/gunicorn", line 8, in <module>
    sys.exit(run())
  File "/home/ubuntu/.local/lib/python3.8/site-packages/gunicorn/app/wsgiapp.py", line 67, in run
    WSGIApplication("%(prog)s [OPTIONS] [APP_MODULE]").run()
  File "/home/ubuntu/.local/lib/python3.8/site-packages/gunicorn/app/base.py", line 231, in run
    super().run()
  File "/home/ubuntu/.local/lib/python3.8/site-packages/gunicorn/app/base.py", line 72, in run
    Arbiter(self).run()
  File "/home/ubuntu/.local/lib/python3.8/site-packages/gunicorn/arbiter.py", line 198, in run
    self.start()
  File "/home/ubuntu/.local/lib/python3.8/site-packages/gunicorn/arbiter.py", line 155, in start
  File "/home/ubuntu/.local/lib/python3.8/site-packages/gunicorn/sock.py", line 184, in create_sockets
    sock = sock_type(addr, conf, log)
  File "/home/ubuntu/.local/lib/python3.8/site-packages/gunicorn/sock.py", line 108, in __init__
    st = os.stat(addr)
ValueError: embedded null byte
```

2 questions:
1. Is there any open issue for this? tried to look for such but couldn't find it.
2. Would you accept PR with a fix for this bug?

afonsojanu:
Fixes #2699.

Linux supports abstract namespace unix sockets, whose address starts with a null byte and has no presence on the filesystem. Binding gunicorn to one currently crashes before the server starts:

```
bind = "unix:\0my-socket"
```

```
ValueError: embedded null byte
```

`UnixSocket.__init__` always calls `os.stat()` on the address to check for and remove a stale socket file from a previous run. `os.stat()` rejects any path with an embedded null byte outright, so it never has a chance to return `ENOENT` the way a genuinely missing path would, and the `ValueError` propagates straight out of the constructor.

`bind()` has the same problem a step later: it unconditionally chowns the socket path, which would fail the same way once the stat call above stops being in the way.

Since abstract sockets don't exist on the filesystem, there's nothing to stat, remove, or chown for them, so both operations are skipped when the address starts with a null byte.

Added tests covering:
- the abstract-namespace case skips `os.stat()`
- a normal filesystem path still goes through the existing stat/remove logic
- `bind()` skips `chown()` for an abstract-namespace address

Ran the full test suite plus pylint and pycodestyle on the touched files, all clean.

### DIFF
--- a/gunicorn/sock.py
+++ b/gunicorn/sock.py
@@ -126,7 +126,9 @@ class UnixSocket(BaseSocket):
     FAMILY = socket.AF_UNIX
 
     def __init__(self, addr, conf, log, fd=None):
-        if fd is None:
+        # Abstract namespace sockets (address starts with a null byte) have
+        # no filesystem presence, so there's nothing to stat or remove.
+        if fd is None and not addr.startswith('\0'):
             try:
                 st = os.stat(addr)
             except OSError as e:
@@ -145,7 +147,8 @@ class UnixSocket(BaseSocket):
     def bind(self, sock):
         old_umask = os.umask(self.conf.umask)
         sock.bind(self.cfg_addr)
-        util.chown(self.cfg_addr, self.conf.uid, self.conf.gid)
+        if not self.cfg_addr.startswith('\0'):
+            util.chown(self.cfg_addr, self.conf.uid, self.conf.gid)
         os.umask(old_umask)
 
 
--- a/tests/test_sock.py
+++ b/tests/test_sock.py
@@ -2,6 +2,7 @@
 # This file is part of gunicorn released under the MIT license.
 # See the NOTICE for more information.
 
+import errno
 from unittest import mock
 
 from gunicorn import sock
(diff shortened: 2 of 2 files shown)


benoitc:
Thanks @afonsojanu, merged.

PR #3723, opened 2026-08-31
Prefer an absolute-form request-target's own authority over Host

### LINKED ISSUE #3370: Incorrect Handling of Absolute-Form Request-Target Authority
**Version**
bacbf8a

**Platform**
Ubuntu 11.4.0-1ubuntu1~22.04

**Description**
Hello, I may have found a bug in gunicorn's parsing of absolute-form request-target authority.
RFC 9112 says this:
> When an origin server receives a request with an absolute-form of request-target, the origin server MUST ignore the received `Host` header field (if any) and instead use the host information of the request-target.

However, I noticed that when an HTTP request is sent with an absolute-form request-target where the authority differs from the `Host` header, gunicorn appears to prioritize the `Host` header over the request-target’s authority. 
For example:
```
GET http://evil.com/page HTTP/1.1\r\n
Host: victim.com\r\n
\r\n
```
Gunicorn's response:
```
$ echo -ne "GET http://evil.com/page HTTP/1.1\r\nHost: victim.com\r\n\r\n" | nc 172.18.0.7 80
HTTP/1.1 200 OK
Server: gunicorn
Date: Wed, 26 Mar 2025 09:01:53 GMT
Connection: keep-alive
Content-type: application/json
Content-Length: 113

{"headers":[["SE9TVA==","dmljdGltLmNvbQ=="]],"body":"","version":"SFRUUC8xLjE=","uri":"L3BhZ2U=","method":"R0VU"}
```
From the response body, specifically the URI, it appears that Gunicorn parses the above request as:
```
GET /page HTTP/1.1\r\n
Host: victim.com\r\n
\r\n
```
This shows that gunicorn did not ignore the received `Host` header field, which may be a violation of the protocol specifications.

afonsojanu:
Closes #3370.

When a request comes in as `GET http://evil.com/page HTTP/1.1` with a `Host: victim.com` header attached, RFC 9112 3.2.2 is pretty explicit that the origin server has to ignore the Host header entirely and go with the authority already sitting in the request-target. Gunicorn does parse that authority out (`split_request_uri` gives it back as `.netloc`), but nothing ever kept it around, so it just silently used the Host header instead in every case. That's exactly the mismatch reported in the issue: a proxy that forwards the absolute-form request-target as-is (which is legitimate, some do) ends up with the app seeing whatever Host value the client felt like sending, not the address that was actually requested.

Both request-line parsers (the Python one and the C-accelerated one) now hang onto that authority on the `Request` object as `.authority`, which stays `None` for every request-target form except absolute-form. `wsgi.create()` checks it after building the headers and, when it's set, uses it for `HTTP_HOST` and the local `host` variable instead of whatever came in on the Host header. HTTP/2 requests don't have this attribute at all (they have their own `:authority` pseudo-header story already handled separately), so that path is untouched.

Added `tests/test_absolute_form_host.py` with the exact mismatched-host scenario from the issue plus a control case confirming ordinary origin-form requests (the vast majority of traffic) still behave exactly as before. Ran it against both parser backends, along with the rest of the request-parsing and WSGI-environ suites, and pylint/pycodestyle on the touched files.

### DIFF
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -588,7 +588,13 @@ class Request(Message):
         self.path = parts.path or ""
         self.query = parts.query or ""
         self.fragment = parts.fragment or ""
-        self.authority = parts.netloc or None
+        # parts.scheme is only set for a genuine absolute-form request-target
+        # (e.g. "http://host/path"), never for origin-form, even one whose
+        # path or query happens to contain "://". An absolute-form target can
+        # still have an empty authority ("https://" with nothing after it),
+        # which is distinct from not being absolute-form at all, hence the
+        # separate None below rather than folding both into one falsy check.
+        self.authority = parts.netloc if parts.scheme else None
 
         # Version (validation done by C parser)
         self.version = (1, result['minor_version'])
@@ -924,9 +930,13 @@ class Request(Message):
         self.path = parts.path or ""
         self.query = parts.query or ""
         self.fragment = parts.fragment or ""
-        # populated only for absolute-form request-targets (parts.netloc is
-        # empty for origin-form, asterisk-form and CONNECT's authority-form)
-        self.authority = parts.netloc or None
+        # parts.scheme is only set for a genuine absolute-form request-target;
+        # it stays empty for origin-form, asterisk-form and CONNECT's
+        # authority-form. An absolute-form target can still have an empty
+        # authority ("https://" with nothing after it), so that case needs to
+        # stay distinguishable from "not absolute-form at all" rather than
+        # collapsing both into None on an empty-string check.
+        self.authority = parts.netloc if parts.scheme else None
 
         # Version
         match = VERSION_RE.fullmatch(bits[2])
--- a/gunicorn/http/message.py
+++ b/gunicorn/http/message.py
@@ -482,6 +482,7 @@ class Request(Message):
         self.path = None
         self.query = None
         self.fragment = None
+        self.authority = None
 
         # get max request line size (0 means unlimited per documentation)
         self.limit_request_line = cfg.limit_request_line
@@ -587,6 +588,7 @@ class Request(Message):
         self.path = parts.path or ""
         self.query = parts.query or ""
         self.fragment = parts.fragment or ""
+        self.authority = parts.netloc or None
 
         # Version (validation done by C parser)
         self.version = (1, result['minor_version'])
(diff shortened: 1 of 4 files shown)


benoitc:
Thanks for the patch and the RFC references. I looked into this and I don't think we should merge it as-is.

RFC 9112 3.2.2 has two rules. The proxy rule says a proxy receiving absolute-form MUST rewrite Host from the request-target authority and forward it (normally as origin-form). The origin-server rule says an origin server ignores Host and uses the authority. Behind a proxy, which is how gunicorn is normally deployed, the first rule already applies: the proxy normalizes to origin-form and sets the right Host before gunicorn sees the request. That is exactly what the Apache test in #3370 shows, the upstream received `Host: evil.com` in origin-form. nginx does the same. So the request-target at gunicorn is relative and the Host is already correct.

That means this only changes anything for a directly exposed gunicorn receiving absolute-form. In that case both the Host header and the request-target authority come from the same client, so switching `HTTP_HOST` from one to the other does not fix a wrong value, it just picks a different attacker-supplied string. It is a spec-compliance detail, not a security fix, and there is no victim in the scenario as described.

Two concrete issues with the change:

- Empty authority is not handled. `GET https:// HTTP/1.1` gives an empty netloc, so `authority` stays None and the client Host is used unchanged. RFC 9112 says an empty Host should be sent in that case, and PEP 3333 requires SERVER_NAME/SERVER_PORT to never be empty. This is the edge case already raised in the issue and it is not covered.
- userinfo leaks into Host. `parts.netloc` is copied verbatim, so `GET http://user:pass@host:8080/p` puts `user:pass@host:8080` into `HTTP_HOST`, which is not a valid Host value.

For an origin server that is not a proxy, rejecting a mismatched absolute-form authority with a 400 is a more defensible behavior than trusting it. Happy to discuss that direction, but I would not take the current change.

afonsojanu:
Fair points on both bugs, fixed. `GET https:// HTTP/1.1` now sends an empty Host instead of quietly falling back to whatever the client's Host header said (the parser now tracks whether the request-target was absolute-form at all through `parts.scheme`, since an empty authority and no authority component aren't the same thing, even though both used to collapse into the same falsy check), and userinfo gets stripped out before anything lands in `HTTP_HOST`, so `http://user:pass@host:8080/p` gives you `host:8080` rather than the whole thing verbatim.

I went and read 3.2.2 again before touching the reject-with-400 idea, and I don't think the spec actually leaves room for it. Both the proxy rule and the origin-server rule are phrased as a plain MUST-ignore-Host-and-use-the-authority, full stop, including the empty-authority case where it just says to send an empty Host. There's no branch in there for "and reject if they disagree" — that would be introducing behavior beyond what 9112 asks for, not really complying with it a different way. So I kept the substitution rather than switching to rejection, just made it match what the spec actually describes instead of what the original patch approximated.

Where I do think you're right: this was never a fix for an actual vulnerability, since behind a proxy the request never reaches gunicorn as absolute-form in the first place (matches what you found with Apache/nginx), and directly exposed, both values are equally client-controlled. It's a compliance gap, and the PR description probably should've said as much instead of leaning on the RFC quote and letting it read like a security patch by implication.

Re-ran the whole suite plus the new cases against both parsers, all green.

PR #3726, opened 2026-09-03
fix: announce Connection: close from the ASGI worker

dylanpulver:
The ASGI worker decides not to reuse a connection and then closes it without sending a `Connection` header. RFC 9112 §9.6 requires a server that will not reuse a connection to say so, and the sync WSGI worker already does, in `Response.default_headers()`.

Measured against that sibling, one server each, raw socket:

| case | WSGI | ASGI before | after |
|---|---|---|---|
| client sent `Connection: close` | announces | **closes silently** | announces |
| `--keep-alive 0` | announces | **closes silently** | announces |
| keep-alive default | stays open | stays open | stays open |

A client seeing an HTTP/1.1 response with no `Connection` header treats it as persistent (§9.3) and reuses a socket the server already closed. Also reached by worker shutdown and by a uWSGI response delimited by EOF.

`_send_response_start()` and the keepalive return of `_handle_http_request()` now read one predicate, `_will_close()`, so the announced framing cannot drift from the framing used. An app-supplied `Connection` header is left alone.

Verification, same env each run:
- full suite 2044 passed / 535 skipped, before and after, identical
- mutant A (revert `protocol.py`, keep tests) → 8/8 new tests fail, 0 pre-existing
- **mutant B (`close=request.should_close()`) passes all 2044** — it covers only the client-requested case; live it still closes unannounced under `--keep-alive 0`. Hence the shared predicate rather than a second derivation.

pylint 3.3.2 10.00/10 and pycodestyle clean.

### DIFF
--- a/gunicorn/asgi/protocol.py
+++ b/gunicorn/asgi/protocol.py
@@ -1291,7 +1291,10 @@ class ASGIProtocol(asyncio.Protocol):
                     uses_uwsgi and not has_content_length and not omits_body
                 )
 
-                self._send_response_start(response_status, response_headers, request)
+                self._send_response_start(
+                    response_status, response_headers, request,
+                    close=self._will_close(request, response_requires_close),
+                )
 
             elif msg_type == "http.response.body":
                 if not response_started:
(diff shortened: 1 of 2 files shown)


benoitc:
Thanks for this, and for the clear write-up. I picked up your shared `_will_close` approach in #3731, which also handles the `max_requests` recycle case (decided in the loop after the response starts, so it has to be anticipated when the header is written) and adds a live wire test for it. Crediting you there. Closing this in favor of #3731.

PR #3736, opened 2026-09-06
Remove the legacy code which duplicated settings objects (1-line change)

deanliang06:
According to the git history, it used to be that the KNOWN_SETTINGS would contain the settings objects, and so for different configs to have different objects, copying was necessary. However, through a re-write, this copoy mechanism continued but unnecessarily.


deanliang06:
@benoitc I don't want to sound impatient, but are only more significant changes incorperated into gunicorn? I want to start contributing, so your insight would be very much appreciated. Thanks!!

PR #3751, opened 2026-09-22
Honor Accept when writing worker error pages

### LINKED ISSUE #3246: Server Error -  Respect the Accept header in responses
When a timeout occurs, gunicorn kill the worker (good) and answers the client (good) with a fixed error page in HTML, regardless of the `Accept` header sent by the client (bad).

I would like that this header would be taken into account when returning the error page, so that a client can expect the response to be in his wanted type.

For example, when a client requests XML, he gets XML, when he requests HTML, he gets HTML, and when he requests JSON, he gets JSON.

Related to #2281

Shaisolaris:
Fixes #3246

Timeout and request-error pages were always text/html.

write_error() now looks at Accept and returns application/json or application/xml when those types outrank HTML. The default is still HTML.

Tests cover the default, JSON, XML, and a quality value.


pajod:
(**Not** a change request - certainly not before any interest to actually merge this is signalled. Just a note to other reviewers / later follow-up PRs)

* changes `write_error` (general case) but not `make_fail_app` (special case during startup)
* retains the `text/html` default
* responds with `application/xml` when asked for `text/xml` (the latter not implying preparedness for non-ASCII)
* matching is only partially implemented, e.g. `text/*`
* deviates from rfc9110 in optional whitespace & treatment of unusual floating point like `text/html;q=-1e300`

